PR #6, opened 2014-10-17
Add scroll and loading indicators, multiple click support

jmhmd:
- add image loading progress indicator
- add scroll position indicator
- allow tool mouse button to be array for multiple button trigger (i.e. right and left mouse button together to adjust window/level)
- disable image update on tool switch, triggered image render which for large images like mammograms causes long UI lag


chafey:
Hi Jason - I am going to close this because it was too big of a commit and hopelessly behind trunk.  It would be great if you could break these into smaller pull requests.  Perhaps we can chat sometime to figure out what should go in here vs maybe another library?

jmhmd:
No problem, I agree. I have made a few other changes I am hoping to clean
up in to pull requests as well.

PR #10, opened 2014-12-03
WIP: Rectangular highlighting tool

swederik:
Hello!

I've hacked together a highlighting tool by adding a darkening box to the canvas. It seems to work fine, but I'd like to make the handles a bit easier to see. If you know a better way to do the canvas drawing, I'm open to suggestions.

It looks like this at the moment:

screen shot 2014-12-03 at 01 13 18

The highlight box can also be moved around by clicking anywhere inside the box. There is a new "mouseButtonRectangleTool" to facilitate this, and some changes to the moveHandles and moveAllHandles to prevent the handles from going off screen, rather than just deleting them if they do.

The relevant canvas drawing section is done like this:

        context.rect(rect.width + rect.left, rect.top, -rect.width, rect.height);

The drawing clockwise and counterclockwise is done by following the second part of this StackOverflow question: http://stackoverflow.com/questions/6271419/how-to-fill-the-opposite-shape-on-canvas

Thanks!

Erik


chafey:
Merged, thank you for this contribution!

PR #12, opened 2015-02-04
Remove/Reset all the tool state associated with an enabled element.

rbmrclo:
- Adds `src/imageTools/toolDirectory.js` - this module should be the config file, responsible for global access of tool types
- Adds `clearAllToolData` function that iterates over each tools that are enabled to the element, then clear all data by resetting its value to `[] / empty array`.


chafey:
Thank you for this contribution!  I don't quite understand the idea behind the toolDirectory - can you explain why this is needed?

rbmrclo:
I realized that this isn't necessary anymore. Thanks @chafey!

PR #21, opened 2015-03-31
Fixes background rectangle when highlighting

### LINKED ISSUE #20: highlight does not use the correct width / height for drawing background rect
This is related to / explanation for:
https://github.com/chafey/cornerstone/issues/16#issue-65320330

See:
http://stackoverflow.com/a/6722031/2638485

but basically this call is not correct if `scale != 1`:
```
```
which it likely won't be after `setToPixelCoordinateSystem`.

cancan101:
fixes #20


chafey:
Wouldn't merge so I made the change manually to my local file, thanks for the contribution!

cancan101:
@chafey 
1. I don't think this is the same fix. You are calling `restore` AFTER painting the selected rect. I had:
```diff
+        context.save();
+        context.setTransform(1,0,0,1,0,0);
         context.rect(0, 0, context.canvas.clientWidth, context.canvas.clientHeight);
+        context.restore();
+
         context.rect(rect.width + rect.left, rect.top, -rect.width, rect.height);
```
2. It looks like `crosshairs` ended up in this push.

Issue #19, opened 2015-03-23
All Image Tools Example Does not Update WW/WC or Zoom Text


cancan101:
When changing WW, WC or zooming in on the [All Image Tools Example](http://rawgit.com/chafey/cornerstoneTools/master/examples/allImageTools/index.html), the caption text is not changing. 

It seems that `onViewportUpdated` is not being called.

cancan101:
@swederik Is this fixed?

swederik:
Yep, this can be closed

PR #25, opened 2015-04-13
Cleanup of allImageTools example

swederik:
This is just a bit of cleanup for the allImageTools example. It also fixes issue #19 by changing onViewportUpdated to use CornerstoneImageRendered. I'm not sure if this is appropriate, but it works and I don't see the CornerstoneViewportUpdated event in the cornerstone code anywhere

allimagetoolsgif
.

### DIFF
--- a/examples/allImageTools/index.html
+++ b/examples/allImageTools/index.html
@@ -25,19 +25,21 @@
     <br>
 
     <div class="row">
-        <div class="col-xs-6">
-            <button id="enableWindowLevelTool" type="button" class="btn">WW/WC</button>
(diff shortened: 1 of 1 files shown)


cancan101:
What is the difference between `CornerstoneImageRendered` and `CornerstoneViewportUpdated`?

swederik:
I don't know, but CornerstoneViewportUpdated seems to be deprecated.

chafey:
I believe the original idea between the two events had to do with performance.  An image could be rendered without the viewport necessarily changing (e.g. moving a tool).  I don't think I ever needed this so I removed it, it could return someday

PR #26, opened 2015-04-14
Fixes background rectangle when highlighting

### LINKED ISSUE #20: highlight does not use the correct width / height for drawing background rect

cancan101:
Fixes #20


cancan101:
See: https://rawgit.com/cancan101/cornerstoneTools/3c2db475da7c83a19f4ccbf8be4c8de4525ffd22/examples/highlight/index.html

chafey:
@swederik Erik can you review this change this you contribute the rectangle tool?

swederik:
Sorry for the slow reply, I didn't notice this

The fix works fine, but the PR also includes a regression of the included cornerstone library. The branch needs to be rebased from the master branch. To save @cancan101 some time, I just made a new PR #33 with the same fix but no regression problem.

swederik:
This can be closed since #33 was merged.

PR #29, opened 2015-05-04
Clear measurement tools

shahriar:
Adding a simple feature to clean up measurement tools like length, probe, etc.


cancan101:
Does it make sense to have some sort of clear all functionality that loops over all tools?

swederik:
Yes it would be useful to have that function. This PR is a good idea but it would be better to add a clearToolData function to the toolStateManager, rather than to each individual tool, as is done here. 

The imageId-specific tool state manager (https://github.com/chafey/cornerstoneTools/blob/master/src/stateManagement/imageIdSpecificStateManager.js#L46) should also be able to clear data (along with adding and getting data). This would solve allow you to clear all data, at least for a single image.

swederik:
Closing this since I just merged #40 

Thanks for your contribution! @shahriar

Issue #31, opened 2015-05-07
saveAs Tool


cancan101:
Some way of "saving" the transformed image to PNG, etc.

For example, see:
* http://stackoverflow.com/questions/11112321/how-to-save-canvas-as-png-image
* https://github.com/eligrey/FileSaver.js

chafey:
Technically this should not be hard to do - can you propose a UI design for this?

swederik:
Closing since #34 was merged

Issue #30, opened 2015-05-07
Means of Serializing out Transform + Annotations


cancan101:
For a given input image and a set of transforms and annotations applied, have some way of serializing out these transforms and annotations so that they can be re applied to the same image later.

chafey:
This is something we are discussing in our OHIF architecture meetings.  Will update this issue when we have made some decisions

swederik:
I agree this would be useful. At the moment for viewport I am using something like this (where ````viewportString```` is sent to the server and returned when the page is loaded again):

```` javascript
// every few seconds
var viewport = cornerstone.getViewport(element);
var viewportString = JSON.stringify(viewport);

// on page load
var savedViewport = JSON.parse(viewportString);
cornerstone.setViewport(element, savedViewport);
````

I could image something similar working like 

```` javascript
var toolData = cornerstone.getAllToolStates(element);
````

or perhaps some function attached to the the toolStateManager(s)

**Comment by cancan101** [comment]
This is an example of the data using that technique:
```json
{"scale":4.3359375,
"translation":{"x":11.070270270270273,"y":-0.23063063063063063},
"voi":{"windowWidth":1354.671875,"windowCenter":333.0546875},
"invert":true,
"pixelReplication":false,"rotation":0,"hflip":false,"vflip":false}
```
So WW/WC, invert, rotate, flip, scale and translate seem to be captured. It appears the various annotations are not captured.

swederik:
So I think implementing this in cornerstoneTools will be a bit more complicated (because of the different types of toolStateManagers), but if you need to do this right now you can just do something like this:

```` javascript
// --- To serialize the tool states ---
var element = document.getElementById("dicomImage");
var allToolData = {};

// Decide which tool types you want to serialize
var toolTypes = ['length', 'angle'];

// Loop through all the tool types
for (var i = 0; i < toolTypes.length; i++) {
    // Get the tool data for this tool type
    var toolType = toolTypes[i];
    var toolData = cornerstoneTools.getToolState(element, toolType);

    if (toolData !== undefined) {
        // Put it into an object
        allToolData[toolTypes[i]] = toolData;
    }
}

var toolDataString = JSON.stringify(allToolData);
````

Then to put the tool data back:

```` javascript
// Try pasting this block into the allImageTools example: 
var element = document.getElementById("dicomImage");
[...]

PR #34, opened 2015-05-08
SaveAs tool and example

swederik:
Small function for saving PNG files from the canvases. Addresses #31.

~~Seems it can't be automatically merged because of changes to dist/cornerstoneTools... How should we approach this? I assume I stop using grunt before commits? But then the rawgit examples won't work...~~

Update: Rebased and fixed

saveas


cancan101:
Is it possible to use any of this code for send the binary image over HTTP, for example using [`window.fetch`](https://github.com/github/fetch#file-upload)?

swederik:
Yeah, sure, but what for? You want to save the canvas as an image on your server?

I think you can just use canvas.toDataURL(); and then POST that data to the server. Then you just decode the base64 string to produce an image on the server.

cancan101:
Yes, exactly. I want to save the save the canvas (which has all of the user's modifications, transforms and annotations) to the server. This along with #30 so that I have both the steps to re-create the modifications and the image of the canvas.

swederik:
Ok so you want to use the image to show the saved-state then. Seems easy enough to do, but I don't know if it's worth having extra code for it in cornerstone itself.

```` javascript
// Get the canavas
var canvas = $(element).find("canvas").get(0);
var imageData = canvas.toDataURL();
$.ajax({
  method: "POST",
  url: "/your/server/",
  data: {"image": imageData}
})
````

Then on the server (at least in Python):

```` python
import base64

// Assuming you have a request like in Django
data = request.POST["image"]

filename = "MyNewImage.png"
out_file = open(filename, "w")
out_file.write(base64.decodestring(data))
out_file.close()
````

Hope that helps

**Comment by cancan101** [comment]
Sure that seems simple enough. I do think though that a save as tool for in browser saves still stands on its own as a useful feature.

File: dist/cornerstoneTools.js [hunk 1]
```diff
@@ -2229,6 +2229,55 @@ var cornerstoneTools = (function ($, cornerstone, cornerstoneMath, cornerstoneTo
...
+
+    "use strict";
+
+    if(cornerstoneTools === undefined) {
+        cornerstoneTools = {};
+    }
+
+    function saveAs(element, filename)
+    {
+        var canvas = $(element).find("canvas").get(0);
```
cancan101: Does it make sense to extract `$(element).find("canvas").get(0);` as a helper method?

swederik: Probably not, it's boilerplate jQuery.

PR #41, opened 2015-05-19
marke highlighted shapes as active. Possibility to delete/manipulate selection

ra00l:
marked drawn shapes as active, to be able to identify &
delete/manipulate them from outside the library.

Now I can do:

``` javascript
$(document).on('keyup', function (evt) {
    //might need to filter based on target
    if (evt.keyCode == 46) { //delete
        var removed = false;
        _.each(['length', 'ellipticalRoi', 'rectangleRoi', 'angle', 'probe'], function (shape) {
            var data = (cornerstoneTools.getToolState(rawElement, shape) || {}).data || [];
            for (var idx = data.length - 1; idx >= 0; idx--) {
                if (data[idx].active) {
                    data.splice(idx, 1);
                    removed = true;
                }
            }
        });
        
        if(removed)
            cornerstone.updateImage(rawElement);
    }
});
```

### DIFF
--- a/src/imageTools/probe.js
+++ b/src/imageTools/probe.js
@@ -69,13 +69,7 @@ var cornerstoneTools = (function ($, cornerstone, cornerstoneTools) {
     }
     ///////// BEGIN IMAGE RENDERING ///////
     function pointNearTool(data, coords) {
-        var point = {
-            x: data.handles.end.x,
-            y: data.handles.end.y
-        };
-
-        var distanceToPoint = cornerstoneMath.point.distance(point, coords);
-        return (distanceToPoint < 5);
+        return  cornerstoneMath.point.distance(data.handles.end, coords) < 5;
     }
 
 
--- a/src/imageTools/probe.js
+++ b/src/imageTools/probe.js
@@ -70,13 +70,11 @@ var cornerstoneTools = (function ($, cornerstone, cornerstoneTools) {
     ///////// BEGIN IMAGE RENDERING ///////
     function pointNearTool(data, coords) {
         var point = {
-            left: data.handles.end.x,
-            top: data.handles.end.y,
-            width: 0,
-            height: 0
+            x: data.handles.end.x,
+            y: data.handles.end.y
         };
 
-        var distanceToPoint = cornerstoneMath.rect.distanceToPoint(point, coords);
+        var distanceToPoint = cornerstoneMath.point.distance(point, coords);
         return (distanceToPoint < 5);
     }
 
--- a/src/imageTools/probe.js
+++ b/src/imageTools/probe.js
@@ -68,6 +68,18 @@ var cornerstoneTools = (function ($, cornerstone, cornerstoneTools) {
         return suv;
     }
     ///////// BEGIN IMAGE RENDERING ///////
+    function pointNearTool(data, coords) {
+        var point = {
+            left: data.handles.end.x,
+            top: data.handles.end.y,
+            width: 0,
+            height: 0
+        };
+
+        var distanceToPoint = cornerstoneMath.rect.distanceToPoint(point, coords);
+        return (distanceToPoint < 5);
+    }
+
 
     function onImageRendered(e, eventData) {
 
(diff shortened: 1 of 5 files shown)


swederik:
Hello!

I like the goal: Press delete to remove a tool, but there are some issues with the PR:

A few small things:

- The spacing on all the data.active statements is not in-line with the rest of the code in the same block
- You are using a copy/pasted version of pointNearTool for the probe tool. The one you pasted is used to calculate the distance between a point and a rectangle, so it is not appropriate for the probe tool. You should change it to something like cornerstoneMath.point.distance (https://github.com/chafey/cornerstoneMath/blob/master/src/point.js#L33) in order to calculate the distance between the cursor and the probe tool.

By the way, you can use github flavoured markdown to print code more nicely:

https://help.github.com/articles/github-flavored-markdown/#syntax-highlighting

```` javascript
$(document).on('keyup', function(evt) {
    //might need to filter based on target
    if (evt.keyCode == 46) { //delete
        var removed = false;
        _.each(['length', 'ellipticalRoi', 'rectangleRoi', 'angle', 'probe'],
            function(shape) {
                var data = (cornerstoneTools.getToolState(rawElement, shape) || {}).data || [];
                for (var idx = data.length - 1; idx >= 0; idx--) {
                    if (data[idx].active) {
                        data.splice(idx, 1);
                        removed = true;
                    }
                }
            });

        if (removed)
            cornerstone.updateImage(rawElement);
    }
});
````

File: src/imageTools/probe.js [hunk 1]
```diff
@@ -68,6 +68,18 @@ var cornerstoneTools = (function ($, cornerstone, cornerstoneTools) {
...
     ///////// BEGIN IMAGE RENDERING ///////
+    function pointNearTool(data, coords) {
+        var point = {
+            left: data.handles.end.x,
+            top: data.handles.end.y,
+            width: 0,
+            height: 0
+        };
+
+        var distanceToPoint = cornerstoneMath.rect.distanceToPoint(point, coords);
```
swederik: This should probably be 

```` javascript
var distanceToPoint = cornerstoneMath.point.distance(point, coords);
````

but you should test it to make sure it works.

ra00l: Nope, it doesn't work.

What I did with my code was to treat the point like a rectangle.

distance function expects not complex objects, but numbers. The output 
is NaN in your case.

I will have a closer look this evening at the code, see if I can pull it 
through.

Sorry about the bad spaces, I'm kind-of new at social coding + my editor 
didn't saw the right spacing for the source files (probably because I'm 
on Windows).

Thanks for the help and suggestions.

swederik: Ah yeah sorry I see what you mean. You should just delete the definition of point and use

```` javascript
var distanceToPoint = cornerstoneMath.point.distance(data.handles.end, coords);
````

Try Sublime Text (http://www.sublimetext.com/) for a good editor. It can help with the spacing issue.

ra00l: Updated the probe's pointNearTool.

BTW man, this library is FREAKING AWESOME. Thanks for open-sourcing it!

It's the kind of lib that empowers you, as a dev, to do things that will 
impress.

This is probably the first commit of many others :)

swederik: Great! More collaborators is always good. @chafey deserves all the credit

ra00l: Erik, is the last change OK? Do I need to do anything else to accept my PR?

Not in a hurry, just asking :)

swederik: It seems fine to me, I think you just have to wait and see what Chris thinks :-)

swederik:
@ra00l - Thanks for your contribution!

ra00l:
I've got some more goodies, like deleting the highlight shape. But need to make a bit of time for it, maybe tomorrow.

Thanks for accepting the contribution.

Sent from my Phone

-----Original Message-----

swederik:
There are some improvements coming for clearing specific / global tools. I'm actually in the process of merging it right now. See #40.

PR #44, opened 2015-05-26
highlight tool activation + prevent creating >1

ra00l:
1. Added activate option for highlight tool (as a side note, I could not
figure out how to activate the when the mouse is inside the rectangle).
Now, activation only occurs when hovering the handles.
2. I found it weird to be able to draw multiple highlights, but could
not see them unless I deleted the current one. So now, you can only draw
one rectangle.
3. The handles were drawn under the overlay, which made them difficult
to see/select. I moved the code that draws them last, so they will
appear on top of the overlay.


src/imageTools/mouseButtonRectangleTool.js
```diff
@@ -22,6 +22,9 @@ var cornerstoneTools = (function ($, cornerstone, cornerstoneMath, cornerstoneTo
         function addNewMeasurement(mouseEventData)
         {
             var measurementData = mouseToolInterface.createNewMeasurement(mouseEventData);
+            
+            //prevent adding new measurement if tool returns nill
+            if (!measurementData) return;
```
swederik: Can you make this

````` javascript
if (measurementData === undefined) {
    return;
}
````

Since this may be used for tools other than Highlight later on, and it will keep things consistent.

File: src/imageTools/highlight.js [hunk 2]
```diff
@@ -83,9 +88,15 @@ var cornerstoneTools = (function ($, cornerstone, cornerstoneMath, cornerstoneTo
 
         context.save();
         var data = toolData.data[0];
-
-        var selectionColor="white",
-            toolsColor="white";
+        if (!data) return;
+
+        if (pointNearTool(data, cornerstoneTools.toolCoordinates.getCoords())) {
```
swederik: You can use the pointInsideRect function to check if the cursor is inside the highlight tool.

Something like this should work:

````javascript
var coords = cornerstoneTools.toolCoordinates.getCoords();
if (pointNearTool(data, coords) || pointInsideRect(data, coords))
{
...
}...
````

File: src/imageTools/highlight.js [hunk 3]
```diff
@@ -11,6 +11,11 @@ var cornerstoneTools = (function ($, cornerstone, cornerstoneMath, cornerstoneTo
     ///////// BEGIN ACTIVE TOOL ///////
     function createNewMeasurement(mouseEventData)
     {
+        //if already a highlight measurement, creating a new one will be useless
+        var existingToolData = cornerstoneTools.getToolState(mouseEventData.event.currentTarget, toolType);
+        if (existingToolData && existingToolData.data && existingToolData.data.length > 0)
+            return null;
```
swederik: You can just use

```` javascript 
return;
````

since that's what is usually done everywhere else in the library.

File: src/imageTools/highlight.js (hunk 2 above)
Comment (ra00l) [inline src/imageTools/highlight.js]: Thanks for the suggestion, I will have a look. But I remember that the 
issue was that onImageRendered wasn't called unless the mouse was near 
the handles. I will check again, using your suggestion.

Thx!

**Comment by swederik** [comment]
It seems I can't merge this from Github. Could you rebase from upstream/master and resolve any conflicts? (Or create a new branch from upstream/master and cherry-pick this commit)

````
git remote add upstream git@github.com:chafey/cornerstoneTools.git
git fetch upstream
git checkout master
git rebase upstream/master
````

Could you also build with grunt, and paste a link from rawgit.com with the Highlight example in your branch, so that I can test the changes online?

By the way, in the the future it's better if you create your changes in a new branch (other than Master), so that you can have more than one open at a time.

Thanks!

**Comment by ra00l** [comment]
"You can use the pointInsideRect function to check if the cursor is inside the highlight tool."
-> actually, the onImageRender of the rectangle is ONLY called when move is near the 2 handles. Never when mouse is inside the rectangle. This might be a second bug, not sure.

I will do all those suggestions to make it easy for you to merge to master. Sorry about the mess.

**Comment by swederik** [comment]
That's odd, if I remember correctly, the highlight tool used the moveAllHandles function, so you can grab the whole rectangle from the middle and move it around. Highlight was the first tool I made for cornerstone, so I'm not surprised it's full of bugs (sorry!).

**Comment by ra00l** [comment]
I think I fixed the conflict. Still need to learn more on how git & github works :)

The URL to check would be this: http://rawgit.com/ra00l/cornerstoneTools/master/examples/highlight/index.html

However, it doesn't seem like cornersoneTools.js is the one on my master. Maybe they're caching the files?

File: dist/cornerstoneTools.js [hunk 4]
```diff
@@ -1661,10 +1669,19 @@ var cornerstoneTools = (function ($, cornerstone, cornerstoneMath, cornerstoneTo
         cornerstone.setToPixelCoordinateSystem(eventData.enabledElement, context);
 
         //activation color
-        var color=cornerstoneTools.toolColors.getToolColor();
+        var color='white';
```
swederik: You can remove this, seems to be an artifact from merging/rebasing

swederik:
Yeah rawgit seems to do some caching. You sometimes have to wait a bit before the latest changes appear. It might also help to open the chrome dev tools to clear your browser cache.

Changes look good. Just remove the unnecessary line I noted above and I'll merge it. 

You may have also noticed that the active color isn't changing properly. I believe this is an issue with my recent change from activeToolcoordinate to toolCoordinates, and I'm looking into it.

ra00l:
will do the change tomorrow morning.
do you know if anyone is working on an adnotation tool? would need something like this, and don't know if it's already implemented or I should do it.

swederik:
Yeah, actually, for text+arrow annotations, see #47 (and feel free to add any suggestions over there).

ra00l:
wow!!! this is exactly what I needed :) I will put some feedback there once I test it out.
Thanks a lot, you saved me a couple of days with that link.

I also removed the default color assignment, should be fine

swederik:
Seems fine to merge. Thanks!

ra00l:
Glad I could help!

PR #45, opened 2015-05-26
WIP: Basic magnification tool and example

swederik:
**Don't merge yet, work in progress**: I plan to add an option to make the magnification window a circle

This is a magnifying-glass tool for examining images. It also supports touch gestures as magnifyTouchDrag.

**Live demo:**

https://rawgit.com/swederik/cornerstoneTools/enh/magnify/examples/magnify/index.html

**Screencast:**

magnifyexample


ra00l:
wooow, Erik, you're reading my mind :)

Got a fix for you: the element reference in createMagnificationCanvas is undefined (at least in my version). My fix was to add it as parameter to the function, and it works fine. 

Now I have an offset error when shoing the magnified image, might be something in my code, will investigate and come back

swederik:
Ah great yeah you're right, it was a bug that turned up while I was refactoring it to add touch input.

ra00l:
Is there any way to get a copy of this branch? Would make things easier.

Some more stuff:
1/ the container for the canvas-es needs to be relative. Otherwise the zoom will be off (it's my case). I would enforce it directly when creating the small canvas.
2/ in the demo, you're creating the canvas twice (just refresh and inspect with dev tools) (this happened in chrome, at least). so I would make a check to see if the element is already created:
```javascript
if (!document.getElementById(magnificationCanvasId))
            createMagnificationCanvas(element);  
```
3/ using ID is not extensible at all. Imaging having 2 cornerstone instances on the same page. I would just create the zoom canvas, and keep a reference to it. Or, since you already have the parent element, use className + parent element to uniquely identify the instance.
4/ Mouse cursor as arrow takes a bit of space. Don't have a suggestion here, because I'm still off. Ideally, you would hide the cursor while the zoom is visible.

Back with more :)

ra00l:
Ah, forgot something. The width/height of the zoom region is now hardcoded. I am using a full-page dicom viewer, and this seems too small. I would be a good idea to either:
1/ allow passing of width/height to the tool (or, since you're using a square, just one of them)
2/ use percentages for width/height. If you need fixed w/h for canvas, maybe calculate 20% of the viewer region as the w/h of the zoom canvas.

swederik:
1/ Hadn't run into this yet, thanks for reporting it
2/ Just fixed this, haven't pushed it yet
3/ Agreed, I was also just updating it to use a class selector and jQuery .find under the specific parent element
4/ Agreed, I'll look into how to do that

I am looking at width/height/shape options right now. Not sure where to pass them in (in enable, I guess).

swederik:
I forgot to answer your earlier question: Yes it's easy to get a copy of the branch. 

What you can do is look at the top to see that I'm trying to merge from swederik:enh/magnify. So what you need to do is add my repo as a remote repository, then fetch and checkout the enh/magnify branch. Do this:

```` bash
git remote add swederik git@github.com:swederik/cornerstoneTools.git
git fetch swederik
git checkout swederik/enh/magnify
git checkout -b YourNewBranchName
** make some changes **
git push origin YourNewBranchName
** Make a pull request to my fork **
````

Having said that, it's often easier to just suggest changes rather than making them and filing a PR to someone else's fork (in my opinion).

**Comment by ra00l** [comment]
sounds way too complicated for me (if it makes your life easier, I will do it, otherwise no). 

I will have one thing to fix/see why it's happening (zoom is fine, mouse if offset-ed because of a container's parent padding), then I can help you with some of my observations. 
Which one(s) can you leave me to fix/implement?

**Comment by swederik** [comment]
I already did all the stuff above (see the last commit. The live example is also up to date) except for making the container relative. Which container do you mean? The #dicomImage or the div.cornerstone-enabled-image?

Can you post a screenshot?

**Comment by ra00l** [comment]
in your example, you need to disable position absolute on the immediate container (#dicomImage) + position: relative on the next 2 parents (last being the boostrap div.col-xs-12.col-sm-6) to reproduce my initial bug. 

The new round zoom looks reaaaly nice!

**Comment by swederik** [comment]
Ok yeah I see what you mean, but I don't see how to avoid this. The position of the magnifying glass is set as absolute but is relative to the div container above it. I'm open to suggestions, but I don't think the tool should change the CSS on the enabled element.

**Comment by ra00l** [comment]
Well, on initializing the tool, make sure the container is relative, by setting position: relative on it from JS. Or, at least, document this as a requirement, because it might be tricky for a first time user.

New version, new bug :)

In createMagnificationCanvas
```javascript
            magnify.width = magnifySize;
            magnify.height = magnifySize;
```
This is working for you because you declared magnifySize / magnifyLevel globally. Otherwise you will get an undefined error. You should keep a reference to it in the tool class, at set those on activate/enable.

ra00l:
I have a weird behaviour in my scenario: the left position if offset because of a padding I have in one of the dicom image's parents (main container in the image).
I had to add this line to the drawMagnificationTool function, after var canvasLocation...
```
canvasLocation.x += $(main).parent().offset().left;
```

It's still offset by 30px, but don't have ANY idea why. I had a look on how image.x and image.y are calculated in cornerstone.pageToPixel, seems to be fine.

Note that drawing whatever shape (line, ellipse) is working fine.
loupe-bug

swederik:
OK I got rid of the stupid global variables and made magnifySize and magnificationLevel part of magnify's toolData. I also added a warning about the CSS rule to the magnify example.

As for the last problem, I have no idea (yet!), but I suspect there is a smarter / more robust way to position the magnifying glass.

ra00l:
I will look into it once I get some free time. I see that for drawings, you use .getCoords to get the mouse coordinates. Maybe this is causing the offset; will need to check.

Anyway, REAALY amazing stuff you did here...

swederik:
OK I see what you mean. I have the same problem now in another project. It also fails when you start using a rotated or flipped image, for example. I'll look into this now.

swederik:
So I changed a lot. The magnify tool now appears directly on the mouse cursor / touch event's clientX/clientY coordinates. It also handles rotated/zoomed/translated images now by getting the mouse location on canvas with ````canvas.getBoundingClientRect();````.

I hope this fixes your offset issue..

ra00l:
Thanks a lot for your work. At first look, the magnify canvas is not created because the code below returns null. If it works for you, I will digg into the issue some more
```
cornerstoneTools.getToolState(element, toolType); 
```

swederik:
Try updating the branch, I fixed that issue a few commits ago

````
git pull swederik enh/magnify
````

**Comment by ra00l** [comment]
Figured out the crash: I didn't call enable on the tool. But it used to work without that :)
Maybe you can be more permissive and allow enable on activate? And not return underfined, when activating without calling enable?

Now, I get a crash 

```
 var canvasLocation = {
            x: eventData.currentPoints.client.x - rect.left,
            y: eventData.currentPoints.client.y - rect.top
        };
```

There's only page and image, never client. You changed some other files too?

swederik:
That's odd. I added client in the latest commit (to touchInput and mouseInput) and I can see it in the changed files... The demo also works.

swederik:
I'd prefer to keep the functions of enable and activate distinct, I think. I'll merge this since it seems stable now.

ra00l:
It's working perfectly in my scenario! Thanks a lot!

I let you know if I find something else; but it looks really good so far.

PR #46, opened 2015-05-26
Window width/level by region

swederik:
A simple tool to set the window width and level from the minimum and maximum pixel values in the selected region. Also supports touch input.

*Side note:* We should really add a CT image to the example data, since this demo looks horrible on an MRI.

**Live demo:**
https://rawgit.com/swederik/cornerstoneTools/enh/wlregion/examples/wwwcRegion/index.html

**Screencast:**

windowwidthbyregion


swederik:
Merging this since it seems to work as expected.

PR #47, opened 2015-05-26
WIP: Arrow+Text annotation tool

swederik:
**Work in progress, Don't merge yet** 

Remaining issues
- Text cannot be changed after generation => *Double click/tap?*
- Mouse and touch tools re-use a ton of code, and this is sloppy

This is a simple annotation tool. It allows you to draw an arrow and then asks you for a text annotation to attach. It also works on touch devices.

**Live demo:**
https://rawgit.com/swederik/cornerstoneTools/textAnnotation/examples/annotation/index.html

**Screencast:**
arrowannotation


ra00l:
this is reaaly cool.

I would change 2 things:
1/ draw handles on each arrow ends.
2/ the confirm ... is not a good idea (blocks the UI thread). is better to use a callback.

Editing the text will be troublesome in my opinion. Just let it uneditable; if someone mistakes something, remove and redraw the arrow.

How can I help?

ra00l:
ah, also 
3/ clicking cancel or closing the prompt should also remove the arrow, not keep it

swederik:
Your comments are help enough! Thanks.

I still think editing should be a standard feature, it's frustrating to have to delete things just to change the text. I'll probably merge it without editable text and get back to that later, but your other comments make sense.

swederik:
I fixed your comments @ra00l, thanks. There is one piece of odd behaviour still left to fix before merging:

touchStartCallback and mouseMoveCallback seem to both fire at once when the arrow is drawn, making the end handle stop moving. I don't know how to fix this or why it's occurring...

swederik:
I just disabled touch in the example. I don't know why I can't have both touch and mouse tool enabled at once, but I'll merge this for now and see if I can fix it later.

PR #48, opened 2015-05-27
Batched parallel requests for stack prefetching.

swederik:
This PR is an attempt to improve the stackPrefetch tool.

The old tool waited for each loadAndCacheImage to finish before making the next request. This one works in parallel, as follows:

- The tool sets a maxSimultaneousRequests variable (either from stack size or user-defined)
- The tool is given an initial image index (in the stack, e.g. the middle image of the volume)
- The prefetch function recursively identifies maxSimultaneousRequests/2 images around the initial image until all images are fetched
- Image fetching is done in batches of maxSimultaneousRequests images (rather than, e.g. immediately firing 200 requests) to prevent overloading the server
- Prefetching can be disabled, as before, and the next batch won't be requested.

In my tests on a cheap server running nginx/django the stack prefetching performs substantially faster than it did previously.

One issue that I've noticed is that the UI thread seems to lock up a bit during the prefetching. I am debating converting this to use WebWorkers, but it could also be an issue that my test has 3 distinct elements prefetching at once, and so the settings may not be optimal.

This tool is a bit tricky to test and I would welcome any feedback.


chafey:
The image promise failures are due to a bug in your code, you need to wrap the call to errorHandler() in a function.  Same thing with the done() call to removeFromList().

chafey:
Image #0 is usually the first image displayed in a stack.  In this case, the prefetcher should go from image #0 up as that is the way the user will normally scroll.  Currently this prefetcher will start pulling images from the back of the stack which is not what we want

chafey:
The prefetcher should dynamically change the images it prefetches based on where the user scrolls to.  If you have 1000 images and the user starts at image 0, but jumps to image 500, the prefetcher should adapt and start prefetching around image 500.

chafey:
We want to guard against the case where the cache pushes out images that the user is likely to scroll to because the prefetcher is putting images in the cache the user is less likely to view.  This will happen if the cache can only hold 100 images and the stack has 1000 images in it.

chafey:
We should avoid changing the prefetch behavior for existing users.  If different prefetch behaviors are desirable, we should encapsulate the behavior in a strategy function so users can select which one they want.  See how I did this with the wwwc tool in the wwwc example (user can select between two different wwwc behaviors).

chafey:
If the image loader is doing decompression in the UI thread (which cornerstoneWADOImageLoader does with JPEG2000 images), the interactivity of the viewer will suffer.  Does this explain the UI thread lockups you are seeing or is it also happening with other image types?  The prefetch logic is simple enough that we should be able to keep it in the main thread.  The UI thread responsiveness issue should be resolved in the image loader itself (e.g. decode the JPEG2000 data in a web worker)

swederik:
1 - Ugh, thanks. That deferred bug was a dumb mistake, yeah. Quite annoying, since the functions still ran.

2 & 4 - Yeah user-selectable strategies is a good idea. In some viewers, yes starting at 0 is preferable, but not always. E.g. in 3D volume viewers it's much more useful to display the middle of the stack first. That was the logic I had in mind when I did it this way (hence the prefetching around the initial image). I'll add a strategy choice.

3 - Agreed, that's sensible but hadn't occurred to me at all. Shouldn't be hard to do since we have currentImageIdIndex in the stack tool data

4 - Cache issue I also hadn't really thought about. So the desired behaviour would be:

**When starting** (at image 0 or somewhere else in the stack)
- Prefetch the next X images until the cache limit is reached.

**When the current image is switched:**
- Remove images below the currentImageIdIndex from the cache
- Prefetch the next X images if they aren't already in the cache, until the cache limit is reached.

5 - Yep that would definitely explain it. I haven't experience lockups with the web image loader

PR #49, opened 2015-05-28
Simple viewport rotation tool for mouse/touch

swederik:
Here's a quick rotate tool that also works with touch devices.

In the future I would prefer to have a two-finger touch rotate, though.

**Live demo:**
https://rawgit.com/swederik/cornerstoneTools/enh/rotateTool/examples/rotateTool/index.html

**Screencast:**
rotatetool


swederik:
Merging this because it's so simple.

PR #54, opened 2015-06-11
Updates to arrow annotation. 

swederik:
This PR updates the annotation tool to use a better callback system rather than the 'prompt' it used before.

**Changes:**

- Removed prompt in favour of dialog callback.
- Added drawHandles and arrowFirst options to allow customization
- Switched from setToPixelCoordinateSystem to pixelToCanvas for placing tool
- Switched to using tool configuration to manage callback and other options

**Live demo:**
https://rawgit.com/swederik/cornerstoneTools/bf/annotation/examples/annotation/index.html


swederik:
Seems to work fine, merging this.

Issue #32, opened 2015-05-07
Grabbing the vertex point of the angle measurement tool "breaks the V"


cancan101:
Grabbing the vertex point of the angle measurement tool "breaks the V":
![image](https://cloud.githubusercontent.com/assets/51059/7509876/f50acff6-f44a-11e4-8076-2f75bb583321.png)

chafey:
This behavior is by design as this tool is intended to support cobb angles.  I can see how this behavior is not always desired and am open to a PR that would allow the V not to be broken (either via an option or another tool entirely)

swederik:
Closing since #67 was merged

PR #67, opened 2015-06-17
Simple Angle tool

swederik:
This PR adds a simple angle measurement tool. It just uses three points and cannot be separated as in the Cobb angle tool. This addresses #32.

**Live demo:**
https://rawgit.com/swederik/cornerstoneTools/angleTool/examples/angleTool/index.html

**Issues:**
- What do we call this?
- Should I rename the other tool "cobb angle"?
- Should I update the allImageTools example?


swederik:
Merging for now, we can deal with the smaller stuff later

PR #73, opened 2015-06-18
missing line added

onluiz:
Hi Chris. I found out, at version 0.6.2 of cornerstoneTools that he is missing this line at annotation.js:

cornerstoneTools.getConfiguration = getConfiguration;

The method touchDragCallback in wwwc.js was not working because of this. It have a call for getConfiguration:

var config = cornerstoneTools.getConfiguration();

I hope it can help, I tried to post it at google groups but I'm not sure if there is the right place for this kind of thing.

### DIFF
--- a/src/imageTools/annotation.js
+++ b/src/imageTools/annotation.js
@@ -596,5 +596,7 @@ var cornerstoneTools = (function ($, cornerstone, cornerstoneMath, cornerstoneTo
         deactivate: deactivateTouch
     };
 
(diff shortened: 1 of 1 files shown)


swederik:
Hey,

Thanks for the report. That's actually not quite the right fix, but it did point me to the actual problem. I just fixed it, see commit: https://github.com/chafey/cornerstoneTools/commit/7881fc39a3ec9797bc69338019989e8a1a4d9a90

And yes I meant to use @onluiz in the commit message....

Closing this because it's now fixed.

PR #79, opened 2015-06-24
Prefetch fixes to better handle already loaded images

jmhmd:


swederik:
Thanks a lot these fixes are very helpful.

PR #80, opened 2015-06-24
Add arrow tool

jmhmd:


swederik:
Could you add an example (or add this to the 'All Image Tools' example?). I think maybe we should also unify the drawArrowhead function with the one from the arrow annotation tool.

Thanks

jmhmd:
Oh wow, you're right- I had initially put this together before the current arrow tool had been added. I'll have to look into it and see if there is anything to add to the current functionality.

swederik:
No worries. If you're still interested in a pure arrow tool it's easy to make one without the annotation (or perhaps add a config setting which will disable the actual text portion of the arrowAnnotate tool). Might work to just set the getTextCallback to do nothing.

PR #81, opened 2015-06-24
Add progress/scroll indicator tool

jmhmd:
1. Draw bar at bottom of image showing scroll position of stack, as well as which images in the stack have loaded. (`scrollIndicator`)
2. While image is loading into element, show percent loaded (`loadingIndicator`)


swederik:
Thanks for the PR, they seem like useful tools. Could you add an example as well (you can mostly copy/paste the stack scroll / stack progress examples)?

PR #83, opened 2015-07-01

Rionmonster:
Currently any of the available synchronizers do not have the ability to capture event data from events that trigger them. This functionality could be useful if you need to capture which direction a mouse wheel is scrolling or just to generally pass in some additional data.

Basically, if you wanted to implement this into an existing synchronizer, you would simply need to add your data into a trigger that targets an event that your synchronizer is listening to. For example, if you instantiate a synchronizer as follows: 

```
var synchronizer = new cornerstoneTools.Synchronizer("CornerstoneNewImage", cornerstoneTools.stackImageIndexSynchronizer);
```

This synchronizer will be triggered when the `CornerstoneNewImage` event is called. So if you had an existing event that you wanted to pass along to the synchronizer, you could simply trigger it as follows, passing in the data you want to access in the synchronizer: 

```
$(element).trigger("CornerstoneNewImage", { foo: 'bar' });
```

Then simply update your existing synchronizer to support an additional `eventData` parameter as seen in this custom synchronizer example below: 

image

And you'll be able to access the eventData within the synchronizer to use as you please as seen by the `direction` parameter that was passed in which denotes the direction that a mouse wheel was scrolled over a viewport.


swederik:
It looks good, but you'll need to rebase from the master branch again. Sorry that's my fault I committed some stuff directly to master.

Thanks for the contribution!

Rionmonster:
Closing due to more recent build.

PR #84, opened 2015-07-01

Rionmonster:
Currently any of the available synchronizers do not have the ability to capture event data from events that trigger them. This functionality could be useful if you need to capture which direction a mouse wheel is scrolling or just to generally pass in some additional data.

Basically, if you wanted to implement this into an existing synchronizer, you would simply need to add your data into a trigger that targets an event that your synchronizer is listening to. For example, if you instantiate a synchronizer as follows:
```
var synchronizer = new cornerstoneTools.Synchronizer("CornerstoneNewImage", cornerstoneTools.stackImageIndexSynchronizer);
```
This synchronizer will be triggered when the CornerstoneNewImage event is called. So if you had an existing event that you wanted to pass along to the synchronizer, you could simply trigger it as follows, passing in the data you want to access in the synchronizer:
```
$(element).trigger("CornerstoneNewImage", { foo: 'bar' });
```
Then simply update your existing synchronizer to support an additional eventData parameter as seen in this custom synchronizer example below:

image

And you'll be able to access the `eventData` within the synchronizer to use as you please as seen by the direction parameter that was passed in which denotes the direction that a mouse wheel was scrolled over a viewport.

You can [see an example that overrides the previous stackImageIndexSynchronizer here](https://rawgit.com/Rionmonster/cornerstoneTools/master/examples/stackImageIndexSynchronizer/indexWithSynchronization.html), which captures mouse events.


swederik:
Hi. Sorry I made changes to master again so you'll need to rebase once again (you don't need to make a new PR for this, just use git push -f)

A few things:

- What is the purpose of exampleCornerstoneToolsWithScrollEvent? The built version in /dist should be up to date...
- Can you move indexWithSynchronization.html to a separate folder and just title it index.html, that would keep it consistent with the other examples.
- Don't forget to add a link to the new example in examples/index.html

Thanks!

Rionmonster:
Hi Erik,

Just to address your questions : 

**What is the purpose of exampleCornerstoneToolsWithScrollEvent? The built version in /dist should be up to date...**

I wasn't sure if we wanted to add this to the existing cornerstoneTools as a change was necessary to trigger the appropriate event for the synchronizer for the appropriate scrolling events : 

```
$(eventData.element).trigger("CornerstoneImageScroll", {direction :images});
```

Would you recommend making this change for all of the scrolling events and simply including it in? I wasn't sure of this, so I figured that I would ask.

**Can you move indexWithSynchronization.html to a separate folder and just title it index.html, that would keep it consistent with the other examples.**

Sure.

**Don't forget to add a link to the new example in examples/index.html**

Will do, I'll make those changes shortly after I hear back from you regarding the first item.

swederik:
Hey,

Great. Umm I don't really see any problem with including the CornerstoneImageScroll event, so you might as well just include it in cornerstoneTools.

Thanks

swederik:
Hi Rion,

I'm going to close this PR. I just cherry-picked your commits and will merge them soon in a separate PR: #108. The reason is that I had already added a stack scroll event recently so some of your changes became unnecessary.

PR #87, opened 2015-07-04
Change pointNearHandle to compare distance based on screen pixels

jmhmd:
- Previous implementation compared distance based on image pixels, which changed the behavior for different image resolutions. For example, a high resolution chest radiograph will not "snap" correctly as 5 image pixels translates to <1 screen pixel when zoomed to fit.

### DIFF
--- a/src/imageTools/freehand.js
+++ b/src/imageTools/freehand.js
@@ -72,9 +72,13 @@ var cornerstoneTools = (function ($, cornerstone, cornerstoneMath, cornerstoneTo
             return;
         }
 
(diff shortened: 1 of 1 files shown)


swederik:
Ah great, yeah I was planning to fix this exact problem this week. I'll take a look at this and see about migrating all the tools.

Thanks

swederik:
I'm closing this because #92 handles the same issue for the entire library using canvas points instead of image points. Thanks for your help though!

jmhmd:
No worries, this is way better to have all the tools fixed, thanks!

PR #91, opened 2015-07-09
WIP: Improvements to stackPrefetch tool

### LINKED ISSUE #90: range function in src/stackTools/stackPrefetch.js endless.
I was fiddling around and got my app locked up.

the range function needs assertions because it can end up in a long loop.

suggested patch:

    function range(lowEnd, highEnd) {
        // Javascript version of Python's range function
        // http://stackoverflow.com/questions/3895478/does-javascript-have-a-method-like-range-to-generate-an-array-based-on-suppl

        //don't allow undefined / null to be passed to this function.
        lowEnd = lowEnd||0;
        highEnd = highEnd||0;
        var arr = [],
            c = highEnd - lowEnd + 1;
        c=c>0?c:0;//lets not do this.
        while ( c-- ) {
            arr[c] = highEnd--;
        }
        return arr;
    }

swederik:
**Don't merge yet. I'm just putting this here for easier review**

This PR reworks the stackPrefetch tool to cover some of the recent issues that have been brought up.

The prefetch behaviour is now as follows:

1. The user's browser name and version are determined, and the maxSimultaneousRequests variable is set (e.g. N images, defaults to 6)
2. Starting from one image above currentImageIdIndex, the prefetcher grabs N images above the current image asynchronously
3. As soon as any of the asynchronous functions finishes loading an image, it grabs the next highest (lowest) uncached image relative to the currentImageIdIndex. The choice of highest / lowest is determined by any previous scrolling behaviour (stackPrefetch tool data is now updated in scrollToIndex).
4. When scrolling, the prefetcher is temporarily disabled. As soon as a new image is displayed (on CornerstoneNewImage), after 50ms the stack prefetcher is re-enabled. This prevents the user from fighting with the prefetcher for bandwidth.
5. If the user chooses, they can set config.maxSimultaneousRequests. This overrides the default browser value.
6. If the CornerstoneImageCacheFull event is fired, all further prefetching is stopped (**Depends on https://github.com/chafey/cornerstone/pull/25**)
7. If the CornerstoneImageCachePromiseRemoved event is fired, the imageIdIndex is re-added to the indicesToRequest array so it can be retrieved again if the user changes to a nearby image

**New:**
8. If the image is scrolled, all current prefetch-initiated requests are manually aborted. This ensures that the network priority is focused on the scrolling behaviour

**Issues:**
- Number of maximum requests does not yet take into account multiple elements with prefetching enabled.
- Only images above currentImageIdIndex will be retrieved, by default. Some type of strategy options need to be added for cases where currentImageIdIndex begins above 0 (e.g. in the middle of the stack)
- I have seen high "Waiting" times before image downloading begins. It's unclear where this is coming from.

**Live demo using cornerstone-test-image-loader (check the console):**
https://rawgit.com/swederik/cornerstoneTools/fixPrefetch/examples/stackPrefetch/index.html

This PR also closes #90.


swederik:
Cl

PR #92, opened 2015-07-10
Transitioning tool/handle selection to canvas points for entire library

swederik:
This switches all pointNearTool and similar methods to use canvas points rather than image points.

Try zooming in / out very far from the image, you should be able to select handles / tools normally. In the current version of cornerstoneTools, tool selection becomes very difficult at different zoom levels or with high-resolution images. 

**Live demo:**
https://rawgit.com/swederik/cornerstoneTools/canvasPoints/examples/allImageTools/index.html

*Please test this and report any bugs! It changes every image-specific tool in the library and I don't want to merge it and break things*

**Update: I've tested all of the tools and fixed a number of bugs, this seems ready to merge to me**


swederik:
Merging since I haven't run into any bugs yet...

PR #97, opened 2015-07-22
Added frame of reference tool state manager

swederik:
Just adds another tool state manager. This is useful for building tools which should only be displayed on images preset within a specific frame of reference. One example use case is to share 3d information between stacks of images.


swederik:
Merging because it doesn't really affect anything

PR #100, opened 2015-07-30
Added shadow: boolean config for length and angle tools

leonardorame:
In some very light image areas (bones for example) the measurement cannot be seen properly, so, 
I created this simple config that brings shadow to length and angle tools.

Usage:

        cornerstoneTools.length.setConfiguration({shadow: true});
        cornerstoneTools.angle.setConfiguration({shadow: true});


swederik:
Sounds nice. Do you think it's worth including for all tools? Any chance you could include a checkbox in one of the examples so that other users can see it in action?

Thanks for the contribution!

leonardorame:
Yes, I'll add the example.

I think all tools should have a uniform graphic representation and every one must try as much as possible to stand out above the underlying image, one way is by adding this shadow (Fuji Synapse PACS uses something like this and it looks really nice). Having said that, I'd focus on defining one set of general rules, then try to apply it to each tool.

leonardorame:
I added a checkbox to AllImageTools example.

swederik:
Nice! The shadow looks good. Really good actually, we should consider turning it on by default when the textbox has no background. Or just making it an option for all on-canvas text.

**Without:**
screenshot 2015-07-31 11 56 01

**With shadow:**
screenshot 2015-07-31 11 56 07

Tested via rawgit at:

https://rawgit.com/leonardorame/cornerstoneTools/master/examples/allImageTools/index.html

Could you add one more small thing before I merge it? Just add something like:

```` javascript
cornerstone.updateImage(element);
````

inside the checkbox click function so when the checkbox changes the image onscreen updates. 

You may also want to change from ```` javascript .click(function() {})```` to ```` javascript .on('change', function() {})```` to handle cases where the checkbox is triggered via the keyboard (e.g. using the spacebar). This is done in the textMarker example: https://github.com/chafey/cornerstoneTools/blob/master/examples/textMarker/index.html.

Thanks!

**Comment by leonardorame** [comment]
Done, please take a look at it.

**Comment by swederik** [comment]
Looks good! My bad though, I just merged another PR. Could you rebase this PR on top of master? Then I'll merge

**Comment by leonardorame** [comment]
I don't know how to, just git rebase master?

**Comment by swederik** [comment]
First you need to make sure you have this repo as a remote in git:

This adds it with the name "upstream"
```` bash
git remote add upstream git@github.com:chafey/cornerstoneTools.git
````

Then fetch, checkout your branch, and rebase with upstream master
```` bash
git fetch upstream
git checkout master (or whatever branch you want, but according to your PR it's master)
git rebase upstream/master
````

You may need to iron out any conflicts. Git will tell you which files have conflicts and then you have to open them and decide which lines from which branch you should keep.

To resolve any conflicts in generated files (like dist/cornerstoneTool.js or dist/cornerstoneTool.min.js), just run ````grunt```` again, and then use ````git add theFileName```` and ````git rebase --continue```` as necessary.

**Comment by leonardorame** [comment]
Done, please check it out.

File: dist/cornerstoneTools.js [hunk 1]
```diff
@@ -1,4 +1,8 @@
-/*! cornerstoneTools - v0.6.2 - 2015-07-31 | (c) 2014 Chris Hafey | https://github.com/chafey/cornerstoneTools */
+<<<<<<< HEAD
```
swederik: Everything looks good except for this! Looks like a rebase / merge artifact. Try running grunt again and committing the result, then it looks good to merge!

leonardorame: Done. Leonardo M. Ramé
http://leonardorame.blogspot.com

swederik:
Great! Merging :-)

PR #103, opened 2015-08-06
Request pool behaviour for prefetch / interaction

swederik:
**Changes:**
- All prefetch or scroll interaction requests are put into a stack.
- The system takes requests in order of priority (interaction first, then thumbnail, then prefetch)
- When image changes, prefetch stack is cleared and re-added based on what is left in the stack that still needs be grabbed (and scrolling direction)
- When user scrolls, all interaction requests are erased and the current request gets top priority

Replacing #91 with this, since that branch is now out of date.


swederik:
I think this has been tested enough to merge now!

PR #105, opened 2015-08-22
Plane/plane intersection for calculating reference lines

swederik:
**Changes:**
- Calculates Plane/Plane intersection between two slices in order to draw reference lines robustly

**Method:**

1.  Identifies Origin and Direction of intersection Ray for infinite planes in the given slices' orientation
2.  Turns intersection Ray into Line3 by adding arbitrarily far-away endpoint
3.  Gets all intersection points between this line and the four sides of the reference image slice
4.  If there are two intersection points, convert these to image points on the target plane, and draw the reference line between them. If not, do nothing.

**Live demo:**
https://rawgit.com/swederik/cornerstoneTools/plane2plane/examples/referenceLineTool/index.html

**Notes**
- Depends on https://github.com/chafey/cornerstoneMath/releases/tag/0.1.2 so make sure to run:

````
bower install
````

and then ````grunt```` to make sure you have the latest versions if you want to run the examples.


idmpacs:
The plane2plane reference line example seems to be removed. Is there any possibility that I can get code for it ?

swederik:
Of course! I guess what you're looking for is in cornerstoneMath now:

https://github.com/chafey/cornerstoneMath/blob/master/src/plane.js#L180

PR #108, opened 2015-09-01

swederik:
Rebased and updated @Rionmonster's PR #84. Still need to check that the example works.

@Rionmonster Could you verify that this is working as you intended?
https://rawgit.com/swederik/cornerstoneTools/scrollSync/examples/stackScrollSynchronizer/index.html


swederik:
Seems fine in my testing, so I'll merge this too

PR #111, opened 2015-09-08
Synchronizer improvements & Stack Image Position Offset Synchronizer

swederik:
**Changes**
- Adds stackImagePositionOffsetSynchronizer to allow for linking viewports with the current differences between their locations in the stack. This is similar to the imagePositionSynchronizer, which displays the closest image in the stack to the source image's location.
- Synchronizers now remove elements that have been disabled
- Synchronizers now have a 'destroy' function to remove all reference to any source/target elements
- Synchronizers now have setHandler and getHandler to change handlers after initialization

Instead, this synchronizer records the initial starting positions and difference vectors between each image. It then finds the target image in the stack closest to the initial image + the initial difference.

Some updates were necessary to the basic Synchronizer functions in order to facilitate this behaviour.

*To be honest, the example attached to the stackImagePositionOffsetSynchronizer really sucks since we only have a few example images. I've been testing this with full stacks though and it seems to work well.*


swederik:
I might want to remove getDistances from Synchronizer later on, but for now this works so I'll merge it.

PR #114, opened 2015-09-11
Fix equation for calculating if point is in ellipse

jmhmd:
Current equation used in pointInEllipse function has some weird behavior and does not accurately approximate the ellipse (example [here](https://rawgit.com/jmhmd/cornerstoneTools/bugfix-circle-example/examples/highlight/ellipse.html)).

Proposed fix to use a new equation which seems be more accurate (example [here](https://rawgit.com/jmhmd/cornerstoneTools/bugfix-circle-example/examples/highlight/ellipseFixed.html)).

Source of new equation: http://math.stackexchange.com/a/76463

### DIFF
--- a/src/imageTools/ellipticalRoi.js
+++ b/src/imageTools/ellipticalRoi.js
@@ -66,16 +66,8 @@
             x: ellipse.left + xRadius, y: ellipse.top + yRadius
         };
 
-        /* This is a more general form of the circle equation
-         *
-         * X^2/a^2 + Y^2/b^2 <= 1
-         */
-
-        var normalized = {
-            x: location.x - center.x, y: location.y - center.y
-        };
-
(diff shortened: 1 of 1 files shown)


swederik:
Hello!

Thanks for the PR. The equation looks fine and I have no problem merging it, but the highlighting in the examples still looks wrong to me (I guess this is independent of the PR).

This is what I get for the first link:
screenshot 2015-09-14 11 25 20

And this is what I get for the second:
screenshot 2015-09-14 11 25 23

swederik:
Whoops, just noticed that it looks like the equation you're using and the current equation seem to be exactly the same, Chris just did the subtraction (to create 'normalized') before you. There is indeed, though, a typo in his version (normalized.x*normalized.y) in the first term should be (normalized.x*normalized.x). I'll close this and just fix that one character. Thanks for the report!

swederik:
This issue should now be fixed in 0.7.4.

I've described a similar issue here: https://github.com/chafey/cornerstoneTools/issues/115 though.

jmhmd:
Looks like that fixes it for me too- thanks!

PR #117, opened 2015-09-17
Bug fixes for mouseButtonTools with only one handle

swederik:
This fixes an annoying bug for tools relying on mouseButtonTool that only had a single handle (i.e. textMarker, probe).

It also fixes some weird tool movement issues related to not-unbinding event handlers before using moveAllHandles.

Previously, you had to click twice to place one of these tools. Now you can once again place it in one click. This was a side effect of the recent changes allowing click-to-place behaviour for other tools (e.g. three clicks to place angle tool)


swederik:
Seems fine in my tests, so I'll merge this

PR #131, opened 2015-12-04
Fixing requestPoolManager "race condition"

hroenick:
- startAgain's timeout logic should not set the awake flag to false.
   Otherwise the flag may be set to false when still exists requests
   in pools' array and the number of request counter is equal to the maxValue.
   This condition was met when performing the prefetch of series images.


swederik:
Thanks for the PR! Could you rebase on top of the current master, though, so I can use Github to merge it?

hroenick:
Sure!

swederik:
Closing since I just merged the other identical PR #132

PR #132, opened 2015-12-08
Fixing requestPoolManager "race condition"

hroenick:
- startAgain's timeout logic should not set the awake flag to false.
   Otherwise the flag may be set to false when still exists requests
   in pools' array and the number of request counter is equal to the maxValue.
   This condition was met when performing the prefetch of series images.

- pull request from master, as requested: https://github.com/chafey/cornerstoneTools/pull/131#issuecomment-162216556


swederik:
Thanks for your contribution! In the future you don't need to make a new branch when rebasing, though.

hroenick:
I sure need to understand the git usage better. Too many years working with svn and I am still making the transition to git. Thanks for the heads up!

Issue #118, opened 2015-09-22
Allow cache prevention for image stacks


swederik:
As discussed on the Google group, we need a way to mark certain stacks as 'uncacheable' so that developers can build sites that depend on dynamic image fusion.

Something like this:

```` javascript
var stack = {
    currentImageIdIndex : 0,
    imageIds: imageIds,
    preventCache: true
};
cornerstoneTools.addToolState(element, 'stack', stack);
````

If this flag is set, all tools (e.g. Crosshairs, synchronizers...) that use loadAndCacheImage should use loadImage instead.

**Relevant threads:**
https://groups.google.com/forum/#!topic/cornerstone-platform/mst1b6ZB4T8
https://groups.google.com/forum/#!topic/cornerstone-platform/FHM4tWQcy8s

PR #133, opened 2015-12-10
Add a preventCache flag to image stacks

swederik:
This is meant to prevent certain stacks from being cached. I haven't had time to test it yet, so don't merge it. Addresses issue #118.


hiscodeness:
Looks good, I tested this PR against my current implementation and didn't notice any issues. 

In the code, I think it would be nice if repetition of 

    var loader;
    if (stackData.preventCache === true) {
          loader = cornerstone.loadImage(stackData.imageIds[newImageIdIndex]);
    } else {
          loader = cornerstone.loadAndCacheImage(stackData.imageIds[newImageIdIndex]);
    }
    loader.then(function(image) { ...

could be replaced with something like:

    cornerstone.loadImage(stackData.imageIds[newImageIdIndex], stackData.enableCache)

This would require a change in the core cornerstone API. Another option would be setting a loader function dynamically to stackData when the tool state is initially created. All tools using stackData would then call `stackData.loadImage(imageId, preventCache)` instead of the core cornerstone image loader functions.

swederik:
Yep I agree it's annoyingly repetitive. I will discuss with @chafey about whether or not to change Cornerstone's loadImage API. The other idea is interesting but seems unnecessarily confusing to me.

Thanks for your help!

swederik:
I'm just going to merge this for now. If / when we change the loadImage API we can revisit these changes.

chafey:
FYI - My gut tells me we need to rethink the design approach to caching a bit, but I haven't had the time to do that yet, hence my lack of response.  Lets stick with the current API's for now and revisit this as part of a 2.0 type effort

PR #148, opened 2016-05-06
fixed multitouch panning

freund17:
The "panmove"-Event only fires when using a single finger.
More fingers may be added while panning.

But if the panning ist started with multiple fingers, the "panmove" event doesen't fire.

Added "rotate.recognizeWith(pan);" to fix this.
Now the "panmove" event is fired when starting with multiple fingers.


swederik:
Looks good to me! Sorry for the long delay.

Thanks for your contribution!

Erik

PR #151, opened 2016-06-30
Update Shadow in RectangleRoi

SevenM7:


swederik:
Hello!

Could you also build with grunt and commit cornerstoneTools.js and cornerstoneTools.min.js?

Thanks for the contribution!

PR #153, opened 2016-07-12
Fix 'element not enabled' errors.

mscheper:
The fact that the errors were occuring for the currentTarget made me suspect I was somehow enabling the wrong element in my application, but I've done a lot of testing and I'm 99% sure this is not actually
happening.

For more information, please see this post: https://groups.google.com/forum/#!msg/cornerstone-platform/uU09JPyWJTU/1NhQXaMeBQAJ

### DIFF
--- a/src/stackTools/stackPrefetch.js
+++ b/src/stackTools/stackPrefetch.js
@@ -221,7 +221,7 @@
         // When the user has scrolled to a new image
         clearTimeout(resetPrefetchTimeout);
         resetPrefetchTimeout = setTimeout(function() {
-            var element = e.currentTarget;
+            var element = e.target;
             prefetch(element);
         }, resetPrefetchDelay);
     }


swederik:
Hi Michael,

Thanks for the detailed report and PR. I'm surprised currentTarget isn't correct, since the event listener is definitely attached to a viewport element as well. In the OHIF Viewer when I place a breakpoint at that line I see that currentTarget and target are identical.

I'll merge this because I don't see anything that it could break, but I wonder what is special about your case that could be breaking it? You just have two viewports?

Thanks!

Erik

mscheper:
Hi Erik,

Yes, I have two viewports, as I mention in the [forum post I linked to](https://groups.google.com/forum/#!msg/cornerstone-platform/uU09JPyWJTU/1NhQXaMeBQAJ) above. I found that when the error was logged, the `currentTarget` is an ancestor of the two viewports. This honestly doesn't make sense to me either, but I'm glad you agree that my change won't break anything.

Thanks!

PR #156, opened 2016-07-20
Add basic paintbrush tool & example

swederik:
Special thanks to @zelle-github

Here is a first pass at a paintbrush tool that will hopefully be extensible for use by the community. It uses the layering changes made in https://github.com/chafey/cornerstone/pull/68

### Demo

https://rawgit.com/swederik/cornerstoneTools/overlayTool/examples/segmentationBrush/index.html


zelle-github:
One thing that would be very important is transparency. For the segmentation overlay as well as for the painting. I think it is important to be able to see what is behind the drawing. Maybe we can set the alpha channel in the overlay configuration?

One question then would be, how the transparency should behave if multiple segmentation are placed over each other. It would then be nice if only the top layer would be shown, otherwise transparency would decrease with the number of layers over each other. But that could be a little complicated to implement.

zelle-github:
Also one important thing we would need is to show the current position of the pointer. Ideally it would be just a darker border of the pointer circle. Something like this:
brush

zelle-github:
Image rendered event for all layers

It would be useful to have the image rendered event for all layers separate in stead of, like now, having it only for the visible canvas.

I am asking because for border detection I need the currently shown image as an array. But, as the border detection is only based on the image but not on the drawing, I need only the image itself. Every time the image is changed (because of zooming for example), the array of the canvas has to be updated.

Currently, the image rendered event is fired every time the brush is shown and moved as the visible canvas is updated every time. That means, my canvas array is also updated every time.

Swederik suggested to not fire the events for all layers in general. The idea is to enable layer dependent events only for activated overlay tool.

zelle-github:
Canvas sizes

The hidden canvases should all have the same size as the visible canvas. Currently they are having the same size as the image. This is extremely confusing ;-)

PR #160, opened 2016-08-17
now using delta-value from hammerJS

freund17:
The change ist mainly noticable when combining multitouch-panning with pinch-zooming.
Now the zoom will work more as expected by having the center of the zoom in the middle between your fingers instead of having it under whatever Finger you first touched the screen with.


swederik:
Merged this here: https://github.com/chafey/cornerstoneTools/commit/5b0558aa2ee5853ddf90c48aa978510eb081eba7

Thanks!

PR #180, opened 2017-01-26
Purge image cache issue (stackPrefetch)

lscoder:
stackPrefetch should not trigger exceptions for disabled elements removed from cache


swederik:
Rebased and merged here: https://github.com/chafey/cornerstoneTools/commit/8d5fafd65a2d648ae4c016c0a504b3ceeb59c76c

Thanks!

PR #181, opened 2017-02-01
Fix #178 panning with uneven pixel spacing.

### LINKED ISSUE #178: Pan tool does not take non-square pixels into account.
When panning with non-square pixels the image does not correctly follow mouse mouvements.

I think fixing this would required weighting [X and Y translations](https://github.com/chafey/cornerstoneTools/blob/master/src/imageTools/pan.js#L21-L22) with widthScale heightScale found [here](https://github.com/chafey/cornerstone/blob/5fcbcaa3c7151f8f3b432811bce45443a9ea1a49/src/internal/calculateTransform.js#L26) instead of just scale. However, widthScale and heightScale are not available in the mouse event.

@swederik, thoughts?

jpambrun:


swederik:
Thanks!

PR #183, opened 2017-02-09
Bugfix for interactive mode in touch tools

sirrrich:
It is not possible at the moment to make touch tools interactive but not creatable.
The 'CornerstoneToolsTouchStart' should be set to on if the tools is deactivated.
When a measurement is modified the 'CornerstoneToolsTouchStart' event should be used instead of the 'CornerstoneToolsTouchStartActive' to not enable the creatation of new objects


swederik:
Nice catch! Thanks for the contribution!

PR #191, opened 2017-03-08
Bugfix for 'CornerstoneToolsMouseClick'

sirrrich:
Variable is not initialized, so the first click event is not send


swederik:
Another nice catch, thanks @sirrrich!

Issue #186, opened 2017-02-14
requestAnimationFrame instead of setIinterval


oankarberg:
Hi,

Shouldn't requestAnimationFrame be used instead of setInterval when playing a clip of stacks? setInterval doesn't consider the rendering pipeline of the browser and will therefor affect the performance. The browser will automatically take care of when and how its best to run requestAnimationFrame according to the browsers rendering pipeline. 
I ran the same CT stacks with requestAnimationFrame and setInterval. SetInterval showed some "Long Frame" (about 30 fps instead of 60) during garbage collection. 

requestAnimationFrame statistics, continously 60 fps: 
https://gyazo.com/6c24f5c9d9cd25cecbc855f4f93894b2

setInterval, bumps of 30 fps close to garbage collection: 
https://gyazo.com/c5a13b0f857ec538b0f8550d2188c03c

jpambrun:
Yes, using requestAnimationFrame across the board is planned for the near term.

swederik:
Does this seem like a suitable approach? https://github.com/chafey/cornerstoneTools/pull/195

I used some drop-in replacements for setTimeout and setInterval that use requestAnimationFrame (https://gist.github.com/joelambert/1002116).

dannyrb:
It looks like this is still in the works.

https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/stackTools/playClip.js#L245-L255

Any news, @swederik?

PR #195, opened 2017-04-05
Use requestAnimationFrame for clip animations (#186)

swederik:
Switch to requestAnimationFrame for playClip and timeSeriesPlayer.


jpambrun:
This is a lot of code to maintain and there is a bit of duplicated functionality. But I admit I did not thoroughly review it.

I'm wondering if we could just register a CornerstoneImageRendered callback and test date.now() and swap the image for the next render if the delay has passed (or is very near to).

Whatever we do, it's gonna be very hard to enforce a precise payback rate as rendering is locked to the display refresh rate. I assume it is not OK to skip frames, so >60 fps is out of the question. If the render time is >16 ms, 60 fps would now even be possible. The CornerstoneImageRendered callback method would prohibit skipped frames. If the selected rate exceeds the refresh rate or the render rate, the outcome would be the best efforts as regulated by RAF(). 

As a bonus, we could probably implement that in less than 20 LOC.

swederik:
Closing because it will be better to implement the solution mentioned by @jpambrun above.

PR #196, opened 2017-04-07
The first element of stackPrefetch.indicesToRequest isn't being prefetched

lscoder:


swederik:
Just merged this with a cherry-pick commit. Thanks!

PR #198, opened 2017-04-24
LengthTool now uses PixelSpacing provided by MetaDataProvider.

leonardorame:
Hi Erik, please review this PR and push to master if you consider it ok.


swederik:
Hi Leonardo,

Maybe we can add a fallback like this? 

````javascript
var rowPixelSpacing = imagePlane.rowPixelSpacing || imagePlane.rowImagerPixelSpacing || eventData.image.columnPixelSpacing;
````

I just worry that some people won't have imagePlane metadata, so this will break.

I think you will also need to rebase on top of master again, sorry about that.

leonardorame:
Done, could you check this out?.

leonardorame:
Hi Erik, please, could you commit this?.

swederik:
Sorry this took forever. Should be merged now.

PR #199, opened 2017-05-03
creating an object to attach the enabled property to

yungdaht:
You can't attach a property to an undefined object.


leonardorame:
Hi @yungdaht, where can this be used, could you give a sample use-case?.

yungdaht:
I ran into this issue a few times with my current application.  I don't remember the exact execution path that got me there, but the only way that I could resolve the problem was to make the changes that are in this PR.  I figured since it was a valid bug (unless I'm misunderstanding something in regards to applying values to undefined properties), I would submit the request.

swederik:
Hi Shaun,

I merged this via the command line because I screwed it up with the Github web interface...

Thanks for you contribution!

PR #203, opened 2017-05-22
Scroll error event

kofifus:
added a 'CornerstoneNewImageFailed' event sent when scrolling to the stack failed. This is useful to show an error message to the user in that case.


swederik:
Hi @kofifus,

What about using the errorLoadingHandler? That's how we currently display error messages in OHIF.

https://github.com/kofifus/cornerstoneTools/blob/6ace7bbad2f36468758b6a4ae2528fe30b72583f/src/util/scrollToIndex.js#L48

kofifus:
OK I just had a look, I don't actually remember why I added it, I am not using it ATM.

swederik:
OK thanks anyway!

PR #204, opened 2017-05-23
Throttle stack scrolls

kofifus:
Using underscore's throttle function, stack scrolls are throttled so that there is no more than one scroll every 50ms. This prevents too many load requests when the user scrolls the stack.

A better solution would be to throttle only when the images are not in the cache but I am not sure how to fit that into the existing framework.

### DIFF
--- a/src/util/scroll.js
+++ b/src/util/scroll.js
@@ -46,7 +46,7 @@ function throttle(func, wait, options) {
     };
 
     return throttled;
-};
+}
 
 
 function unthrottledScroll(element, images) {
@@ -66,4 +66,4 @@ function unthrottledScroll(element, images) {
   scrollToIndex(element, newImageIdIndex);
 }
 
-export default = throttle(unthrottledScroll, 25);
+export default throttle(unthrottledScroll, 25);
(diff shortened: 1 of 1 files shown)


freund17:
This prevents too many load requests when the user scrolls the stack.

Wouldn't it then make more sense to debounce `loadImage` while the user is scrolling?

kofifus:
Possibly, the best would be to only throttle loading images which are not in the cache I'm not sure how involved that would be ...

But I didn't want to touch loadImage cause I wasn't sure of all the scenarios it may be called ...

freund17:
To find out if a image is in the cache, you just ask the cache:

```js
const cachedImage = cornerstone.imageCache.imageCache['imageId']

if (cachedImage && cachedImage.loaded) {
  // image in cache and fully loaded
  const image = cachedImage.image // access the image
}

if (cachedImage && !cachedImage.loaded) {
  // image currently beeing loaded
}

if (!cachedImage || !cachedImage.loaded) {
  // image not jet loaded
}
```

For the content of `cachedImage` reference:
https://github.com/cornerstonejs/cornerstone/blob/master/src/imageCache.js#L80

The `image`-key gets added a bit later:
https://github.com/cornerstonejs/cornerstone/blob/master/src/imageCache.js#L99

dannyrb:

PR #205, opened 2017-05-23
Fix linux build

zachasme:
There were a few imports that did not correspond to their filenames, which is a problem on case-sensitive operating systems. This pull request renames them according to the case of their exports, and also adds `lodash` to dev dependencies which is required in the build process.


swederik:
Thanks!

PR #212, opened 2017-06-07
Add a canCreate option to activate

kofifus:
This PR adds a canCreate option to activate (defaults to true).

If canCreate is false, then the tool is activated (visible+interactive) but cannot create.

The common scenario for this is starting all image tools with canCreate=false except for one (the current tool) ie length without canCreate=false

The result will be that all existing angles/roi/etc can be manipulated (moved etc) but drag+down where there's no annotation will create a new length.

This is a much more natural user experience.

If needed I can also update the allTools example.

### DIFF
--- a/src/imageTools/mouseButtonTool.js
+++ b/src/imageTools/mouseButtonTool.js
@@ -22,7 +22,8 @@ export default function (mouseToolInterface) {
     }
 
     const eventData = {
-      mouseButtonMask: mouseEventData.which
+      mouseButtonMask: mouseEventData.which,
+      canCreate: true
     };
 
         // Associate this data with this imageId so we can render it and manipulate it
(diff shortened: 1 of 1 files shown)


swederik:
I'm not sure I see why we need this option. If you just 'deactivate' a tool I think you can interact with it but not create new instances.

Try it out here: https://rawgit.com/chafey/cornerstoneTools/master/examples/angleTool/index.html

kofifus:
OK I see ... but there is something wrong I think which confused me

If you go for example to the allTools example (https://rawgit.com/chafey/cornerstoneTools/master/examples/allImageTools/index.html)

and create an eleptical roi, then when you hover over it it will get highlighted.

If you switch to the length tool and create a length, then hovering over the length will highlight it but hovering over the roi will not highlight it although it can still be interacted with.

(I can close this PR and file this is an issue if you prefer)

thx!

swederik:
You're right, there seems to be a highlight issue.

1. Create two elliptical ROIs. Highlighting works fine
2. Create a length ROI. Highlighting works fine on Length ROI
3. Hover over either elliptical ROI - Nothing is highlighted
4. Click and move one elliptical ROI - highlights and moves as expected
5. Hovering over length and either ellipse works now.

Seems to be something strange with the highlight not affecting deactivated tools until they are interacted with.

kofifus:
I opened https://github.com/chafey/cornerstoneTools/issues/217 ... would be great to have a solution to this

PR #214, opened 2017-06-11
Fix import/export extentions

kofifus:
This PR allows direct module loading with <script type="module" ... >. This removes the need to go through the webpack build stage when developing.

Note that ATM type module requires chrome beta and in chrome://flags/ enabling 'Experimental Web Platform features' (see https://jakearchibald.com/2017/es-modules-in-browsers)

Changes are:

- Adds '.js' extensions where missing in import/export commands

- added two new files in cornerstoneTools\src: cornerstone-core.js and cornerstone-math.js
These files import cornerstone and cornerstoneMath respectively from the relevant directories. All files in cornerstoneTools use these two which gives a single place to change paths of the external packages if needed.

- in touchInput.js removed the line
`import Hammer from 'hammerjs';`
Hammer is in the global scope when hammer.js is included 

- in getMAxSimulataneousRequests fixed the circular dependency (see https://github.com/chafey/cornerstoneTools/pull/216)

- in src/index.js and stackSpecificStateManager.js fixed exports for stackSpecificStateManager

see attached allTools example which runs using ES6 modules without webpack ...
(files need to go in the allImageTools directory)

[allToolsES6modules.zip](https://github.com/chafey/cornerstoneTools/files/1070433/allToolsES6modules.zip)


kofifus:
included the fix from https://github.com/chafey/cornerstoneTools/pull/216
I think this is now complete

kofifus:
included the fix from #216
I think this is now complete

kofifus:
included the fix from #216
I think this is now complete

src/cornerstone-core.js
```diff
@@ -0,0 +1 @@
+export * from '../../cornerstone/src/index.js';
```
swederik: I don't think I can merge this. Won't it break users that are using module systems? (e.g. import * as cornerstone from 'cornerstone-core'?

kofifus: I don't think so .. I think it should still work as webpack still exports 'cornerstone-core' (which is different to 'src\cornerstone-core.js' which I added and is used internally. 

I added cornerstone-core.js so that if someone chooses to load modules directly with script type="module" but have cornerstone in a different path than cornerstoneTools, they only need to change the path in 'src\cornerstone-core.js' instead of in each file. This also made the changes I've done simpler and easier to understand/merge.
 
Since this is now available in both chrome beta and firefox beta it will soon be available in the official version and make developement much easier without the separate webpack stage.

kofifus: If you don't have time I'll check tomorrow

kofifus: OK you were wight, it does break the webpack build and in a very incidious way - webpack builds fine but when running for example allTools you end up with two separate instances of cornerstone, one from loading cornerstone.min.js and one from importing cornerstone-core.js 
I have made more commits to resolve this, pls see my comment below.

kofifus:
There is an interesting issue here - webpack allows creating an export such as:
```
   'cornerstone-core': {
      root: 'cornerstone'
   },
```
in webpack-base.js 

and then importing/exporting them with
`   import * as cornerstone from 'cornerstone-core'` or 
`    export * from 'cornerstone-core';`

The problem is that the above two lines are not valid javascript code. For two reasons - first they lack a path, you need at least a ./ path, and second they don't point to an existing file.
Because of that when load files containing such lines directly using <script type="module" ..> they break the page. ATM in chrome & firefox  beta they crash the tab (!).

What I've done is made all fils that need cornerstone or cornerstoneMath include cornerstone-core.js and cornerstone-math.js from the src directory. These files then do `export * from 'cornerstone-core';` and `export * from 'cornerstone-math';`.

This makes webpack build correctly. I have also verified by getting fresh copies, replacing the sources with those from this PR, doing npm install & npm run start, and tested a few examples (allImageTools and allTouchTools).

I have left remarks in cornerstone-core.js and cornerstone-math.js that when loading sources directly with <script type="module"> you need to replace `export * from 'cornerstone-core';` with `export * from '../../cornerstone/src/index.js';`. This is not great but at least it's a single point of change and the change is small. And allow this PR to be merged now.

Please let me know if this is not clear or if there are any further difficulties in merging this.

swederik:
Hey,

Sorry for taking so long on this. Could you rebase this on top of master?

kofifus:
Ok I pulled the latest head, and updated my commit. All ready to merge.

Note that is's not easy (and not that great practice) for me to squash my branch commit. However on your side you just need to click 'confirm squash and merge'

Let me know if anything else is needed.

swederik:
Thanks!

adreyfus:
I have a problem with the first line of src/inputSources/touchInput.js because I don't have global Hammer (I load it as a module).

Is it possible to do same same trick than for `cornerstone-core` and cornerstone-math ?

swederik:
That would be fine with me.

kofifus:
adreyfus are you going to put a PR or do you want me to do it ?

PR #216, opened 2017-06-14
Fix circular dependency

kofifus:
Fix the circular dependency in getMaxSimultaneousRequests.js (see https://github.com/chafey/cornerstoneTools/issues/215)

Added a setMaxSimultaneousRequests called from stackPrefetch.js and removed importing stackPrefetch from getMaxSimultaneousRequests.js

Also added .gitattributes to prevent windows/unix cr/lf mismatch


swederik:
Thanks!

PR #218, opened 2017-06-19
Fix for https://github.com/chafey/cornerstoneTools/issues/217 

kofifus:
This PR closes https://github.com/chafey/cornerstoneTools/issues/217

The highlight issue occurred whenever a new annotation was created, all other annotations would not be highlighted when moved over.

### DIFF
--- a/src/manipulators/moveNewHandle.js
+++ b/src/manipulators/moveNewHandle.js
@@ -29,8 +29,8 @@ export default function (mouseEventData, toolType, data, handle, doneMovingCallb
   }
 
   function whichMovement (e) {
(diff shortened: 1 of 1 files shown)


swederik:
Thanks!

kofifus:
debugging is so much easier now that I can load the sources directly ...

Can https://github.com/chafey/cornerstoneTools/pull/214 be merged as well ?

swederik:
I want to play around with #214 a bit before merging it, since there are a number of scenarios to test (e.g. interactions with dependencies, usage when imported into the OHIF Viewer).

PR #223, opened 2017-08-11
Feature/improve request pool manager

adreyfus:
- new request type `'autoPrefetch'`: allows to load series not expressly requested by the user, when there is nothing more urgent.

- `requestPoolManager.addPriorRequests()`:  adds a list of requests in one step, at the beginning of the array.

- remove initialization of maxNumRequests, because one might think that changing these values has an effect, whereas they are always overridden at the beginning of `startGrabbing()`


swederik:
Retrying seems like a good idea, but I would presume we would want a maximum amount of tries or something. 

What is the autoPrefetch request type? It wasn't clear to me what that was for.

adreyfus:
The 'autoPrefetch' type allows to load series not expressly requested by the user, when there is nothing more urgent.

adreyfus:
Thanks for the suggestion, I will propose a version of the retrying, with a maximum number of attempts.

swederik:
So addPriorRequests is meant for top-priority requests? Maybe instead we could change addRequest to include some option that either puts the request at the top or bottom of the stack? (or a priority value, like we have with the WADO Image Loader task manager) It just feels weird to have two ways to add things to the request pool.

adreyfus:
`addPriorRequests` adds a list of requests in one step, at the beginning of the array. 
If we use `addRequest` with an option to add the requests one by one, at the begininng of the table, performance will be poor.

adreyfus:
Not ready to merge

adreyfus:
Seems to be OK. 

It takes me a long time to maintain all my PRs with each version change. What can I do to make them accepted ?

swederik:
Hi Aloïs,

I am really sorry about this. It's hard to decide what should go in and what shouldn't at this point. Cornerstone Tools has already grown quite large and every new API / feature that goes in needs to be maintained indefinitely. I am therefore extremely hesitant nowadays to add new things that I don't see clear needs for and/or haven't properly found time to review. I'll try to get back to your PRs and see what we can do about them.

adreyfus:
Does not work with cornerstoneTools 2.0, because native Promise prototype has no function `state()` to know the state of the Promise.

adreyfus:
Ok for me (ready to merge).

I am open to all comments and suggestions.

dannyrb:
`requestPoolManager` is part of `cornerstoneTools` I'm not yet well acquainted with. Can you explain what it is, how we use it, and what benefits this new change adds? Please keep in mind that I don't have a long history in the medical field. What helps me _really_ understand are "real world" examples of when and why _feature X_ is important. An answer would go a long way toward me merging something without having to pester @swederik 😜 

Would we modify any tools to use this new feature? What's your motivation for adding it?

It looks like `requestPoolManager` is exposed on `cornerstoneTools` and used in:

- [stackTools/stackPrefetch.js](https://github.com/cornerstonejs/cornerstoneTools/blob/29182180ed3f850ba41c609b98b96464affca5f0/src/stackTools/stackPrefetch.js#L3)
- [util/scrollToIndex.js](https://github.com/cornerstonejs/cornerstoneTools/blob/29182180ed3f850ba41c609b98b96464affca5f0/src/util/scrollToIndex.js#L4)

@swederik has already called out the biggest question about [addPriorRequests](https://github.com/cornerstonejs/cornerstoneTools/pull/223#issuecomment-331091889). It feels like a convenience method that may not have a ton of application, and it muddies how to add requests. If there's a good justification for it (ie. It's something many users would benefit from/use), then I am more okay with it.

The only other question I have is, instead of adding _another_ `requestPoolType`, why not give the end users the ability to add their own types + priorities? We could add new methods for:

- `getRequestPoolTypes`
- `addRequestPoolType(string: name, int: priority, int?: maxRequests)`
- `removeRequestPoolType(string: name)`

This prevents us from adding future arbitrary `requestPoolType`s, and gives the user more flexibility.

adreyfus:
Hi Danny,

It's a good idea permit to give the end users the ability to add their own types + priorities. 
There is a bit of work, but I will change my PR to propose that.

And I will move the  addPriorRequests function in a specific PR. This function is very useful ... for me but maybe I'm the only one ...

PR #224, opened 2017-08-11
Feature/loadHandlerManager.displayLoadingHandler

adreyfus:
displayLoadingHandler: action executed when trying to display a loading frame


swederik:
So this displayLoadingHandler runs when, exactly? I see that you're adding an argument to requestPoolManager.addRequest, but I don't see any changes to that file which would actually make it do anything.

Our strategy until now has been to use start and end functions, so the developer can add

- On Start, display Loading icon
- On End, hide Loading icon,
- On Error, display Error icon (and hide any Loading icons)

How does displayLoadingHandler fit into this?

adreyfus:
`requestPoolManager.addRequest(..., pendingCallback)` is implemented in the PR #223.

Unlike `startLoadingHandler` that is executed whenever you change frames, the `pendingCallback` (=`displayLoadingHandler`) is only executed if the requested `imagePromise` is in the state `"pending"`.

adreyfus:
Not ready to merge

PR #225, opened 2017-08-11
Feature/improve mouse input

adreyfus:


src/inputSources/mouseInput.js
```diff
@@ -89,6 +89,12 @@ function mouseDown (e) {
     type: eventType
   };
 
+  const onMouseDown = $(element).data('onMouseDown');
```
swederik: This seems like a weird way to add a function to run on an element on mouse down. What are you using this for? Why shouldn't users just add an event handler to CornerstoneToolsMouseDown instead?

swederik:
Thanks for your contribution!

adreyfus:
You're absolutely right, I did not see that the event 'CornerstoneToolsMouseDown' was triggered on the viewportElement. Sorry :)

swederik:
Thanks!

PR #226, opened 2017-08-17
scrollLoop

adreyfus:


src/util/scroll.js
```diff
@@ -1,7 +1,7 @@
 import scrollToIndex from './scrollToIndex.js';
 import { getToolState } from '../stateManagement/toolState.js';
 
-export default function (element, images) {
+export default function (element, images, loop) {
```
swederik: Can you set a default value? (false, I guess.

src/stackTools/stackScroll.js
```diff
@@ -30,7 +30,7 @@ function mouseDownCallback (e, eventData) {
 function mouseWheelCallback (e, eventData) {
   const images = -eventData.direction;
 
-  scroll(eventData.element, images);
+  scroll(eventData.element, images, stackScroll.getConfiguration().scrollLoop);
```
swederik: Can you set a default value for this configuration setting? And change it just to 'loop'. I guess that's clear enough, since it's already on the scroll tool.

src/util/scroll.js
adreyfus: done :)

src/stackTools/stackScroll.js
adreyfus: done :)

PR #227, opened 2017-08-30
Fixing tool handles on non-chrome browsers

brunoalvesdefaria:
Tool handles highlighting on mouse over was not working on Firefox, IE and Safari.
The problem was with the use of ["MouseEvent.which"](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/which) attribute, which is a non-standard.
The code was changes to use ["MouseEvent.buttons"](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons) instead, which is the MDN's recommendation.


brunoalvesdefaria:
This would change the way we work with events and impact dependent applications.
The switching to use event.buttons instead of event.which should be done in a major version.

PR #231, opened 2017-09-01
Feature/no global variables

adreyfus:


swederik:
@kofifus One of you guys is removing import $ from 'jquery' and the other wants to add it. Can we settle this?

adreyfus:
That's why I used an intermediate file jquery.js. Can you use it to load jQuery globally ? 

The problem seems to be the same for Hammer and jQuery : without import, they don't work as AMD modules.

kofifus:
if we have `import * as Hammer from 'hammerjs';` in the code, then loading cornerstone directly in the browser with `<script type="module">` will break unless you also load hammer (and etc jquery) with `<script type="module">` which you probably don't want to do.

One solution as said above is to use the same 'trick' as with cornerstone-core. However this is pretty ugly. I expect soon loading with `<script type="module">` will make it to main chrome build and to have to have different version for amd/webpack and for direct loading will become a real hustle. 

But for now that's the only way I can see. I can get to this on Monday unless someone have a better idea, thx.

kofifus:
I think this PR will work well :)

The only thing I would ask is to add remarks in the top of jquery.js and hammer.js:

/*
 * When loading sources directly with <script type="module"> remove the line below
 * keep only the export line 
 */

adreyfus:
Done ^^

swederik:
Thanks!

PR #235, opened 2017-09-12
Feature/shallow copies of images in layers

adreyfus:
Use shallow copies of images in layers, to prevent the function `convertImageToFalseColorImage()` from altering original images.


swederik:
Hmm this is indeed a problem we had overlooked. Thanks for the PR.

src/stackTools/fusionRenderer.js
```diff
@@ -31,9 +31,9 @@ export default class FusionRenderer {
         const currentLayerId = this.layerIds[0];
         const layer = cornerstone.getLayer(element, currentLayerId);
 
-        layer.image = image;
+        layer.image = $.extend({}, image);
```
swederik: Could we do this without jQuery instead? Given that we still eventually plan to remove it.

adreyfus: The ES6 equivalent is Object.assign(), but not IE compatible (and honestly I don't think it's a good idea to remove jQuery).

adreyfus: I didn't see there is already many Object.assign() in the code.

adreyfus: done :)

PR #237, opened 2017-09-15
sensitivityFactor

kofifus:
The point of this PR is accessibility. It add a 'sensitivityFactor' global configuration to touchTool and mouseButtonTool (default to 1.0) which is multiplied to the distance before being used in the pointNearTools of all tools.

This allows a single point of change to make the tools less/more sensitive to touch/mouse:

// make all tools a bit more sensitive to mouse movement/clicks (need to be nearer to select etc):
cornerstoneTools.mouseButtonTool().setConfiguration({ sensitivityFactor: 0.7 });

// make all tools a bit more sensitive to touch (need to be nearer to select etc):
cornerstoneTools.touchTool().setConfiguration({ sensitivityFactor: 0.7 });

// make all tools a bit less sensitive to mouse movement/clicks (can be further to select etc):
cornerstoneTools.mouseButtonTool().setConfiguration({ sensitivityFactor: 1.3 });

// make all tools a bit less sensitive to touch (can be further to select etc):
cornerstoneTools.touchTool().setConfiguration({ sensitivityFactor: 1.3 });

In addition:

-  mouseButtonTool.getConfiguration and touchTool.getConfiguration now return both local and global configuration object 
- mouseButtonTool.setConfiguration and touchTool.setConfiguration now override both local and global configurations with those given (instead of erasing any existing configuration).

### DIFF
--- a/src/imageTools/angleTool.js
+++ b/src/imageTools/angleTool.js
@@ -50,7 +50,7 @@ function createNewMeasurement (mouseEventData) {
 }
 // /////// END ACTIVE TOOL ///////
 
-function pointNearTool (element, data, coords) {
+function pointNearTool(element, data, coords, distance) {
   const lineSegment = {
     start: cornerstone.pixelToCanvas(element, data.handles.start),
     end: cornerstone.pixelToCanvas(element, data.handles.end)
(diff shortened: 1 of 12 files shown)


swederik:
Thanks for the PR. This is indeed something that can be annoying in some cases. Rather than using a sensitivity factor, though, I think we should just allow developers to set some configuration values for the pointNearTool distance directly, rather than a multiple of the default values.

dannyrb:
Something similar to this is in the works in the `vNext` branch. I'll close this for now. Feel free to re-open if `v3` doesn't address your concerns.

PR #241, opened 2017-09-21
Feature/position difference undefined instead of crash

adreyfus:
If `initialData.distances[sourceImageId]` is undefined, keep `positionDifference` undefined instead of crashing.


swederik:
I have never encountered this issue so I would like to understand it before merging.

So you have two synchronized stacks but when you added the sources / targets, there were some images that weren't present? And now when you try to scroll it checks for the initial distance so it can maintain that distance, but that distance doesn't exist. And so we are still firing the event but now with an undefined distance.... and what should happen? Nothing?

adreyfus:
`viewport0`: `stack0.imageIds= ["dicomfile:0"]`, with a correct `imagePositionPatient`

`viewport1`: `stack1.imageIds= ["dicomfile:1"]`, with no `imagePositionPatient`

`fireEvent(viewport1, eventData)`:

`sourceElements`: `[viewport0, viewport1]`

`targetElements`: `[viewport0, viewport1]`

`initialData.imageIds.sourceElements`: `["dicomfile:0"]`

`var sourceImageId = initialData.imageIds.sourceElements[1]` => `undefined`

`initialData.imageIds.targetElements`: `["dicomfile:0", "dicomfile:1"]`

`initialData.distances`: `{}`

`positionDifference = initialData.distances[sourceImageId][targetImageId]` throws the error `"TypeError: Cannot read property 'dicomfile:0' of undefined`

PR #243, opened 2017-09-22
Feature/event CLIP_SCROLLED

adreyfus:


swederik:
What do you plan to do with this event? Does CornerstoneNewImage not suffice?

adreyfus:
Hi,

`'CornerstoneClipScroll'` is the equivalent event to `'CornerstoneStackScroll'` for clips. I have a synchronizer on them. 

The `'CornerstoneNewImage'` event is triggered at each calling of `cornerstone.displayImage`, and does not contain the `direction` information.

adreyfus:
Not ready to merge (last integration of changes of the branch master was hard)

dannyrb:
> The 'CornerstoneNewImage' event is triggered at each calling of cornerstone.displayImage, and does not contain the direction information.

Would you be able to calculate direction by:

- Tracking previous imageId displayed
- Getting the new imageId emitted on `cornerstone.displayImage`
- Find both imageId's index in stack
- Compare indexes to determine direction?

I'm not against adding this, and if you added tests, it would probably be fine to. I just want to make sure we're not emitting data multiple times when an image changes without good reason.

adreyfus:
Hi,

The solution you propose seems more complex. This PR is deliberately small with few changes, we simply add useful information.

adreyfus:
I don't know how to test features that require a complete cornerstone environment.

dannyrb:
It has been several months since this has been touched. In the interest of keeping things clean, I am going to close this for now. I'm not sure this adds enough value to increase our external API. I'm also not convinced that this is the best way to expose this information.

It may be a better fit to determine and add direction to a cornerstone core event? 

I appreciate your contributions! If you feel this is worth discussing, I am more than willing to work with you to try and identify a way to merge similar functionality.

PR #245, opened 2017-09-27
Feature/fusionRenderer does not crash if the layer is destroyed while rendering

adreyfus:


swederik:
"Does not crash" sounds more like a bug fix than a feature... :-) Thanks!

PR #246, opened 2017-09-27
Feature/stackImagePositionSynchronizer does not crash if no position found

adreyfus:


adreyfus:
Not ready to merge (last integration of changes of the branch master was hard)

src/synchronization/stackImagePositionSynchronizer.js
```diff
@@ -15,17 +15,40 @@ export default function (synchronizer, sourceElement, targetElement) {
...
+
+    return;
+  }
 
   const stackToolDataSource = getToolState(targetElement, 'stack');
   const stackData = stackToolDataSource.data[0];
 
   let minDistance = Number.MAX_VALUE;
   let newImageIdIndex = -1;
+  let nbComparedPositions = 0;
```
swederik: I guess this is a remnant of testing? If you remove nbComparedPositions I think we can merge this PR

src/synchronization/stackImagePositionSynchronizer.js
```diff
@@ -15,7 +15,17 @@ export default function (synchronizer, sourceElement, targetElement) {
...
   const sourceImage = cornerstone.getEnabledElement(sourceElement).image;
   const sourceImagePlane = cornerstone.metaData.get('imagePlaneModule', sourceImage.imageId);
-  const sourceImagePosition = convertToVector3(sourceImagePlane.imagePositionPatient);
+  let sourceImagePosition;
+
+  if (sourceImagePlane !== undefined) {
+    sourceImagePosition = convertToVector3(sourceImagePlane.imagePositionPatient);
+  }
+
+  if (sourceImagePosition === undefined) {
```
swederik: I just realized there is no way this condition can happen. convertToVector3 doesn't have any way to return undefined. I guess it's better to move this If statement above the other one and check if sourceImagePlane.imagePositionPatient is undefined.

See https://github.com/chafey/cornerstoneTools/blob/55226c082b2e5ab2de57b6e2bdb82c072eafd516/src/util/convertToVector3.js#L9

adreyfus: Done !

swederik:
That looks better! Thanks!

Issue #253, opened 2017-10-04
Calling brush.deactivate() causes brush.onImageRendered to be used and draw green hover circle


millerjv:
A naive call to brush.deactivate() without ever calling brush.activate() causes brush.onImageRendered() to be called and draws the green hover circle that tracks mouse movements.

This occurs when I call a naive method to disable all the tools before enabling a specific tool.

swederik:
Should be fine now. Thanks for the report!

PR #258, opened 2017-10-12
Deactivate method for Brush tool

andrebot:
# Issue
Described at #253 .
# Solution
Augmented the `mouseButtonTool` deactivation flow to expect a `deactivate` method from `mouseToolInterface`. If there is such method it will execute it.
I added this so we could seamlessly improve the deactivate flow without affecting the other tools. This give us the flexibility to customize the deactivation of each tool. For instance, brush tool uses the onImageRenderer method to draw the green tracking ball in the viewport. All the other tools uses the same method to draw the tool state into the canvas. This is not true to the brush tool which update its state at `getPixelData`


swederik:
This will be merged in with https://github.com/chafey/cornerstoneTools/pull/279

swederik:
Merged in earlier. Thanks!

PR #266, opened 2017-10-16
adding a 3rd parameter to saveAs to allow other mimeTypes, such as im…

leonardorame:
…age/jpeg.

### DIFF
--- a/src/imageTools/saveAs.js
+++ b/src/imageTools/saveAs.js
@@ -1,6 +1,8 @@
 import { $ } from '../externalModules.js';
 
-export default function (element, filename) {
+export default function (element, filename, mimetype) {
+  // setting the default value for mimetype to image/png
+  mimetype = mimetype || "image/png";
   const canvas = $(element).find('canvas').get(0);
 
     // Thanks to Ken Fyrstenber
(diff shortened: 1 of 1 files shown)


leonardorame:
Hi Erik, could you review this again?.

src/imageTools/saveAs.js
```diff
@@ -1,6 +1,8 @@
 import { $ } from '../externalModules.js';
 
-export default function (element, filename) {
+export default function (element, filename, mimetype) {
```
swederik: Can you make "image/png" the default mimetype?

swederik:
Thanks!

src/imageTools/saveAs.js
leonardorame: I made it the *old way*:

    // setting the default value for mimetype to image/png
    mimetype = mimetype || "image/png";

Of course I can change to `param = defaultvalue`, do you prefer that way?.

swederik: It's a minor difference, I guess there's no need to change it. I realized it didn't matter and just merged it :-) Thanks!

PR #271, opened 2017-10-18
Injection proposal

andrebot:
# Issue
We were seeing a lot of people complaining about weird behaviors with cornerstone and cornerstoneTools due to cornerstoneTools using the wrong cornerstone instance, webpack doing a bundling and ending up not getting the right reference throughout the app and etc. This was being caused by different versions of cornerstone in the project. If, for instance, cornerstoneTools is requiring cornerstone version `0.11.0` and in your project you are requiring cornerstone version `0.13.0` NPM would nest those dependencies inside the `node_modules` thus causing the errors. On top of that, we are always expecting to have cornerstone as a global variable, which can cause more unexpected behavior.
# Solution
My proposal is to not expect cornerstone as a global variable anymore and force cornerstoneTools to require a cornerstone initialization so it can use whatever version you are using in your project. This would avoid module nesting by NPM, it would give the user a better control of cornerstone version usage (without being stuck with whichever version cornerstoneTools is locked into) and it is a safer than using the deprecated peerDependencies from NPM.

To avoid restructuring the whole code, I suggest to use the `externalModules.js` file we are already using to export the right cornerstone throughout this library and to set the right cornerstone to use. And we removed the cornerstone dependency.


swederik:
Looks good so far!

To those curious, this should help address: 
https://github.com/chafey/cornerstoneWADOImageLoader/issues/110
https://github.com/chafey/cornerstoneTools/issues/257

kofifus:
yes this file is essential for loading the sources directly in chrome using script type="module" - don't remove it !

I want to write instructions on doing that but am waiting for a resoulution of issue https://github.com/chafey/cornerstoneTools/issues/257

kofifus:
Also I think maybe we need to think further this PR - perhhaps a better/simpler solution can be made ?

externalModules.js already provide a single point of importing. Perhaps we can instead introduce a versioning mechanism - that is in cornerstoneTools\src\externalModules.js check that the cornerstone version imported is the matching  one for that version of cornerstoneTools and otherwise throw an exception ? this will force users to have matching versions of cornerstone and cornerstoneTools while introducing only a small change to the existing codebase.

kofifus:
I still think versioning is the better solution.

- I don't see the problem if there is no backward support. This is a new feature and we simply say - from this version on, cornerstoneTools requires a matching version of cornerstone. If people choose to work with old versions that's their choice but pretty much everyone will soon upgrade.

- we will not need the user to explicitally do any work such as initializing conernerstone or providing versions we can simply have:

in cornerstone.js:
`  export const version =0.14`

in cornerstoneTools\src\externalModules.js:
```
  const minimalCornerstoneVersion=0.14;
  if (cornerstone.version && cornerstone.version < minimalCornerstoneVersion) throw ...

```
or something of the sort.

src/externalModulesES6modules.js
```diff
@@ -4,4 +4,15 @@ import * as cornerstoneMath from '../../cornerstoneMath/src/index.js';
 const $ = window.$;
 const Hammer = window.Hammer;
 
-export { $, Hammer, cornerstone, cornerstoneMath };
+let cornerstone;
```
swederik: are you sure we can't do:
````
let cornerstone = window.cornerstone; 
````

here? Then we wouldn't have to change examples where things are added via the script tag? Not sure if this would cause confusion, though. I just don't want people to have to change code that already works.

andrebot: we actually can. I will change to this setup

kofifus:
Thanks andrebot 

I personally disagree. Restraints over versions are to me a very good thing. They protect the users from running into very hard to find bugs and protects the developers from having to deal with issues reports stemming from someone using some combination of versions where who knows what's going on.

As in my example above we don't need to mandate exact version matches but can require a minimum version of other external modules. This to me gives enough balance between flexibility and protection. If a user can't upgrade then he can't upgrade ... to allow him/her to use a combination of versions that the developers don't know wether or not works is to me wrong.  So I think the opposite - we will just maximize the version management madness in the long run not minimize it.

It is true that there will be an overhead in the sense where the developers will have to decide/update the minimal versions needed when releasing a new version, however to me that's a good overhead. If (and only if) for example a change is made to cornerstone which breakes the current cornerstoneTools version and so requires a change to cornerstoneTools so that it works with the new cornerstone version, then while you make that cornerstoneTools change you also increase the minimal cornerstone version needed. I suspect that this will not happen often.

PR #272, opened 2017-10-21
feat(prefetch): add a maximum number of images to prefetch

Maistho:
Adds a new configuration option to the stack prefetch tool, allowing it
to only prefetch the closest X images, where X is 10 as default. It
significantly improves the stability of prefetching when limiting the
cache size.

### DIFF
--- a/src/stackTools/stackPrefetch.js
+++ b/src/stackTools/stackPrefetch.js
@@ -7,10 +7,12 @@ import { setMaxSimultaneousRequests } from '../util/getMaxSimultaneousRequests.j
 const toolType = 'stackPrefetch';
 const requestType = 'prefetch';
 
-let configuration = {};
+let configuration = {
+  maxImagesToPrefetch: Infinity
+};
 
 let resetPrefetchTimeout;
-const resetPrefetchDelay = 300;
+const resetPrefetchDelay = 10;
 
 function range (lowEnd, highEnd) {
     // Javascript version of Python's range function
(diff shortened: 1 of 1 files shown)


swederik:
Hi Gustav,

Thanks for the PR!

I think by default stackPrefetch should fetch the whole stack, though I can see where this might be useful (e.g. very large stacks and small caches). What is your use case where you have such a limited cache size? Mobile?

Maistho:
We have an application where all series are around 1000-3000 images, and we
load two side by side for comparison. We've had problems with crashes (out
of memory) when not limiting the cache, and other issues when we limit the
cache and use the prefetch.

It seems like a waste to get more images than can fit in the cache, since
it will just perpetually load images into the cache.

I can see why the default would be to fetch all the images though, so there
should probably be a option for that as well.

Would it be alright if I changed the default to Infinity?

swederik:
Oh that's very interesting.... 

Another option that might help you is to hook up an event handler to listen for the event when the cache is full, and then use the handler to stop prefetching. This functionality was removed here (https://github.com/chafey/cornerstoneTools/commit/47dbd667abb2d6756b2ceaf2b8d5dbb4a82bcb26#diff-b5e5d95181782e7ebf78d4fd7591a55a) since it wasn't what we wanted, but it might be good for your use case.

Changing to Infinity is also an option.

Maistho:
Changed the default to Infinity, which should mean that the default behaviour will be to prefetch all images.

PR #274, opened 2017-11-01
trigger custom events in addition to jQuery events

Maistho:
This PR triggers CustomEvents in the same way as cornerstone-core. Currently triggers both CustomEvent and jQuery event.


swederik:
Thanks! I'm trying to sort out some overlay issues but I will get back to this stuff after.

src/inputSources/mouseInput.js
```diff
@@ -108,15 +110,10 @@ function mouseDown (e) {
     type: eventType
   };
 
-  const event = external.$.Event(eventType, eventData);
-
-  external.$(eventData.element).trigger(event, eventData);
+  triggerEvent(eventData.element, eventType, eventData);
 
-  if (event.isImmediatePropagationStopped() === false) {
```
swederik: Any specific reason this was removed? The rest looks fine to me.

Maistho: I made a mistake now I see! I thought that events triggered asynchronously, but it seems they trigger synchronously. DispatchEvent returns a boolean that indicates if any listeners called `preventDefault`.

I guess you learn something everyday!

I'll make a change to reinstate the old behaviour.

Maistho: I have fixed it and some other minor issues, the old behaviour should be back now.

src/synchronization/stackImagePositionSynchronizer.js
```diff
@@ -22,7 +22,7 @@ export default function (synchronizer, sourceElement, targetElement) {
   let minDistance = Number.MAX_VALUE;
   let newImageIdIndex = -1;
 
-  external.$.each(stackData.imageIds, function (index, imageId) {
+  Array.prototype.forEach.call(stackData.imageIds, function (index, imageId) {
```
swederik: why not just stackData.imageIds.forEach( ... ?

swederik: Oh, and index / imageId are switched when going from jQuery to forEach. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

src/inputSources/mouseInput.js
```diff
@@ -108,11 +108,9 @@ function mouseDown (e) {
     type: eventType
   };
 
-  const event = external.$.Event(eventType, eventData);
+  const eventPropogated = triggerEvent(eventData.element, eventType, eventData);
```
swederik: minor comment, but propagated is spelled wrong

src/inputSources/mouseInput.js
```diff
@@ -108,11 +108,9 @@ function mouseDown (e) {
...
 
-  external.$(eventData.element).trigger(event, eventData);
-
-  if (event.isImmediatePropagationStopped() === false) {
+  if (eventPropogated) {
```
swederik: Something is wrong with the changes to mouseInput but I am not sure what. I guess it has to do with the stopImmediatePropogation or preventDefault changes.

If you create a length measurement, and then try to grab and move it, it will fail and instead create a new one. This is because mouseDownActivate is firing instead of mouseDown. Any ideas?

src/synchronization/stackImagePositionSynchronizer.js
Maistho: I wasn't sure that imageIds was a real array, but if it's always a real array then it should be fine to use `.forEach` directly.

Fixing!

src/inputSources/mouseInput.js
Maistho: I think it might be because some of the listeners use jQuery, and I only return the dispatchEvent value. I'll try changing the return value of `triggerEvent` to combine the jQuery and dispatchEvent.

Maistho: Fixed, I think.

I changed it so that it will not keep triggering regular events if any jQuery listeners call `stopImmediatePropagation()`, and so it returns the correct value (false if stopped/preventDefault)

src/inputSources/mouseInput.js
Maistho: Fixed

src/synchronization/stackImagePositionSynchronizer.js
Maistho: Looked it up, and in both cases they are real arrays.

Maistho:
I found another issue in the implementation of triggerEvent, the events were not set as `cancelable`, which has now been fixed. It seems to fix the issue with creating new tools instead of moving existing ones around.

swederik:
Closing because I rebased and had to cherry-pick these commits here: https://github.com/chafey/cornerstoneTools/pull/288/

Thanks!

PR #275, opened 2017-11-01
fix stackscrolling with touch on large image sets

Maistho:
This PR fixes touch scrolling for large image sets.

The previous behaviour led to scrolling more than 1 image per pixel if there were more images than the height of the element in pixels, which wasn't a good user experience.

New behaviour is to scroll at most 1 image per 2 pixels, and at least 1 image per 1/8 of the element height.

### DIFF
--- a/src/stackTools/stackScroll.js
+++ b/src/stackTools/stackScroll.js
@@ -55,8 +55,8 @@ function dragCallback (e, eventData) {
 
   const config = stackScroll.getConfiguration();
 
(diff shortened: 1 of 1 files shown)


swederik:
Hmm.. I always liked that if you clicked and drag from the top of the image and moved your mouse to the bottom, it would scroll the whole stack. I guess this only affects a minority of cases though, so I suppose we can just merge it.

Thanks!

PR #276, opened 2017-11-03
Update README.md

nicomlas:
It seems the external module key word as NO 's'


swederik:
Wow you are right. I guess I wasn't paying attention when changing the README. Thanks!

nicomlas:
You're welcome. Thanks to your work :) Great job.
FYI: I believe it is the same on other libs (WebImageLoader etc...)

swederik:
Ugh yeah thanks for the heads up!

PR #299, opened 2017-12-05
bug/startAgain() may give up while there are requests to be processed

adreyfus:
Adding a request wakes up the requestPoolManager. Without this change, `startAgain()` may give up while there are requests to be processed.


swederik:
Hi Aloïs,

Thanks for the bug fix. Before merging it, though, I want to understand what the issue is.

So the problem seems to be like this?

- You start requesting things
- Everything finishes, awake is then set to false
- You add some more requests (manually with addRequests, instead of the stackPrefetcher?)
- startGrabbing is never called, and awake is still false, so the requests are never sent?

Maybe instead we should have this at the end of addRequest?

````javascript
if (awake === false) {
  awake = true; // Or maybe startGrabbing should set awake = true itself
  startGrabbing();
}
````

adreyfus:
Hi,

The bug I encountered is:
- I start requesting things
- Everything starts, `awake` is then set to `false`
- I add some more requests (with `addRequests`)
- when the last running requests end, callback executes `startAgain()` but the function does nothing because `awake` is `false`

swederik:
Thanks! (and sorry for the delay)

Issue #295, opened 2017-11-22
Travis CI build passed despite errors during compilation


adreyfus:
Is it normal ?

dannyrb:
@adreyfus, can you post a copy of your output? I can get it to build with warnings, but not errors.

dannyrb:
@adreyfus is this still an issue? What are the next steps? Anything I can do to help?

adreyfus:
It seems to be resolved (Travis CI build fails as expected, if there is an error on compilation).

PR #301, opened 2017-12-06
Feature/test build errors validation

adreyfus:


swederik:
Hi Aloïs,

I'm sorry but I don't understand this PR. All it does is remove 'let'. What is the reason for that?

adreyfus:
Not for merging. I try to reproduce the behavior denounced in the issue #295 (but the tests do not start, I don't know why).

swederik:
Maybe try cleaning your node modules and doing npm install again? Not sure why they wouldn't run. Is it Windows maybe?

adreyfus:
The problem is not in my computer, but in the github repository.

swederik:
Ah maybe it has something to do with our move from @chafey to @conerstonejs ...

I'll check that. Thanks for the heads up.

PR #308, opened 2017-12-13
OHIF-187: Fix incorrect scroll direction, OHIF-188: Stack Scroll Tool is not working while dragging

ayselafsar:


freund17:
@swederik Can this get merged?

swederik:
Sorry for the delay. I also think we need to address the point you raised above as well.

PR #311, opened 2017-12-18
Add configuration and functional option to scroll a stack without skipping images

jdnarvaez:
I wanted a way to consume scroll events and prevent the stack state from progressing to the next index until the currently request index had been drawn. I *think* this covers that.

### DIFF
--- a/src/stackTools/stackScroll.js
+++ b/src/stackTools/stackScroll.js
@@ -47,12 +47,9 @@ function mouseWheelCallback (e) {
   let loop = false;
   let allowSkipping = true;
 
-  if (config && config.loop) {
-    loop = config.loop;
-  }
-
-  if (config && config.allowSkipping !== undefined) {
-    allowSkipping = config.allowSkipping;
+  if (config) {
+    loop = config.loop !== undefined ? config.loop : false;
+    allowSkipping = config.allowSkipping !== undefined ? || true;
   }
 
   scroll(eventData.element, images, loop, allowSkipping);
--- a/src/stackTools/stackScroll.js
+++ b/src/stackTools/stackScroll.js
@@ -45,12 +45,17 @@ function mouseWheelCallback (e) {
   const config = stackScroll.getConfiguration();
 
   let loop = false;
+  let allowSkipping = true;
 
   if (config && config.loop) {
     loop = config.loop;
   }
 
-  scroll(eventData.element, images, loop);
+  if (config && config.allowSkipping !== undefined) {
+    allowSkipping = config.allowSkipping;
+  }
+
+  scroll(eventData.element, images, loop, allowSkipping);
 }
 
 function dragCallback (e) {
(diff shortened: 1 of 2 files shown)


swederik:
Thanks for the contribution! I'll try to get to this ASAP, but it will likely be after the holidays.

src/stackTools/stackScroll.js
```diff
@@ -45,12 +45,17 @@ function mouseWheelCallback (e) {
...
 
   let loop = false;
+  let allowSkipping = true;
 
   if (config && config.loop) {
     loop = config.loop;
   }
 
-  scroll(eventData.element, images, loop);
+  if (config && config.allowSkipping !== undefined) {
```
dannyrb: You should be able to merge this check and default with the above check. For example:

_Previous:_
```
if (config && config.loop) {
  loop = config.loop;
}
```

_Becomes:_
```
if (config) {
  loop = config.loop !== undefined ? config.loop : false;
  allowSkipping = config.allowSkipping !== undefined ? || true;
}

File: src/util/scroll.js [hunk 2]
```diff
@@ -1,7 +1,37 @@
...
+      stackData.pending.splice(stackData.pending.indexOf(pendingEvent), 1);
+
+      if (stackData.pending.length > 0) {
+        scrollWithoutSkipping(stackData, stackData.pending[0], element);
+      }
+
+      return;
+    }
+
+    element.addEventListener('cornerstonenewimage', function(event) {
```
dannyrb: Maybe I'm missing it. Is this the right event?

jdnarvaez: I believe it is, as this seems to be the event that is triggered when the index *actually* changes.

src/stackTools/stackScroll.js
jdnarvaez: Can do.

dannyrb:
Trying to make sure I understand. In your new implementation, you bypass the `scrollToIndex` and instead:

1. Record the "pending" image change
2. Call `scrollWithoutSkipping`

Where we:

1. Verify the "pending" final scroll image is not already being displayed
2. Register a `cornerstonenewimage` event listener that
    - When the next image rendered matches the next "pending image"
        - We pop of the pending image event (it's no longer pending :P)
        - Remove the listener
        - Call `scrollWithoutSkipping` if we have more pending items
3. ^ Now that something is listening, we call `scrollToIndex`

I _think_ this would force `stackScroll` requests to process in order, So if the stack scroll calls are:
- display image 1
- display image 2
- display image 3
- etc.

Then they should all display in order without skipping. 

**Note:** I've also added some comments to your code with additional questions.

Follow-up question for @swederik, does `scrollToIndex` uhh.. Scroll? It looks like it queues up and loads whatever newImageId you give it?

jdnarvaez:
Correct. As far as I can tell scrollToIndex just queues the image call, which may or may not execute in order depending on how long the loadImage request takes.

swederik:
Looks good to me! Indeed scrollToIndex just queues it up.

adreyfus:
```
ERROR in ./src/stackTools/stackScroll.js
/home/travis/build/cornerstonejs/cornerstoneTools/src/stackTools/stackScroll.js
  52:58  error  Parsing error: Unexpected token ||
```

swederik:
Thanks @adreyfus I don't know why it wasn't clear that the build was failing before I merged that...

jdnarvaez:
Sorry about that, I didn't see that either when building.

PR #324, opened 2018-01-12
Feature/scrollToIndex change currentImageIdIndex before calling startLoadingHandler

adreyfus:


dannyrb:
Is this related to a specific issue in the backlog? Can you provide a real world example of why this change should be merged?

adreyfus:
Hi Danny,

There is no specific issue in the backlog. Should I create one for each new PR ?

The `startLoadingHandler` function permit to display a "Loading" message. 
But if we don't change the currentImageIdIndex before calling it, the frame number will correspond to the previous frame, instead of the currently loading frame.

dannyrb:
It may be a personal preference, but I prefer seeing an issue (or more info in the original PR). A lot of times, an issue gives you room to state:

_For Bugs_

- I expect behavior to be X, but it is Y
- Here is a solution that aligns the behavior with my expectations

_For Features_

- I want to be able to do X, or I want it to be easier to accomplish X, etc.
- Here is how I would approach modifying the code base to make X easier/possible

Both of these routes create a discussion around a change before any code is written. Sometimes there's a better idea for an implementation, or the functionality/behavior is outside the scope of the code base. I guess what I'm getting at is: 

> Discussion is cheap. Writing and revising code is expensive. 

(and nobody likes their PR being rejected). They also provide valuable context that make _reviewing_ an eventual PR easier :P If I have a better understanding of what the PR is trying to accomplish, it's easier for me to provide feedback and merge requests.

dannyrb:
> But if we don't change the currentImageIdIndex before calling it, the frame number will correspond to the previous frame, instead of the currently loading frame.

Does something update the frame number after it has loaded? And does it make more sense for the `currentImageIdIndex` to the be displayed image's index, or the loading image's index?

Does this have any repercussions where we purposefully wait for the previous image to be loaded and displayed before allowing the next image to load/display? (_I may be thinking of a different PR that's floating around_)

jdnarvaez:
I believe you may be thinking of my PR, @dannyrb

dannyrb:

Issue #326, opened 2018-01-15
Give the end users the ability to add their own types + priorities


adreyfus:
I will propose a PR to give the end users the ability to add their own requestPoolTypes + priorities. It will add new methods for:

getRequestPoolTypes
addRequestPoolType(string: name, int: priority, int?: maxRequests)
removeRequestPoolType(string: name)

cf. https://github.com/cornerstonejs/cornerstoneTools/pull/223#issuecomment-357245745

I remain open to your comments and suggestions.

adreyfus:
The current operation of the requestPoolManager is not clear for me, because it seems to have two priorities.

- In the function `getNextRequest()`: we first check the interaction requests, then the thumbnail requests, and then the prefetch requests.

- In the function `sendRequest()`:  a priority of -5 is assigned to `prefetch`, 0 to `interactive`, and 5 to `thumbnail` (about that, should not it be 'interaction' instead of 'interactive' ?). 

Can someone explain to me how these priorities work, please?

adreyfus:
@swederik @dannyrb

dannyrb:
To be honest, I'm not 100% sure myself.

When I was reviewing your previous PR, I noticed that too and thought it was odd that we specified a priority, but then appeared to be queuing up requests in a static order regardless of priority. It almost looks like we have a "first priority" (queue X tasks), and a "second priority" for already queued tasks that an imageLoader can choose to use if it so wishes.

I'm _not sure_ if I'm right, or if this is be design. @swederik may have more insight. I'll take a harder look later this week.

It looks like `priority` is passed to the image loader to use (if it so chooses). You can see it in use here in the `cornerstoneWADOImageLoader`

- [src/imageLoader/webWorkerManager.js](https://github.com/cornerstonejs/cornerstoneWADOImageLoader/blob/eab9926e6120afeb0a7092a77aced7187cd05aba/src/imageLoader/webWorkerManager.js#L198)
- [src/imageLoader/decodeImageFrame.js](https://github.com/cornerstonejs/cornerstoneWADOImageLoader/blob/1498ba391d690794d37da83e135b93ac4350fd33/src/imageLoader/decodeImageFrame.js#L5)

Note: It does look like `interactive` should be `interaction`. I don't think anyone has noticed, as the default priority is 0. So not receiving a match should have the same result.

adreyfus:
Do you think we should define a requestsPriority and a taskPriority ?

adreyfus:
[...]

PR #327, opened 2018-01-16
Feature/define own request types

adreyfus:
This PR is related to the issue https://github.com/cornerstonejs/cornerstoneTools/issues/326.


adreyfus:
This PR is currently NOT ready to merge.

adreyfus:
This version give the end users the ability to add their own types + request priorities, without breaking the backward compatibility.

src/requestPool/requestPoolManager.js
```diff
@@ -60,7 +112,7 @@ function addRequest (element, imageId, type, preventCache, doneCallback, failCal
 function clearRequestStack (type) {
   // Console.log('clearRequestStack');
```
dannyrb: Can you remove this comment while you're updating this file?

src/requestPool/requestPoolManager.js
```diff
@@ -60,7 +112,7 @@ function addRequest (element, imageId, type, preventCache, doneCallback, failCal
...
   if (!requestPool.hasOwnProperty(type)) {
-    throw new Error('Request type must be one of interaction, thumbnail, or prefetch');
+    throw new Error(`Request type ${type} is not defined`);
```
dannyrb: I love string interpolation <3

src/requestPool/requestPoolManager.js
```diff
@@ -175,21 +221,26 @@ function startGrabbing () {
...
-  if (requestPool.thumbnail.length && numRequests.thumbnail < maxNumRequests.thumbnail) {
-    return requestPool.thumbnail.shift();
+    if (requestPool[name].length && numRequests[name] < getMaxRequests(i)) {
+      return requestPool[name].shift();
+    }
   }
 
-  if (requestPool.prefetch.length && numRequests.prefetch < maxNumRequests.prefetch) {
-    return requestPool.prefetch.shift();
+  for (let i = 0; i < requestPoolTypes.length; i++) {
```
dannyrb: I wonder. Do we _need_ to iterate over each request type again? In the first loop, we already check for `requestPool[name].length`. We could change...

```
if (requestPool[name].length && numRequests[name] < getMaxRequests(i)) {
    return requestPool.thumbnail.shift();
}
```

To something like...

```
let hasRequestsInQueue = false;

const hasPooledRequests = requestPool[name].length > 0;
const isUnderMaxActiveRequests = ...
if(hasPooledRequests && isUnderMaxActiveRequests){
   ...
}else if(hasPooledRequests) {
    hasRequestsInQueue = true;
}
```

Or at least something like: [Array.Prototype.Every](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every)

adreyfus:
Travis failed because of changes made by the recently merged PR #311.

src/requestPool/requestPoolManager.js
adreyfus: Done :)

src/requestPool/requestPoolManager.js
adreyfus: Done :)

dannyrb:
I like this PR, but I think because of its importance we should write a few more tests for it. I'm adding it to my list to pull down this branch, play with it, and write a few tests against it.

Thanks for all of your hard work @adreyfus

adreyfus:
- addRequestPoolType (name, priority, maxRequests) is renamed putRequestPoolType (name, priority, maxRequests), and now removes already existed same named type
- added a test for this function

adreyfus:
The function getIndex seems to be to complex for codeclimate, because it removes the already existed same named type and return the priorityIndex, by browsing the array only one times.

Do you have suggestions to decrease the Cognitive Complexity ?

dannyrb:

FYI, @adreyfus, I particularly like this PR. Unfortunately its sat here for a very long time do to mixed priorities. If this is important to you, then I think this is something we now have a bit more bandwidth to address.

PR #334, opened 2018-01-17
Feature/export events

adreyfus:
related to issue #333


dannyrb:
Making this public means we need to be careful about changing them, but they feel pretty static to me. I'll pull the merge lever. Thanks @adreyfus

dannyrb:
It might be doing this with the events we've surfaced in `core`. Thoughts, @medihack?

PR #336, opened 2018-01-21
Allow inputs to be enabled before loading an image. Fixes #317

### LINKED ISSUE #317: Allow input types to be enabled on elements that have not yet loaded/displayed an image
In many examples the input sources are enabled straight after the element is enabled, e.g. [in the probe tool](https://github.com/cornerstonejs/cornerstoneTools/blob/master/examples/probe/index.html#L68). This throws an error when no image is displayed (or while it is loading) and the event is triggered on the element (e.g. the mouse cursor is moved there) as the input sources depend on an already loaded image, because of the `cornerstone.pageToPixel` function.

Two options:
- Change the examples to enable input sources after the image was loaded (to make this more clear when to enable those).
- Check in the event listeners if there is an attached image to the enabledElement and if not then just return.

I would prefer the second option. What do you think?

medihack:


swederik:
Thanks!

PR #337, opened 2018-01-28
Fix addEventListener bug in examples

### LINKED ISSUE #307: Seed point example addEventListener bug
I forgot to switch this when changing .on to .addEventListener. We can't use multiple event types with addEventListener.

https://github.com/cornerstonejs/cornerstoneTools/blob/29182180ed3f850ba41c609b98b96464affca5f0/examples/seedPoints/index.html#L252

lord:
Fixes #307

### DIFF
--- a/examples/alltouchtools/index.html
+++ b/examples/alltouchtools/index.html
@@ -180,7 +180,7 @@
       });
     });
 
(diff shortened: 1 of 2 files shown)


swederik:
Thanks!

PR #344, opened 2018-02-05
Expanded freehand ROI tool

JamesAPetts:
I have fixed some issues with the freehand ROI tool, and added statistical calculations and some nice UI/UX features:

- Added general area calculation for irregular polygons of N sides.
- Finds pixels inside the freehand polygon by means of an even-odd rule algorithm, and calculates the mean and stdev based on the pixels found.
- Displays these stats in a textbox in a similar fashion to the rectangular and ellipsoidal ROI tools. The text box is disabled for the shift-click 'pencil' mode.
- Disallows the user to join up to any previous point besides the starting node (i.e. shape must be complete before tool deactivates).
- Disallows the user to cross lines (using a orientation algorithm).
- Fixed a UI bug, where upon editing the first node the line connecting the first and last nodes would not reactively update.
- The textbox-polygon tether snaps to nearest node of the polygon when moved.

PR #345, opened 2018-02-06
Freehand tool bugfix

artemShelest:
Freehand tool has a bug, reproducible from the example page:
1. Draw a closed-loop shape of more than 2 points.
2. Click on the first point of the shape to activate modification mode.
3. Move this point away from its original position.
4. Observe the line still starting from the old position of the point.


swederik:
Thanks for this!

There is another freehand PR open here, actually: https://github.com/cornerstonejs/cornerstoneTools/pull/344
I'll try to find time to check if that one fixes this issue as well.

artemShelest:
Seems it does not. I can merge after you integrate #344

swederik:
Hi @artemShelest,

@galelis split the commit out and now it's in master. Thanks!

dannyrb:
@swederik, you won't see the changes in the live examples until you publish a new release. We want the examples to reflect the stable/release.

Local dev will show changes if you build and have a dist.

We can talk about a next branch if you'd like that has examples that always reflect unpublished/@next source code?

swederik:
Yeah lets do that, it would be good to at least be able to point to working live examples once fixes are merged.

PR #348, opened 2018-02-13
Fix memory leak from use of get/setToolOptions

### LINKED ISSUE #309: Memory leak with set/getToolOptions
There is an issue with the new set / get tool options module which was used to help replace jQuery's passing of additional data (e.g. mouse button mask) into event handlers. It never clears the data, not even if the element or tools are disabled.

Relevant code is here:

https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/toolOptions.js

Options:
- We could set up a 'clearToolOptions' function to be run when a tool is disabled
- We could attach a listener for 'cornerstoneelementdisabled' to clear all element-specific data when an element is disabled
- We could update cornerstone core to incorporate an optional callback to run when an element is disabled, and we could register clearToolOptions to run in this callback.

Not exactly sure what's cleanest, so I'm open to suggestions, and of course pull requests, should anyone wish to fix this themselves!

nspin:
Fixes #309


dannyrb:
@nspin thanks for this! Almost positive this will get merged here soon. Bonus points if I can persuade you to write unit tests to cover this new functionality?

swederik:
Looks good! Thanks! As @dannyrb said, if you feel like adding some tests and / or jsdoc comments, we would be very happy to merge that as well.

nspin:
For sure - I'll write some tests and jsdoc.

Issue #338, opened 2018-01-30
Bug when activating stackScroll before stackScrollTouchDrag at the same time.


medihack:
When activating `stackScrollTouchDrag` after(!) `stackScroll` then `stackScroll` (using the mouse) is not working anymore (only the touch one is working). When activating them the other way around (`stackScrollTouchDrag` before `stackScroll`) both work fine. I didn't get that error with other tools yet (e.g. `wwwc`). Not a critical bug, but an annoying one ;-)

dannyrb:
Thanks for reporting this. I'll see if I can duplicate it in a codepen and post the example here. I'll mark this as a `verified bug` after I can reproduce it.

medihack:
Is there a preconfigured Codepen template for testing such things?

medihack:
I created a Codepen that shows the issue: https://codepen.io/medihack/pen/mXVYdX
If you simply swap lines 23 and 24 then both work fine again (tested on a Microsoft Surface Pro).

amitpanwar123:
I am unable to scroll XA images in cornerstone. Used URL like : Dicom//201712221750420360/Standered/1.2.840.114257.1.16356791.20175100.dcm?frame=2

vaibhav2383:
I have created a pull request that fixes this issue. https://github.com/cornerstonejs/cornerstoneTools/pull/352

swederik:
Closing because I believe @vaibhav2383 fixed this.

PR #352, opened 2018-02-20
Bug when activating stackScroll before stackScrollTouchDrag at the same time. #338

vaibhav2383:
Fix for Issue #338.

Both _stackScroll_  & _stackScrollTouchDrag_ were using the same _toolName_ for setting _options_. 
This was creating a condition of _options_ being overridden.


swederik:
Thanks!

Issue #296, opened 2017-11-27
Zoom Enhancement: Add invert option to mouse and touch tool (like zoomWheel's invert)


philem35:
hi all, 
I tried to set the 'invert' option to true on Zoom tool.
```
    let config = {
      invert: true
    };
cornerstoneTools.zoom.setConfiguration(config);
```
The change only impacts scroll events and not mouse down events, 

is there a trick to change the way zoom behave on mouse move ?
thanks, 

PS : there's a lack of documentation in cornerstone tools, not very easy to know the good configuration

dannyrb:
Hi @philem35, 

It looks like only the zoomWheel used the config value `invert` . You can see that here: 
https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/imageTools/zoom.js#L238

You might be able to accomplish what you want by modifying the zoom tool's dragcallback here:
https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/imageTools/zoom.js#L227

Or by modifying your zoomStrategy of choice to account for config.invert property.

If you make a PR, I'd be happy to merge. I'll mark this issue as an enhancement item that is up for grabs.

dannyrb:
All zoom tools should now have an invert option. It's worth noting that there is an issue with zoom wheel caused by inconsistent scroll direction across browsers: Chrome v. FF

If you're having issues setting the invert option in `vNext`, feel free to create a separate documentation issue, question, or PR

PR #353, opened 2018-02-22
Added invert option to zoom with mouse drag

posegae:
Small change to address #296. 

Now inverting the config changes how zoom behaves on mouse drag events, in addition to mouse wheel events.


Saucistophe:
Any news on this PR? I'd need it, inverting the mousewheel only is not enough for us!

PR #357, opened 2018-02-27
Just update dist files (yarn run build)

medihack:
It seems dist files were not rebuilt after latest commits. This commit just rebuilt them (`yarn run build`).


swederik:
Thanks!

PR #360, opened 2018-03-01
Use pixelSpacing from MetaDataProvider in ROI tools

zachasme:
This fetches pixel spacing metadata from metadata provider, in order to get area in mm in OHIF.


swederik:
Should be merged in now, thanks!

PR #362, opened 2018-03-04
Make overlays unselectable by using CSS. Fixes #351

### LINKED ISSUE #351: Let the native events bubble up
There are times when I want to be informed of some native browser events that come from an enabled element inside my app, but this is blocked by `pauseEvent`. An example would be to activate a specific viewer in a layout with many viewers (that all have enabled elements with images in them) when the mouse is down. To fetch these events they must be allowed to bubble up from an enabled element resp. its canvas.

In the [`pauseEvent` file](https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/util/pauseEvent.js#L4) it is mentioned that the bubbling of events is stopped to prevent text element selection. Wouldn't it make more sense to use CSS for this? E.g. see [this StackOverflow answer](https://stackoverflow.com/a/924937/166229).

I see two solutions:
- Delete the `pauseEvent` stuff altogether and let the user handle this with CSS (maybe add a helper function to cornerstoneTools like `makeUnselectable(element)`).
- Add an options parameter to `mouseInput` and `touchInput` to disable this behavior (e.g. `mouseInput.enable(element, {letEventsBubble: true})`).

medihack:


swederik:
Thanks!

Issue #364, opened 2018-03-06
Ability to import individual tools from the module, without pulling in all files


dannyrb:
**Problem:** Our documentation currently recommends importing `cornerstoneTools` like so:

```
// Cornerstone Libraries
import * as cornerstone from 'cornerstone-core'
import * as cornerstoneMath from 'cornerstone-math'
import * as cornerstoneTools from 'cornerstone-tools'
```

Which has the unfortunate side-effect of including all of `cornerstoneTools` in your output bundle. Minified, that's just under 200KB. Unminified? over 500KB.

**Ideally, we want:**

- To make it easier to decrease output bundle size
- To know which parts of `cornerstoneTools` your application depends on
- To increase developer familiarity with the library's namespacing
- To make it easier to leverage advanced bundler features (splitting, dynamic imports, tree shaking, etc)

**Okay, but how do we accomplish this?**

_Here are a few examples of potential ways to enable two elements, and then activate the `wwwc` and `wwwcTouchDrag` tools for each element:_

_Current Approach:_ 
```
import * as cornerstone from 'cornerstone-core'
import * as cornerstoneMath from 'cornerstone-math'
import * as cornerstoneTools from 'cornerstone-tools'

// Create two enabled elements/canvases
cornerstone.enable(myEnabledElement)
cornerstone.enable(myEnabledElement2)

// Setup tools for first element
cornerstoneTools.mouseInput.enable(myEnabledElement)
cornerstoneTools.wwwc.activate(myEnabledElement, 1)
cornerstoneTools.wwwcTouchDrag.activate(myEnabledElement)

// Setup tools for second element
cornerstoneTools.mouseInput.enable(myEnabledElement2)
cornerstoneTools.wwwc.activate(myEnabledElement2, 1)
cornerstoneTools.wwwcTouchDrag.activate(myEnabledElement2)
```

_New Approach 1:_ Uses a per instance config/settings to define behavior
```
import * as cornerstone from 'cornerstone-core'
import * as cornerstoneMath from 'cornerstone-math'
import CornerstoneTools from 'cornerstone-tools'

// Contains registration logic for wwwcMouse and wwwcTouchDrag
import { wwwc } from 'cornerstone-tools/core-tools'

cornerstone.enable(myEnabledElement)
cornerstone.enable(myEnabledElement2)

const csTools = new CornerstoneTools({
    // csTools.activateTool() applies to all enabledElements
    snycToolStatesAcrossEnabledElements: true,
    
[...]

PR #368, opened 2018-03-12
Feature/ability to import individual tools from the module #364

dannyrb:
@swederik, this is _NOT_ ready for merge. I still have a few articles to read to make sure I'm taking the correct approach here. Feel free to weigh in with thoughts. Biggest changes:

- Using `webpack`'s `alias` feature

_ Tool Changes_
- Moved all tools to own folder, and removed them from top level export
- Created a separate entry point and output file for tools
    - `import { angleTool } from 'cornerstone-tools/tools'`

_Top level API Changes_
- Moved all top level exports to a default export:
    - `import cornerstoneTools from 'cornerstone-tools'`

If we did this as an intermediary step, it would feel very similar, but with the added ability to import the specific tools you want. I think there are a large number of other improvements we _could_ make during this process, but then the balance between _incremental changes_ and reducing the frequency of _breaking changes_ begins to get weird.

I do think we need to remove a large number of our top level exports. I think there's also a great opportunity here to do some of the refactoring/optimizations that we've talked about.

This relates to #364


dannyrb:
Adding some notes here from Slack:

> 1. Shouldn’t use Webpack-only features, some community members want to use native ES6 modules
> 2. Agree that manipulators (anyHandlesOutsideImage, drawHandles, etc..) don’t need to be in top level API
> 3. Agree, we can move inputSources down a level as well. I don’t think anyone really depends on them

-- @swederik 


> @kofifus out of curiosity, why do you rely on native es6 modules? Is there a particular reason you can't use the transpiled output or consume cornerstone as an npm dependency? What problem are you trying to solve?

-- Me

> I see a few steps:
> 
> 1) We can import only the tools we want and not the whole bundle. We want this for Cornerstone Core as well. This could even be 2.2 I guess.
> 2) We start handling activation / deactivation / etc optionally inside the library. (sort of like OHIF’s tool manager)
> 3) API Refactoring for 3.x

-- @swederik 

*Instead of another Major release:*
_Use a pre-release tag until we hit an API target_

- https://medium.com/@mbostock/prereleases-and-npm-e778fc5e2420
- https://gist.github.com/schmich/d04acc4b02b45e489f329cfdf3280a3f
- https://github.com/semantic-release/semantic-release

kofifus:
> -- @swederik
> 
> @kofifus out of curiosity, why do you rely on native es6 modules? Is there a particular reason you can't use the transpiled output or consume cornerstone as an npm dependency? What problem are you trying to solve?
> 

This is a general issue but anyways I think this is the wrong question - why do we rely on webpack etc when developing when we can directly run the code as is in the browser ? why go through an unnecessary 'compilation' phase and loose the power of the browser's debugger with its breakpoints and profiling etc and in general make development much more difficult when the browser can load and run the code as is ?

dannyrb:
- Existing example appear to be working as expected :+1:
- Publishing to `@next` to test individual tool import
    - importing individual tools is working :+1:

**Example:**

_Old way, still works:_
```
import * as cornerstoneTools from 'cornerstone-tools'
...
cornerstoneTools.wwwc.activate(myEnabledElement, 1)
```

_New way:_
```
import cornerstoneTools from 'cornerstone-tools'
import { wwwc } from 'cornerstone-tools'
...
wwwc.activate(myEnabledElement, 1)
```


**Changes of note:**
- [src/index.js](https://github.com/dannyrb/cornerstoneTools/blob/e7edea55ba1bb54c8122bd3b93916b0e9b0b0751/src/index.js)
- Updated `.eslintrc` to resolve module paths
- Created a `tools` folder that groups tools by tool type
    - Includes a `base` folder

**For this PR, it might be worth looking at:** (CC: @swederik)

- Additional directory structure changes
    - CC: @medihack (https://github.com/cornerstonejs/cornerstoneTools/issues/376)
- Limiting the scope of our public API
    - IE. default exports in `src/index.js`
- Building in a deprecation warning for `import * as` usage


It's worth noting that while we continue to preserve the old way of consuming `cornerstone-tools`, tree shaking and dead code elimination doesn't aide much. I'll post examples here shortly.

kofifus:
Hi

Please avoid `import cornerstoneTools from 'cornerstone-tools'` that is an import statement without a path or extension. 

import's "module-name" _must_ resolve to a file. When using native browser imports, 'cornerstone-tools' does _not_ point to a valid file and will break the loader. Please _only_ use ie `import cornerstoneTools from '.\cornerstone-tools\index.js' etc

Thanks!

dannyrb:
This _was_ working before the merged updates. Doing this on the fly so others can touch the code. It's worth noting, importers of our library will have trouble treeshaking unless we follow the advice here:

https://webpack.js.org/guides/author-libraries/

> The module property should point to a script that utilizes ES2015 module syntax but no other syntax features that aren't yet supported by browsers or node. This enables webpack to parse the module syntax itself, allowing for lighter bundles via tree shaking if users are only consuming certain parts of the library.

PR #370, opened 2018-03-16
Chore/get dist out of repo

dannyrb:

* **Other information**:

This PR removes `dist` from the repo, and update our published module to contain almost only the `dist/` directory.

I haven't tested this, so don't merge it yet.

### Note

It looks like a lot of our templates deviate from each other in simple ways. It might be nice to use a simple static site generator or something similar so our template is reusable. It's silly to update our example's header tags in 30+ places, or the CDN target for a CSS file.


dannyrb:
The few examples I've tried seem to be working, so I'm going to pull the lever on this. If it blows up, it'll be on a per-tool example basis, and I'm more than happy to patch them.

@swederik I think I'll play with templating the examples so we have a little more consistency. Thinking about using Webpack's `handlebars-loader`. Let me know if you have thoughts

PR #371, opened 2018-03-19
Add in viewport provider for stack navigation

jdnarvaez:
**Feature**

I wanted a way to have a factory or service manage what viewports are used for display of a set of images, so that changes can be remembered within a session or presentation states can be applied at display time, and I wanted the tooling to respect that as well, and this seemed like the easiest way to do that. 

```
const viewportProvider = {
    getViewport : function(enabledElement, image, stackData) {
      //return viewport....
    }
};

// set the stack as tool state
cornerstoneTools.addStackStateManager(element, ['stack', 'viewportProvider']);
cornerstoneTools.addToolState(element, 'stack', stack);
cornerstoneTools.addToolState(element, 'viewportProvider', viewportProvider);
```

Currently `cornerstone.displayImage` is simply called with no viewport so a default one is created by cornerstone. The code as is should not break anything as if no viewport provider is registered, the `displayImage` method is called with the same parameters s before.

### DIFF
--- a/src/util/scrollToIndex.js
+++ b/src/util/scrollToIndex.js
@@ -40,11 +40,13 @@ export default function (element, newImageIdIndex) {
       return;
     }
 
+    let enabledElement;
+
     // Check if the element is still enabled in Cornerstone,
     // If an error is thrown, stop here.
     try {
       // TODO: Add 'isElementEnabled' to Cornerstone?
-      cornerstone.getEnabledElement(element);
+      enabledElement = cornerstone.getEnabledElement(element);
     } catch(error) {
       return;
     }
@@ -53,7 +55,9 @@ export default function (element, newImageIdIndex) {
       stackRenderer.currentImageIdIndex = newImageIdIndex;
       stackRenderer.render(element, toolData.data);
     } else {
-      cornerstone.displayImage(element, image);
+      const viewportProvider = getToolState(element, 'viewportProvider');
+      const viewport = viewportProvider && viewportProvider.data && viewportProvider.data.length > 0 ? viewportProvider.data[0].getViewport(enabledElement, image, stackData) : undefined;
+      cornerstone.displayImage(element, image, viewport);
     }
 
     if (endLoadingHandler) {


dannyrb:
Let me see if I follow. Currently, every time you display an image you can either:

- Not specify a `viewport`, and the current `viewport` for the `enabledElement` is used.
- Specify a `viewport`, and it is used/merged with the `enabledElement`'s `viewport`.
(see: https://github.com/cornerstonejs/cornerstone/blob/master/src/displayImage.js#L42-L48)

You want to add the ability to associate a `viewport` with an `enabledElement`, single `image`, or stack of `image`s? You could use this a number of ways, but your PR specifically updates `scrollToIndex` to check for a matching `viewport` in the provider to pass to `displayImage`

I'm wondering if this makes more sense to have in `cornerstone-core` and used inside of `displayImage`? (similar-ish to `metaDataProvider`). I also don't know much about presentation states, but it would be interesting to see the `stackTool` add a `viewport` to the `viewportProvider` based on its stack's tags.

As viewports are merged in `displayImage`, I think you'd need to be careful to not set `viewport` data in the `viewportProvider` that would override anything the user would still expect to control? IE. `scale`, or `rotation`?

Thinking out loud. Feel free to drop some knowledge bombs on me.

jdnarvaez:
As far as the attributes of `viewport` being overwritten, you're correct. With this strategy the consumer would be expected to manage whatever the overall stack state is for those attributes if they are provided. I'm doing this so that as users view different series, any transformations they made to a series previously will still be there when they come back to it within the same session. The idea here is to associate a `viewport` with a particular `image` for a provided `stack` within a single viewing session, but not necessarily an `enabledElement`. The `enabledElement` is passed as a parameter so that if the `viewportProvider` needs to perform any calculations for display (scale, translation, etc) based on the viewport size, it can. 

As for presentation states, it's sort of the same idea. Unless the user has changed the attributes of the `viewport`, then the presentation state should provide all of that display transformation information as they navigate through it.

It may make sense to have this in `cornerstone-core`, but I opted to leave it out of it since it seemed like the `displayImage` function was basic enough in that if someone was going to do attribute state management for session the could go about it in a way like this. 

The idea of associating a viewport to an imageId might be a good one, unless a user were wanting to compare, say, the same image in two different `enabledElements` with different display values (window comparison, etc) then that may not be unique enough. If there are better ways, I'm all for it.

jdnarvaez:
however, the more I think about it, the more it seems like perhaps registering a `viewportProvider` for a specific `enabledElement` makes sense. I would imagine that provider would still want a stack context when being asked to create a `viewport` for a particular `image`.

dannyrb:
Does `presentation state` ever contain more information than the current `viewport` object? Are you thinking it would also contain some extra information that tools might need to accomplish displaying `presentation state`? (Making it kind of a mashup of `viewport` and [`stackSpecificStateManager`](https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/stateManagement/stackSpecificStateManager.js)

My only concern is that if we want this consistently applied _everywhere_ `displayImage()` is called in a tool, then it makes more sense (too me) to modify `displayImage`.

I think all of this makes sense, but I think we'd need the following before we could merge:

- A default/example `viewportProvider`
- A new example page showing a "real life" application of the provider
- Unit tests for new public API methods on `cornerstone-tools`

It might also be worth discussing:

- What methods should _all_ `viewportProvider`s expose?
- Determining priority/order when more than one is registered?
- How tools should interact with/use `viewportProvider`s?

Sorry to expand on this so much. I just want us to be extra deliberate with public API changes, and the ones we do make, I want to be well documented so others can properly take advantage of them.

jdnarvaez:
the presentation state would contain more than the viewport object. it can contain graphic annotations, look up tables, text annotations, shutters, different presentation properties, etc. I think we would want it displayed everywhere for display consistencies sake, so perhaps it makes sense to add a method to the metadata provider to provide a viewport as well? I imagine a default implementation would simply call getDefaultViewport from within cornerstone.

dannyrb:
@swederik, you self-assigned this a while back. What do we need to discuss or review to action this PR?

dannyrb:

PR #373, opened 2018-03-21
Free hand improvements

galelis:
Improvements and new functionalities

New functionalities:
- Added general area calculation for irregular polygons of N sides.
- Added textbox to show data from the area of the polygons of N sides, it does not affect 'pencil' mode 
- The textbox-polygon tether snaps to nearest node of the polygon when moved.

Improvements:
- Disallows the user to join up to any previous point besides the starting node (i.e. shape must be complete before tool deactivates).
- Disallows the user to cross lines (using a orientation algorithm).

Tests: 
All the old functionalities and the new ones are working fine.

### DIFF
--- a/src/imageTools/freehand.js
+++ b/src/imageTools/freehand.js
@@ -348,10 +348,18 @@ function mouseHover (eventData, toolData) {
       imageNeedsUpdate = true;
     }
 
-    if ((pointNearHandleAllTools(eventData) && !data.active) || (!pointNearHandleAllTools(eventData) && data.active)) {
-      data.active = !data.active;
-      data.highlight = !data.highlight;
-      imageNeedsUpdate = true;
+    let isPointNearHandle = pointNearHandle(eventData, i);
+
+    if (isPointNearHandle === 0) {
+      // JPETTS - if returns index 0, set true as otherwise shows up falsy for condition bellow.
+      isPointNearHandle = true;
+    }
+
+    if ((isPointNearHandle && !data.active) || (!isPointNearHandle && data.active)) {
+      if (!data.lockedForEditing) {
+        data.active = !data.active;
+        imageNeedsUpdate = true;
+      }
     }
 
     if (data.textBox === true) {
--- a/src/imageTools/freehand.js
+++ b/src/imageTools/freehand.js
@@ -3,6 +3,13 @@ import external from '../externalModules.js';
 import toolStyle from '../stateManagement/toolStyle.js';
 import toolColors from '../stateManagement/toolColors.js';
 import drawHandles from '../manipulators/drawHandles.js';
+import drawTextBox from '../util/drawTextBox.js';
+import handleActivator from '../manipulators/handleActivator.js';
+import pointInsideBoundingBox from '../util/pointInsideBoundingBox.js';
+import freeHandArea from '../util/freeHandArea.js';
+import calculateFreehandStatistics from '../util/calculateFreehandStatistics.js';
+import { freeHandIntersect, freeHandIntersectEnd, freeHandIntersectModify } from '../util/freeHandIntersect.js';
+import calculateSUV from '../util/calculateSUV.js';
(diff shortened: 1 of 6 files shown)


galelis:
I have noticed only one possible issue: 

After dragging the textbox around the image:
- sometimes close to the borders, it disappears and only appears again after moving the mouse
- sometimes you change the position and it does not stick to the exactly position. (it moves a little bit)

I was unable to reproduce these issues on the Rectangle ROI tool.

galelis:
I will check the codeclimate issues.

hu3402379:
@galelis 
hi,  galelis:
Thank you for your contribution，when i test it, i find a bug:
1. I add the polygons of N sides, and I can move it when I drag the point,
2. then I change other function, like click wwwc button, then I can not move the polygons.
and if i draw a rectagleRoi or elipticalRoi, when i click other button, when my mouse is on rectagleRoi or elipticalRoi point, i can move it and redraw it.  

i think it's a bug for freehand. 

how do you think?

thanks for your contribution！

best regards

src/util/freeHandIntersect.js
```diff
@@ -0,0 +1,185 @@
...
+  const p1 = {
+    x: dataHandles[k].x,
+    y: dataHandles[k].y
+  };
+  const q1 = {
+    x: newHandle.x,
+    y: newHandle.y
+  };
+
+  if (doesInteresctOtherLines(dataHandles, p1, q1, k)) {
```
swederik: Typo, "Interesct" should be "Intersect"

src/util/freeHandIntersect.js
```diff
@@ -0,0 +1,185 @@
...
+  const p1 = {
+    x: dataHandles[k].x,
+    y: dataHandles[k].y
+  };
+  const q1 = {
+    x: dataHandles[l].x,
+    y: dataHandles[l].y
+  };
+
+  if (doesInteresctOtherLines(dataHandles, p1, q1, k, l)) {
```
swederik: rewrite as:

return doesInteresctOtherLines(dataHandles, p1, q1, k, l)

src/util/pointInFreehandROI.js
```diff
@@ -0,0 +1,66 @@
+// JPETTS - Calculates if "point" is inside the polygon defined by dataHandles by
```
swederik: Can you switch this to a comment on the actual function, rather than in the file?

src/imageTools/freehand.js
```diff
@@ -386,12 +558,239 @@ function onImageRendered (e) {
       fill: fillColor
     };
 
-    if (data.active) {
+    if (data.active && !data.polyBoundingBox) {
```
hu3402379: how about add  drawHandlesOnHover:
```    
if (config && config.drawHandlesOnHover) {
      // Draw the handles if the tool is active
      const handleOptions = {
        drawHandlesIfActive: true
      };

      if (data.active && !data.polyBoundingBox) {
        drawHandles(context, eventData, config.mouseLocation.handles, color, handleOptions);
      }
      // If the tool is inactive, draw the handles only if each specific handle is being
      // Hovered over
      drawHandles(context, eventData, data.handles, color, handleOptions);
    } else {
      if (data.active && !data.polyBoundingBox) {
        drawHandles(context, eventData, config.mouseLocation.handles, color);
      }
      // If the tool has no configuration settings, always draw the handles
      drawHandles(context, eventData, data.handles, color);
    }

galelis:
PR is done. 

#Issue reported by @hu3402379
The scenario regarding not being able to perform any action once the FreeHand ROI is deactivated, its actually a functionality not developed yet for this tool so it will not be part of this PR. I spoke with Erik, and it will not be done as part of this PR. 

#Improvement by adding drawHandlesOnHover
As a new functionality, it will not be done as part of this PR.

JamesAPetts:
Thanks for all the hard work getting this in a more presentable state, it was actually some of the first JavaScript I've written, so I'll take note to your improvements!

Kind regards,
JamesAPetts

PR #375, opened 2018-03-25
Improved default strategy of rotate. Fixes #359

### LINKED ISSUE #359: Non intuitive rotation
Rotation is IMHO currently very unintuitive (mouse and touch input). When the user holds and drags somewhere in the image the rotation "jumps" dependent on how far the pointer is away from an imaginary 12 o'clock line. It would be more natural if the initial button down determines the zero line and every move is relative to this line.

An example to illustrate this:

The current way:
- Click and hold mouse button at 3 o'clock position of the image.
- Move mouse a bit while holding the mouse button
- Image rotation jumps by 90° (plus/minus some degrees dependent on the movement)

The improved way:
- Click and hold mouse button at 3 o'clock position of the image.
- Move mouse a bit while holding the mouse button
- Image rotates just a few degrees depending on the pointer movement

I would be happy to come up with a PR for that, just need some time.

medihack:

Fixes #359 

See #359

See #359 

No

* **Other information**:
I had to introduce a `simpleTouchTool` function as `touchDragTool` does too many assumptions (as the `dragCallback` must have the initial rotation and mouse location added by using `bind`).

### DIFF
--- a/src/imageTools/rotate.js
+++ b/src/imageTools/rotate.js
@@ -137,11 +137,7 @@ rotate.strategies = {
 
 rotate.strategy = defaultStrategy;
 
-const options = {
-  fireOnTouchStart: true
-};
-
-const rotateTouchDrag = simpleTouchTool(touchStartCallback, toolType, options);
+const rotateTouchDrag = simpleTouchTool(touchStartCallback, toolType);
 
 export {
   rotate,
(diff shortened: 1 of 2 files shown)


swederik:
You're right, I just tested this and it feels much more intuitive.

Thanks!

Issue #182, opened 2017-02-07
Magnifying glass doesn't use full resolution image


hardmaster92:
I recently noticed that the zoom of the magnification glass is not a real zoom. It just gets the area you click and resizes it with interpolation. If you use the zoom tool to zoom out the image and make it really small and then use the magnify tool it all gets pixelated.

jpambrun:
The canvas itself is used as pixel source here: 

https://github.com/chafey/cornerstoneTools/blob/a28f8973360c3072487880c4f6fb9bb82b91eb79/src/imageTools/magnify.js#L131

I agree with @hardmaster92, it is not very useful in this state.

diego0020:
I have an rough implementation of a magnifying glass that uses the full resolution image as source in my fork. If you are interested I can clean it up and make a PR

dannyrb:
Oooh. I'd actually love a link to this? I'm curious to see the path you
took to implement.

diego0020:
here it is 
https://github.com/diego0020/cornerstoneTools/blob/master/src/imageTools/magnify2.js

diego0020:
I basically render the image again in an offscreen canvas of the correct size, and then copy the pixels to the magnifying glass view

swederik:
I'd be happy to review a PR for this if you make one

dannyrb:
This has been merged and fixed :+1:

PR #378, opened 2018-04-03
Improvement on Magnify Tool.

galelis:
Its a BUG FIX
https://github.com/cornerstonejs/cornerstoneTools/issues/182

Magnifying glass doesn't use full resolution image

This version of the magnify tool renders the image in the target scale in a different element, and then copies pixels from it, therefore it can reach up to the resolution of the original image.

No

* **Other information**:
This PR comes from https://github.com/cornerstonejs/cornerstoneTools/pull/286
In the time the PR was created, we had DIST folder and also we were dependent on jQuery. 
I had to add changes in order to not use jQuery.


swederik:
FYI, while we experiment with Netlify, you have to add /examples to the deploy preview link: 
https://deploy-preview-378--cornerstone-tools.netlify.com/examples/

Issue #387, opened 2018-04-13
Tools 


floryst:
### Prerequisites

* [x] Are you running the latest version?
* [x] Are you reporting to the correct repository?
* [x] Did you perform a cursory search?

For more information, see the `CONTRIBUTING` guide.

### Description

Tools that are hidden using the `visible` flag are still interactive (assuming tool is in active/passive state). This shouldn't happen since interacting with invisible stuff is confusing.

This follows from adfcc573bd17ad2bbfb5775d44e078a1ee27743b.

### Steps to Reproduce

This is assuming you are using the codepen below.

1. Add a length tool to the image
2. Hide the tool by checking the checkbox on the right
3. Click and drag where the length tool handles used to be

**Expected behavior:**

A new tool appears since visually no tool handle is at the interaction point.

**Actual behavior:**

No new length tool appears, and the hidden length tool is updated instead.

### Codepen With Reproduction of Issue:

- https://codepen.io/anon/pen/zWVOjg

dannyrb:
Oh wow. Weird.
We're _slowly_ moving to consolidating the tool state change logic, which should make a refactor that accounts for this easier.

The issue may live in one or more of the base tools not removing the appropriate listeners on a state change.

swederik:
Thanks for the report and PR! Now that it's merged, I guess we can close this.

floryst:
Thanks!

PR #388, opened 2018-04-13
fix(interaction) No tool interaction when tool is invisible

floryst:

If tool data has the `visible` flag set to false, then `pointNearTool` will return false. This is to prevent mouse interaction with the tool when the data is invisible.

#387

Clicking on an invisible tool will not update the invisible tool.

No

* **Other information**:


swederik:
Yeah this makes sense. I managed to totally overlook that.... 

Thanks!

PR #394, opened 2018-04-30
Feature: Add a .color attribute to the tool data

timleslie:
This change adds a `.color` attribute to the tool data for those tools which currently have a `.active` attribute. If `.active` is true and `.color` is defined, then the tool data will be displayed using the custom color. If `.active` is true and `.color` is undefined then the tool data will be displayed using the default active color (`activeColor`). If `.active` is false, then the default tool color (`white`) will be used.

This is my first contribution to the project, so please advise if this change is not appropriate or could be achieved in a better way.

This PR partially address the backlog item from `README.md`: 

> Config object that allows tool appearance to be customized (e.g. line color, text color, handle size, shape, etc)

### DIFF
--- a/src/imageTools/angleTool.js
+++ b/src/imageTools/angleTool.js
@@ -17,6 +17,7 @@ function createNewMeasurement (mouseEventData) {
   const angleData = {
     visible: true,
     active: true,
+    color: undefined,
     handles: {
       start: {
         x: mouseEventData.currentPoints.image.x - 20,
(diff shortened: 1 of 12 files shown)


swederik:
Hmm this is a good idea. I wonder if people would also want to set the default tool color on a tool-by-tool basis. What do you think?

timleslie:
@swederik I think that's definitely something worth considering. I have a number of other changes in the pipeline which depend on this PR going through so perhaps we could save that question for a future PR later this week?

PR #398, opened 2018-05-07
Add a drawLinkedTextBox function

timleslie:
The new `drawLinkedTextBox` consolidates common code for drawing a text box which is linked to an image tool.

This PR also introduces the `textBoxAnchorPoints` function to each appropriate image tool, which abstracts out the code for computing possible anchor points for the link to the text box.


swederik:
Thanks!

PR #399, opened 2018-05-07
Add a lineSegDistance function

timleslie:
This utility function simplifies the logic involved in the `pointNearTool` function for a number of image tools.


swederik:
Thanks!

PR #401, opened 2018-05-08
added cornerstonetoolsmouserightclick event

lastlegion:

Exposes right click event

Nothing. 

Triggers `MOUSE_RIGHT_CLICK` event  when right clicked on a shape. 

No.

* **Other information**:


pieper:
LGTM 👍

lastlegion:
Hmm I didn't think about using `mouseclick` and then using `event.which` or `event.buttons`. Right now the only use case I can think of is the contextmenu key on windows keyboards.

swederik:
I think that one is covered by cornerstonetoolskeydown/press/up (https://github.com/cornerstonejs/cornerstoneTools/blob/bb1189ac7e20896846dbf906d8e612d396acf3fa/docs/essentials/input-sources.md#keyboard-input-source). Doesn't seem worth maintaining another event if it only has that function.

Sorry! Don't let this discourage you from continuing to contribute!

PR #402, opened 2018-05-11
Freehand pencil

JamesAPetts:
This is kind of a big one so bare with me!

Demo: https://deploy-preview-402--cornerstone-tools.netlify.com/examples/freehandroi/

========New features========
The ability to add or remove handles after completion:
-- Ctrl + clicking on a line will insert a handle there.
-- Ctrl + clicking on a handle will delete that handle, provided there are more than 3 handles.

Revamped Pencil mode:
-- Shift + click will once again enter pencil mode, however the resulting object will be the same data structure as the manual polygon mode: a set of handles by which statistical calculations can be calculated (more suitable for storage and e.g. export to, RT-STRUCT or AIM imageAnnotationCollections).
-- After one shift+click the user can move the mouse freely to draw the ROI.
-- If the user crosses lines, the ROI will light up as its invalidColor and no points will be placed until the mouse is moved to a valid position.
-- When the user moves the mouse back to the origin of the ROI, the handle will increase in size indicating that the user can click to complete the ROI (This has also been added to manual mode).

Enhanced Freehand Configuration:
-- spacing: The minimum spacing between points in pencil mode (default= 5).
-- activeHandleRadius:  The handle radius when the tool is being edited (default = 3),
-- completeHandleRadius: The size the handle appears when the ROI can be completed.(default = 6),
-- alwaysShowHandles: When not in use, the handles of the ROI are not shown by default, such that the user sees just the contours. However, if alwaysShowHandles is set to 'true', the handles are always displayed (default = false),
-- invalidColor: When drawing/editing an ROI if the mouse is placed in an incorrect position, such that the new line would intersect another, the ROI flashes up a warning color (default = 'crimson').

Other:
--manipulators/drawHandles.js can now draw a custom handle size if options argument contains the property 'handleRadius'. This is not a breaking change as if the property is not set the old default behavior is used.
--Unit tests added for pointInFreehand.js, freehandArea.js, freehandIntersect.js and calculateFreehandStatistics.js.

Bug fixes:
-- PR #404 didn't return anything in textBoxText, resulting in the textbox displaying undefined. This has been fixed.

======Changes/Refactoring=======
I've done some major refactoring of the tool, as it already significantly more complex than the other region of interest tools.
-- The freehand.js file contains mostly event handling, with the implementation of abstracted functionality found in in utils/freehand/. I thought a sub-directory is necessary as to not clog up the utils directory, but this could be changed should the team desire. Athough this isn't perfect yet I will refactor more as I expand the tool.
-- Renamed all instances of "node" to "handle" to stay consistent with the nomenclature across cornerstoneTools.
-- Renamed all variable/function instances of "freeHand" to "freehand".

I hope you'll all be excited about this one as much as I am!


JamesAPetts:
Note I'll get on Docs etc in a few weeks, but I am currently skimped for time whilst preparing for a conference.

JamesAPetts:
P.S. Something we need in here soon is some way of cancelling the creation of an ROI, (e.g. right click whilst drawing). Especially with the pencil mode, one can get in a pickle with some ease. This shouldn't take that long to implement and again I'll work on this when I'm back from ITCR.

swederik:
In OHIF we use the Esc key to cancel drawing something, maybe that could work? I don't know if I want to include it in the tool itself though.

JamesAPetts:
> In OHIF we use the Esc key to cancel drawing something, maybe that could work? I don't know if I want to include it in the tool itself though.

Ahh, right, I didn't know that functionality exists, I'll look into it.

> Ctrl + Click to place a new point is maddeningly difficult.. Maybe we could make it easier somehow?

I'm guessing the issue is the handles blinking on and off? In e.g. OHIF I was planning to simply have a toggle for alwaysShowHandles for editing, but how about if when the ctrl key is held down all handles are shown anyway?

Or was it is a different issue? Feedback would be welcome.

Thanks,
James

JamesAPetts:
Thanks @diego0020 and @swederik, nice catch. This is an easy fix, i'm going to turn my attention to this after the conference this week.

Thanks,
James

JamesAPetts:
Note this branch is currently a little busted whilst I write some tests and resolve these issues.

swederik:
Do you have the browser focus on the proper element?  I don't think you'll receive the events otherwise.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

JamesAPetts:
The browser focus must be on the proper element, as the mouse interface works fine. I tried explicitly forcing focus, but this didn't fix the issue. [The test source can be found here](https://github.com/JamesAPetts/cornerstoneTools/blob/freehand_pencil/examples/freehandRoi/index.html)

JamesAPetts:
Hey @swederik, @diego0020,

I've fixed the issue diego found (thanks!) and made it such that holding ctrl pops up all the handles of ROIs, such that it is easier to modify them with the add/remove point ctrl+click functionality.

Also added a load of jsdocs and a few unit tests.

JamesAPetts

JamesAPetts:
Updated with recent commits to master branch, and fixed an issue introduced in PR #404:
-- PR #404 didn't return anything in textBoxText, resulting in the textbox displaying undefined. This has been fixed.

I feel this PR is complete, so i'm going to make another branch for the next bit of functionality.
Welcome to more feedback/criticism.

Cheers,
JamesAPetts

JamesAPetts:
@swederik V1.0 of the freehandSculpter tool is tested and ready for submission. I'll wait for this PR to be accepted before submitting another that builds upon it.

Cheers,
JamesAPetts

src/util/freehand/dropObject.js
```diff
@@ -0,0 +1,87 @@
...
+
+    // Snap back to previous position
+    currentHandleData.x = config.dragOrigin.x;
+    currentHandleData.y = config.dragOrigin.y;
+    previousHandleData.lines[0] = currentHandleData;
+
+    dataHandles.invalidHandlePlacement = false;
+  }
+
+  return;
```
swederik: Pointless return?

src/util/freehand/FreehandHandleData.js
```diff
@@ -0,0 +1,29 @@
+/**
+* @author JamesAPetts
```
swederik: Could we remove the `@author` tags? Git already tracks who did the work.

JamesAPetts: No problem, I've just got into this habbit when jsdoc-ing, I'll remove them.

src/util/freehand/dropObject.js
JamesAPetts: Woops, don't know how he got there, good catch.

swederik:
Thanks! Looks good!

PR #404, opened 2018-05-15
Add a textBoxText function to all tools which use drawLinkedTextBox

timleslie:
Moving the code for generating the text box text into a separate function for each tool reduces the complexity of the `onImageRendered` function and also makes the common patterns across tools more obvious.

My aim with this collection of refactorings is to have the `onImageRendered` for each tool follow the same template, with the tool-specific details factored out into separate functions (like `textBoxText`).

The motivation here is to make it simpler to maintain our own custom tools.


swederik:
Thanks! I really appreciate all the cleanup and design work you're putting in! 👍

Issue #405, opened 2018-05-17
New drawing library - drawing.js


timleslie:
# Introduction

One of the fundamental concerns of `cornerstoneTools` is drawing markings on top of an image canvas.

The fundamental object used to achieve this is the [HTML Canvas 2D Context](https://www.w3.org/TR/2dcontext/) (or `context` for short).

### 1. Required knowledge

In order to write a new tool which draws to the canvas, the developer must familiarise themselves with the concept of a `context` and learn the API and how to correctly use it.

Directly using the `context` API involves a large amount of boiler plate code to correctly initiate the context, save/restore the context state stack, and begin/end paths.

#### Example: Drawing a line between two points
```
    context.save();
    context.beginPath();
    context.strokeStyle = color;
    context.lineWidth = lineWidth;
    context.moveTo(start.x, start.y);
    context.lineTo(end.x, end.y);
    context.stroke();
    context.restore();
```

The above example demonstrates how to draw a line between two points with a given color and width. The developer needs to know about 6 different API methods and 2 API attributes in order to achieve a simple task.

### 2. Interleaving the *what* and the *how*

The verbose `context` API encourages a pattern of interleaving the calculation of *what to display* with the API calls saying *how to display it*. For example

```
    context.save();
    context.beginPath();

    const color = getColor();

    context.strokeStyle = color;

    const lineWidth = getLineWidth();

    context.lineWidth = lineWidth;

    const start = getStart();

    context.moveTo(start.x, start.y);
  
    const end = getEnd();
   
    context.lineTo(end.x, end.y);

    context.stroke();
    context.restore();
```

This makes it difficult to follow exactly what is being done on the canvas, as the single conceptual item (draw a line) has been interspersed with other lines of code.

### 3. Duplication

Working directly with the `context` API means that we have a large amount of duplication, as every time we want to draw a line (or any other common pattern), we need to write the same 8 lines of code. 

## Proposed Solution

In order to address these issues, I am proposing to introduce a module in `util/drawing.js` which will provide a simple API which matches the existing use cases of `cornerstoneTools`.

This module will be used internally and and also exposed as part of the public interface to allow developers to write custom tools more easily.

#### `getNewContext(canvas)`

Create a new `context` object and set the transform to the [identity transform](https://www.w3.org/TR/2dcontext/#transformations). 

#### `draw(context, fn)`

[...]

PR #406, opened 2018-05-17
[0/N] drawing.js API

timleslie:
See https://github.com/cornerstonejs/cornerstoneTools/issues/405 for full discussion.

This PR provides a prototype implementation of the `drawing.js` API as well as a complete refactoring of `cornerstoneTools` which takes advantage of the new API.

This branch is for demonstration and reference purposes only. It can be expected to contain bugs and regressions. The changes here will be broken into a series of smaller PRs which can be individually reviewed and merged.


dannyrb:
@timleslie are we at a point with the Drawing API changes that this is safe to close?

PR #407, opened 2018-05-17
[1/N] drawing.js API

timleslie:
This change introduces the `drawing.js` API. This API simplifies interaction with the `context` API, allowing for simpler tool building.

This PR changes the public API of `drawCircle` and `drawEllipse`. As such, it should bump at least the minor version number when released.

See #405 for full discussion.


timleslie:
Hi @swederik, when you get a chance could you have a look at this and let me know if you think it's heading in the right direction?

swederik:
Yes! Sorry it's taken so long

src/index.js
```diff
@@ -6,6 +6,20 @@ export { default as orientation } from './orientation/index.js';
 
 export { default as requestPoolManager } from './requestPool/requestPoolManager.js';
 
+export {
```
swederik: Maybe we should export them as cornerstoneTools.drawing.xyz ?

src/util/drawCircle.js
```diff
@@ -1,3 +1,4 @@
+// @deprecated Use drawing.js:drawCircle()
```
swederik: Could you use jsdoc comments (http://usejsdoc.org/about-getting-started.html#adding-documentation-comments-to-your-code)? I think it needs /* instead. That way this function will start to get a strikethrough in people's code editors (e.g. WebStorm).

src/util/drawing.js
```diff
@@ -0,0 +1,175 @@
+import external from '../externalModules.js';
+import { toolStyle, textStyle } from '../index.js';
+
+export function getNewContext (canvas) {
```
swederik: JSDoc for all of these functions would be great. That helps people use the API because their editor will auto-suggest the arguments. It will also be auto-generated and displayed here: https://tools.cornerstonejs.org/api.html

(yeah I know it needs some reorganization...)

src/util/drawing.js
```diff
@@ -0,0 +1,175 @@
...
+    context.fill();
+  }
+  context.stroke();
+  if (lineDash) {
+    context.setLineDash([]);
+  }
+}
+
+export function setShadow (context, options) {
+  if (options.shadow) {
```
swederik: This seems a bit weird. If I was using this function I would assume it would just be:

````javascript
options: {
    shadowColor: '#000',
    shadowOffsetX: 1,
    shadowOffsetY: 1
}
````

rather than a subobject with .shadow: { ... } 

Thoughts?

Comment (swederik) [reply src/util/drawing.js]: Ah right you're including the whole if block. OK I guess it's fine

File: src/util/drawCircle.js (hunk 2 above)
Comment (dannyrb) [reply src/util/drawCircle.js]: Worth noting, some editors have plugins that allow you to right click and autogenerate `jsdoc` stubs. IE.https://marketplace.visualstudio.com/items?itemName=joelday.docthis

File: src/index.js (hunk 1 above)
Comment (timleslie) [reply src/index.js]: Done.

File: src/util/drawCircle.js (hunk 2 above)
Comment (timleslie) [reply src/util/drawCircle.js]: Done.

File: src/util/drawing.js (hunk 3 above)
Comment (timleslie) [reply src/util/drawing.js]: Done.

File: test/coverage_test.js [hunk 5]
```diff
@@ -6,5 +6,7 @@ import * as cornerstoneTools from '../src/index';
 describe('A test that pulls in all modules', function () {
   it('pulls in all modules', function () {
     expect(cornerstoneTools).to.exist;
+    expect(cornerstoneTools.drawing).to.exist;
```
swederik: I don't think this stuff is necessary. That "test" is just a way to make Istanbul include all the files when it makes the coverage report.

swederik:
Live link is here: https://deploy-preview-407--cornerstone-tools.netlify.com/examples

timleslie:
Thanks for the feedback @swederik, I've updated the JSDoc with much more explicit object types, which are generating links to the appropriate online docs, which is nice!

swederik:
Thanks!

PR #425, opened 2018-06-27
[2/N] Use the getNewContext() function from drawing.js

timleslie:
This PR uses the `getNewContext()` function in all places where we need a `context` object.

There is one exception, in [`referenceLinesTool.js`](https://github.com/cornerstonejs/cornerstoneTools/blob/3e587198a6d1038cb3c1a240d1e6906cf2214536/src/referenceLines/referenceLinesTool.js#L25). This function does not follow the pattern of creating a `context` and immediately settings the identity transform. Refactoring this particular use case will be left for a future task, and may involve modifying either the behaviour of this tool.

See #405 for full discussion

### DIFF
--- a/src/imageTools/magnify.js
+++ b/src/imageTools/magnify.js
@@ -4,6 +4,7 @@ import { getBrowserInfo } from '../util/getMaxSimultaneousRequests.js';
 import isMouseButtonEnabled from '../util/isMouseButtonEnabled.js';
 import { setToolOptions, getToolOptions } from '../toolOptions.js';
 import { clipToBox } from '../util/clip.js';
+import { getNewContext } from '../util/drawing.js';
 
 const toolType = 'magnify';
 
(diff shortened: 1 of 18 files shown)


src/imageTools/magnify.js
```diff
@@ -111,13 +112,7 @@ function drawMagnificationTool (eventData) {
...
   // No classes of it's own on the canvas we want to select
   const canvas = element.querySelector('canvas:not(.magnifyTool)');
-  const context = canvas.getContext('2d');
-
-  context.setTransform(1, 0, 0, 1, 0, 0);
-
-  const zoomCtx = magnifyCanvas.getContext('2d');
-
-  zoomCtx.setTransform(1, 0, 0, 1, 0, 0);
+  const zoomCtx = getNewContext(magnifyCanvas);
```
dannyrb: I'm just doing a quick pass over. Were you able to drop `context` because it's not used? I was under the impression that the magnify tool used two separate context's for rendering (that may have been the old version)

timleslie: Yep, the other `context` is unused so can be safely removed.

swederik:
LGTM :-)

PR #430, opened 2018-07-02
[4/N] Use the path function from drawing.js

timleslie:
This PR replaces calls to `context.beginPath()` and `context.stroke()` with calls to the `path()` function from the drawing.js API.

See #405 for full discussion.

### DIFF
--- a/src/util/drawEllipse.js
+++ b/src/util/drawEllipse.js
@@ -20,6 +20,5 @@ export default function (context, x, y, w, h) {
     context.bezierCurveTo(xm + ox, y, xe, ym - oy, xe, ym);
     context.bezierCurveTo(xe, ym + oy, xm + ox, ye, xm, ye);
     context.bezierCurveTo(xm - ox, ye, x, ym + oy, x, ym);
-    context.closePath();
   });
 }
--- a/src/util/drawEllipse.js
+++ b/src/util/drawEllipse.js
@@ -1,3 +1,5 @@
+import { path } from './drawing.js';
+
 // http://stackoverflow.com/questions/2172798/how-to-draw-an-oval-in-html5-canvas
 
 /**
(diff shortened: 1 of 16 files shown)


src/util/drawEllipse.js
```diff
@@ -12,12 +14,12 @@ export default function (context, x, y, w, h) {
...
-  context.bezierCurveTo(xm - ox, ye, x, ym + oy, x, ym);
-  context.closePath();
-  context.stroke();
+  path(context, {}, (context) => {
+    context.moveTo(x, ym);
+    context.bezierCurveTo(x, ym - oy, xm - ox, y, xm, y);
+    context.bezierCurveTo(xm + ox, y, xe, ym - oy, xe, ym);
+    context.bezierCurveTo(xe, ym + oy, xm + ox, ye, xm, ye);
+    context.bezierCurveTo(xm - ox, ye, x, ym + oy, x, ym);
+    context.closePath();
```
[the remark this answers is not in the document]
timleslie: Good catch. Fixed.

src/util/drawLink.js
```diff
@@ -29,11 +30,12 @@ export default function (linkAnchorPoints, refPoint, boundingBox, context, color
...
-  context.beginPath();
-  context.strokeStyle = color;
-  context.lineWidth = lineWidth;
-  context.setLineDash([2, 3]);
-  context.moveTo(start.x, start.y);
-  context.lineTo(end.x, end.y);
-  context.stroke();
+  const lineDash = [2, 3];
+
+  path(context, { color,
```
swederik: Could we make the options into a separate variable? Seems like it might be cleaner to read

timleslie: Yep, I agree it would make it clearer. It might be easier to do that clean up after the next couple of PRs, which will convert many of these `path` calls into calls to things like `drawLine`. Also, a lot of these option objects will reduce down to `{ color }` once we take account the default value of `lineWidth`.

timleslie:
Hey @swederik, are we able to merge this PR, or would you like to see some more changes?

swederik:
Ah whoops, forgot about this one after the review. Seems fine now!

PR #431, opened 2018-07-03
Freehand sculpter

JamesAPetts:
This PR introduces a new tool, the freehandSculpter, which can manipulate freehand ROIs. The demo speaks for itself: [Click here!](https://deploy-preview-431--cornerstone-tools.netlify.com/examples/freehandsculpter/).

In words: One can ctrl+click an ROI and then begin editing it by holding down the mouse. A circle will appear around the mouse with a radius equal to the distance from the mouse to the nearest point on the ROI. Moving the mouse will push the edges of the ROI, creating and removing new points as needed. Releasing the mouse will complete the edit. By starting an edit with a mouse close to the edge of the ROI, fine adjustments can be made. By clicking far away/near the center of the ROI, one can make coarser, rough adjustments.

@dlwire , this PR also exposes pointNearTool for you to use!

Sorry this one took a while to get out. Its been 99% ready (and actually in use) for a while, but hadn't had time to get it up to PR standard!


JamesAPetts:
P.S. This is pretty similar to Osirix's repulser tool, in that you can mess up an roi if you aren't being precise (try editing and moving the mouse around the whole screen fast). This just emphasises how good a global deleting tool (ala @dlwire suggested) would be.

I tried adding limitations to stop self-intersection, similar to what I did with the freehand tool itself, but any way I tried felt too restrictive. I think it feels better to just let the user do what they want in this case and be able to delete and start over.

.gitignore
```diff
@@ -1,7 +1,6 @@
 ## Directorties
 node_modules
 coverage/
-dist/
```
swederik: Could you remove this change?

JamesAPetts: Oops, didn't mean to leave that in, will remove it when I'm back in the office tomorrow!

JamesAPetts: Done.

swederik: Could you also remove the checked-in dist files below? Then I think it's fine to merge

src/util/freehand/ClickedLineData.js
```diff
@@ -0,0 +1,22 @@
+/**
+* @author JamesAPetts
```
swederik: Any chance you can remove this too?

swederik:
Looks good! Thanks!

PR #432, opened 2018-07-04
Cornerstone#245 Remove default 0 for fractional seconds of seriesAcquisitionTime since dicomParser handles it

ayselafsar:
See https://github.com/cornerstonejs/cornerstone/issues/245

### DIFF
--- a/src/util/calculateSUV.js
+++ b/src/util/calculateSUV.js
@@ -45,7 +45,7 @@ export default function (image, storedPixelValue) {
     return;
   }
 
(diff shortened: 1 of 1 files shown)


dannyrb:
I don't know that we can guarantee that all metaData is coming from `dicomParser`, right? I'm not sure it's worth removing the `default` 0, when it could be a breaking change for a custom image loader.

swederik:
I agree with @dannyrb. There is no need to change this. We don't know where metadata comes from.

PR #433, opened 2018-07-05
feat(eraser) initial implementation of eraser

dlwire:

Adds an eraser that works on tool markings.

There isn't one. ;)

Adding this tool will allow users to delete other tool's markings by clicking on them.

No.

* **Other information**:
I'm not really certain where to update docs at this point. Let me know what you'd like to see added and I'll get after it.

On the events front, I didn't inherit from mouseButtonTool because this is a lot smaller than the other tools. Also, the MOUSE_DOWN event seems to get swallowed by the first tool that catches it and so never makes it to the eraser tool. This way we can use MOUSE_CLICK which makes it to the eraser.

I imported all of cornerstoneTools so we can just iterate over everything there and operate on any tools with the pointNearTool function.

Finally, I'm not able to test on mobile but there's a good chance it will have the same issue using MOUSE_DOWN does. I imagine those events are dispatched and handled much like MOUSE_DOWN.

I'm sure there's plenty of feedback on this attempt. Happy to update to a more idiomatic approach.

### DIFF
--- a/examples/allImageTools/index.html
+++ b/examples/allImageTools/index.html
@@ -37,6 +37,7 @@
                 <a id="angle" class="list-group-item">Angle</a>
                 <a id="highlight" class="list-group-item">Highlight</a>
                 <a id="freehand" class="list-group-item">Freeform ROI</a>
+                <a id="eraser" class="list-group-item">Eraser</a>
             </ul>
             <div class="checkbox">
               <label><input type="checkbox" id="chkshadow">Apply shadow</label>
@@ -83,6 +84,7 @@
                         <li>Rectangle ROI - A rectangular ROI that shows mean, stddev and area</li>
                         <li>Highlight - Darkens the image around a rectangular ROI</li>
                         <li>Angle - Cobb angle tool</li>
+                        <li>Eraser - Erases marks from other tools</li>
                     </ul>
                 </li>
                 <li>Use the activated tool by dragging the left mouse button</li>
@@ -156,6 +158,7 @@
         cornerstoneTools.rectangleRoi.enable(element);
         cornerstoneTools.angle.enable(element);
         cornerstoneTools.highlight.enable(element);
+        cornerstoneTools.eraser.enable(element);
 
         activate("enableWindowLevelTool");
 
@@ -181,6 +184,7 @@
             cornerstoneTools.angle.deactivate(element, 1);
             cornerstoneTools.highlight.deactivate(element, 1);
             cornerstoneTools.freehand.deactivate(element, 1);
+            cornerstoneTools.eraser.deactivate(element, 1);
         }
 
         // Tool button event handlers that set the new active tool
@@ -234,6 +238,11 @@
             disableAllTools();
             cornerstoneTools.freehand.activate(element, 1);
         });
+        document.getElementById('eraser').addEventListener('click', function() {
+            activate('eraser');
+            disableAllTools();
+            cornerstoneTools.eraser.activate(element, 1);
+        })
     });
 
 
--- a/src/imageTools/eraser.js
+++ b/src/imageTools/eraser.js
@@ -1,0 +1,62 @@
+import * as cornerstoneTools from '../index.js';
+import EVENTS from '../events.js';
+import external from '../externalModules.js';
+
+const toolType = 'eraser';
+
+function eventListeners (event) {
+  return {
+    activate: function (element) {
+      element.addEventListener(event, deleteNearbyMeasurement);
+    },
+    deactivate: function (element) {
+      element.removeEventListener(event, deleteNearbyMeasurement);
+    },
+    disable: function (element) {
+      element.removeEventListener(event, deleteNearbyMeasurement);
+    },
+    enable: function (element) {
+      element.removeEventListener(event, deleteNearbyMeasurement);
+    }
+  }
+}
+
+function pointNearTool () {
+  return false;
+}
+
+function deleteNearbyMeasurement (mouseEventData) {
+  const coords = mouseEventData.detail.currentPoints.canvas;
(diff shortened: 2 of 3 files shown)


dannyrb:
Nice work! 👍 

We have a similar tool at Asteris. It really hits you how nice it would be if tool data was managed in a more global way, so that other tools could modify/play with any tool's data more easily. The same goes for useful utility functions for existing data (ie. pointNearTool).

The biggest differences that jump out at me between this implementation and ours (not necessarily better/worse, just different):

1. We leveraged the `simpleMouseButtonTool` and `simpleTouchTool` to make sure our tool worked (and continues to work) in a similar way as other simple tools:
    - [simpleMouseButtonTool](https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/imageTools/simpleMouseButtonTool.js)
    - [simpleTouchTool](https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/imageTools/simpleTouchTool.js)
2. Instead of iterating over every `key` on `cornerstoneTools`, we have an array of string tool names that we use to check specific tools using `pointNearTool` to limit iterations and for slightly tighter control.

> ...the MOUSE_DOWN event seems to get swallowed by the first tool that catches it and so never makes it to the eraser tool. This way we can use MOUSE_CLICK which makes it to the eraser.

We had a similar issue. Our ToolManager sets all of the other `active` tools to `enabled` when the `eraserTool` is active, so you can view their annotations, but they can't be interacted with.

*Misc. Thoughts:*
These are mostly things we don't do, mostly because the existing API makes it a bit cumbersome, but it would be cool if:

- The eraser tool _highlighted_ the tool data it would delete before clicking
    - You may accomplish this with your alternate implementation; it's a downside of us changing the state of other active tools.
- There were different strategies for deleting:
    - "Top" tool
    - Nearest tool
    - All tools within radius X

Well. That's what's on my 🍜 
Would love to hear your thoughts.

dlwire:
Thanks for all the great feedback! Your tool sounds similar to the one we're running as well. I wrote it outside cornerstoneTools and monkey patched it in. It allowed me easier access to our tool manager, which I'm using to get the list of tools that we care about to erase. I like the ideas here, and thought about some of them but thought this implementation was more usable without having to dig into too much.

1. simple*Tool inheritance - I like the idea of using this and using the same mouse event for triggering. Unfortunately, I think for users new to the system understanding that they have to enable the tools rather than deactivate them might be a bit frustrating to discover.
2. list of tools to delete - I thought about using the config to manage a list of tools that would be deleted on click. My initial implementation actually worked that way. I moved to the key method because I thought most folks would just want it to delete the marks on the screen and configuring it like that isn't exactly intuitive.
3. new highlight - If we deactivate all the other tools, nothing will get highlighted if I understand everything correctly. So something that goes through and finds tool data near the mouse on mouse move would be worthwhile.

@dannyrb - What do you think about making the following changes:
* Do all the investigation looking for tools with `pointNearTool` exposed to set an initial configuration - which could be overridden with `setConfig`
* Using that list of tools `enable` and `deactivate` all the tools we care about as the eraser tool is activated/deactivated
* Inherit from simple[MouseButton, Touch]Tool
* [stretch] highlight the tool data which will be deleted on click

Assuming I don't encounter anything terribly unexpected I could probably update the PR reasonable quickly.

dannyrb:
Great thoughts, @dlwire (: 

> list of tools to delete - I thought about using the config to manage a list of tools that would be deleted on click. My initial implementation actually worked that way. I moved to the key method because I thought most folks would just want it to delete the marks on the screen and configuring it like that isn't exactly intuitive.

That makes a lot of sense. You also gain the advantage of the eraserTool continuing to work with new tools that expose the right methods without needing an explicit update. I do see the advantages :+1:

I'm a fan of merging PRs that satisfy the minimum requirements and iterating. I think the only thing I'd like to see before this is merged would be:

> - Inherit from simple[MouseButton, Touch]Tool

I think the following are nice/simple enhancements:

> - Do all the investigation looking for tools with pointNearTool exposed to set an initial configuration - which could be overridden with setConfig
> - Using that list of tools enable and deactivate all the tools we care about as the eraser tool is activated/deactivated

And while I like highlighting, I think it might be worth looking into some of the API issues that make it difficult to implement before tackling it. (I'd still merge an implementation, but I'd caution that most implementations would likely be brittle against upcoming API changes)

dlwire:
Just noticed I'm missing the check for whether pointNearTool is exposed. Let me clear that up before this gets merged.

dlwire:
Unless there are other comments, I reckon this is ready to go. One thing I encountered while testing a refactor to only sift through cornerstoneTools once is that the ellipticalROI tool's `pointNearTool` function doesn't seem to behave quite how one would expect but we can probably tackle that at a later date.

dannyrb:
I mentioned this PR in the Slack group. I'll give @swederik and any other until EOD Monday to comment, but if I don't hear anything, I'll pull the lever and merge.

Thanks for the great discussion and hard work @dlwire! (:

JamesAPetts:
@dlwire, hey, nice simple approach.

As far as highlighting goes, can't we just let the tools internal state sort this out? "Deactive" tools have this functionality, already. Can't you just replace lines 244-258 of index.html with deactivate rather than enable?

dannyrb:
@dlwire if you send me an e-mail @ danny.ri.brown@gmail.com, I'll pass you an invite to our core contributor Slack group?

dannyrb:
@dlwire thanks for your hard work! We're going to merge this. I think any tool that can delete annotations has been sorely needed. If you have any suggestions around what other features or strategies the tool should employ, please feel free to create an issue. Especially if you think we should work toward a library level change to make that feature easier to accomplish 👍

PR #434, opened 2018-07-10
Brush Changes

JamesAPetts:

New: This PR folds the layer management required for the brush tool to function into the tool itself, so that it can be .e.g. imported straight into the OHIF viewer, plug and play like other tools.

Bug fix: I also fixed a bug whereby if the brush tool was deactivated, a snapshot of the current mask layer would remain fixed on the screen, even if you changed image slices.

Additionally there is a keyboard interface for changing tools with "{" and "}" and changing brush size with "+" and "-". I'm happy to change these keybindings upon request.

There are currently 19 colors to choose from, although I wholeheartedly admit this is rather arbitrary for now.

This shouldn't break anything since the brush tool wasn't in a usable state before.

However any 3rd party tools that uses the cornerstone layers API should take note that we cannot guarantee the layers will load in your desired order at this stage.

EDIT: [Demo up.](https://deploy-preview-434--cornerstone-tools.netlify.com/examples/segmentationbrush/)


examples/segmentationBrush/index.html
```diff
@@ -110,63 +114,95 @@ <h5>Controls:</h5>
...
-  cornerstone.enable(element);  
+  cornerstone.enable(element);
   cornerstone.loadImage(imageIds[0]).then(function(image) {
-    cornerstone.addLayer(element, image);
-    cornerstone.updateImage(element);
-    var viewport = cornerstone.getViewport(element);
-    viewport.pixelReplication = true;
-    cornerstone.setViewport(element, viewport);
+    // TEMP -- This line demonstrates the issue experienced in the OHIF viewer. Ideally displayImage would be called first and then the tool deals with adding an additional layer.
+    cornerstone.displayImage(element, image);
```
swederik: // TEMP... I guess we should get rid of this..

JamesAPetts: Fixed.

swederik:
Looks good! Thanks!

PR #436, opened 2018-07-11
[5/N] Use the setShadow function from drawing.js

timleslie:
This PR replaces blocks of code setting the `context` shadow parameters with calls to the `setShadow()` function from the `drawing.js` API.

See #405 for full discussion.


swederik:
LGTM, thanks!

PR #438, opened 2018-07-11
[6/N] Use the drawLine, drawLines and drawJoinedLines functions from drawing.js

timleslie:
This PR replaces calls to `path`, `moveTo` and `lineTo` with calls to `drawLine`, `drawLines`, and `drawJoinedLines` from the `drawing.js` API.

See #405 for full discussion.


swederik:
Just tested the tools involved and everything seems fine. Thanks!

PR #443, opened 2018-07-12
Freehand pointNearTool refactor

JamesAPetts:
Simply refactors the pointNearTool function in the freehand tool so that it takes the same parameter arguments as other tools, for use with the eraser tool and for longer term compatibility.


swederik:
Looks good to me (but I didn't test it)

PR #453, opened 2018-07-13
chore/add first set of basic tests

dannyrb:
This breaks the implementation because it moves us to setters/getters, but no one except for me is using this branch yet _shrug_


dannyrb:
image

PR #463, opened 2018-07-15
[9/N] Use the drawRect function from drawing.js

timleslie:
This PR replaces calls to `context.rect()` with calls to the `drawRect()` function from the drawing.js API.

See #405 for full discussion.


swederik:
Thanks!

PR #475, opened 2018-07-18
Add a textBoxWidth function

timleslie:
This PR adds a `textBoxWidth` function to `drawTextBox.js`. It abstracts away the common operation of determining how wide the textbox needs to be for a given text and padding.

This change also ensures that the width is calculated in terms of the currently selected font. Some tools did not previously do this and could potentially calculate the wrong value if the default font of the `context` was different to the font specified in `textStyle` (although I suspect this is a fairly rare situation).

This change also allows us to avoid setting `context.font` within each tool, as all the fonts are now managed by helper functions. We also remove setting of `context.fillStyle`, since the `drawing.js` API handles this for us.

Computing the text width with explicit reference to the `padding` brings us closer to having the padding defined once for each tool, rather than explicitly and implicitly in multiple places throughout the code. Future PRs will follow this line of investigation.

### DIFF
--- a/src/imageTools/angleTool.js
+++ b/src/imageTools/angleTool.js
@@ -3,7 +3,6 @@ import mouseButtonTool from './mouseButtonTool.js';
 import touchTool from './touchTool.js';
 import drawTextBox from '../util/drawTextBox.js';
 import roundToDecimal from '../util/roundToDecimal.js';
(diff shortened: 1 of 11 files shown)


swederik:
Thanks! I really like the tool definitions becoming more concise :-)

Issue #464, opened 2018-07-16
Allow an easy enabling/disabling of synchronizers


Saucistophe:
### Description

According to the demo, and having had a look at the code, I think there is no easy way to disable a Synchronizer without manually removing viewers.
In my case, I handle viewers on the go, removing them, adding them to the view, etc. The trouble is we have a "Synchronize" button that toggles the synchronization between viewers.
Instead of removing all viewers and adding them all back, and registering new viewer only if the Synchronized mode is active, it would be way easier to add them systematically to the synchronizers and toggling their status from disabled to enabled:

`panZoomSynchronizer.enabled = false;`

If you think that's a valid idea and/or that there are no alternative, I'd be happy to submit a PR.

dannyrb:
The solution for this was merged, so I will close the issue.

PR #480, opened 2018-07-20
Added a setting to deactivate synchronizers.

Saucistophe:
The default behavior of Synchronizers is unchanged.
Toggling the _enabled_ boolean on a Synchronizer instance toggles the synchronization, so that one can easily desynchronize viewers without having the need to remove all viewers from its scope.
Related to issue #464


swederik:
Makes sense to me. Thanks!

PR #488, opened 2018-07-25
WIP: Migrating freehand tool to v3, and working on API utility for more complicated tools.

JamesAPetts:
Notes: I left 'isAwaitingMouseUp' with its current name for now, as there are a lot of dev branches, I'll review and tidy up the eventDispatcher when we have all the tools in.

J


examples/00-test/index.html
```diff
@@ -51,15 +51,17 @@ <h1>
...
 			<div class="canvas" oncontextmenu="return false" />
 		</div>
 	</div>
 </body>
 
-<!-- include the hammer.fakemultitouch so we can test with a mouse -->
+<!-- include the hammer.fakemultitouch so we can test with a mouse
 <script src="../touch-emulator.js"></script>
 <script>TouchEmulator();</script>
+-->
```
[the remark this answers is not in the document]
JamesAPetts: Hey, yeah I have no touch interface for the freehand tool, so I disabled it for testing. We could add a button to the demo UI to switch it on or off?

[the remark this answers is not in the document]
dannyrb: I think we can safely remove this altogether. The majority of tools now support touch and mouse, so we shouldn't need to emulate touch events to make sure the tools are working. It may even cause issues to attempt emulation.

Can you

dannyrb:
Preview URL: https://deploy-preview-488--cornerstone-tools.netlify.com/examples/00-test/

src/base/baseAnnotationTool.js
```diff
@@ -51,6 +51,17 @@ export default class extends baseTool {
...
+
+  /**
+  * OPTIONAL -- Used to check if there is a valid target for the tool, that
+  * isn't necessarily its own toolData. (e.g. the freehandSculpter)
+  *
+  * @param {*} evt
+  * @returns {Boolean} - True if the target is manipulatable by the tool.
+  *
+  * isValidTarget (evt) {}
+  */
```
dannyrb: Thanks for adding this 👍

src/eventDispatchers/mouseToolEventDispatcher.js
```diff
@@ -19,7 +19,7 @@ import getToolsWithDataForElement from './../store/getToolsWithDataForElement.js
 
 const cornerstone = external.cornerstone;
 
-// Todo: better place for this
+// TODO: better place for this
```
dannyrb: Your OCD is showing :P

src/eventDispatchers/mouseToolEventDispatcher.js
```diff
@@ -97,6 +97,12 @@ function mouseMove (evt) {
         data.active = !data.active;
         imageNeedsUpdate = true;
       }
+
+      // Call the tool's mouseMoveCallback if it exists.
+      if (typeof tool.mouseMoveCallback === 'function') {
```
dannyrb: 💯

src/eventDispatchers/mouseToolEventDispatcher.js
```diff
@@ -136,6 +143,23 @@ function mouseDown (evt) {
   tools = tools.filter((tool) =>
     isMouseButtonEnabled(eventData.which, tool.options.mouseButtonMask)
   );
+
+  // Check if any tool has a special reason to grab the current event
+  for (let t = 0; t < tools.length; t++) {
+
+    if (typeof tools[t].mouseDownCallback === 'function' &&
+        typeof tools[t].isValidTarget === 'function' &&
+        tools[t].isValidTarget(eventData)) {
```
dannyrb: This should probably take `event`.

src/util/freehand/keysHeld.js
```diff
@@ -23,7 +22,7 @@ const keyCodes = {
 */
 export function keyDownCallback (e) {
   const eventData = e.detail;
```
dannyrb: I get why this exists here, but I think we need to shift it up to be a bit more global. I'd like if we tracked which keys were down in `cornerstone-tool`s store, so other tools can use that information in their logic/callbacks without needing to do this themselves.

dannyrb: Maybe each tool has an array of keycodes it cares about, and then the store/dispatcher tracks all keycodes that all added tools collectively care about? (that might be overkill, just noodling)

src/imageTools/freehand.js
```diff
@@ -631,7 +631,6 @@ function modifyObject (e, nearby) {
   const eventData = e.detail;
```
dannyrb: You can probably delete this file if the ported tool covers its use case?

src/util/freehand/keysHeld.js
dannyrb: Are these only fired if keyboard input is enabled? I'm having trouble getting `ctrl + click` behavior to work (I may not understand fully)

src/fancy-tools/freehandTool.js
```diff
@@ -0,0 +1,1143 @@
...
+
+  /**
+  * Adds commas as thousand seperators to a Number to increase readability.
+  *
+  * @private
+  * @static
+  * @param {Number|String} number - A Number or String literal representing a number.
+  * @return {String} - A string literal representaton of the number with commas seperating the thousands.
+  */
+  static _numberWithCommas (number) {
```
dannyrb: `rectangleRoi` and `ellipticalRoi` have this same method. We should move it to shared/utils and import.

src/fancy-tools/freehandTool.js
```diff
@@ -0,0 +1,1143 @@
...
+  /**
+  * Event handler for called by the mouseDownActivate event, if tool is active and
+  * the event is not caught by mouseDownCallback.
+  *
+  * @event
+  * @param {Object} evt - The event.
+  */
+  addNewMeasurement (evt, interactionType) {
+    if (interactionType === 'touch') {
+      console.warn('No touch controls implemented for freehandTool.');
```
dannyrb: This should never be called, as the `touchEventDispatcher` will filter out tools that don't have `touch` in their list of supported interaction types

dannyrb:
I'm wondering if it's possible for it to work without adding/removing event listeners
I'm a little torn, because I think it makes it more efficient
Is the logic/use case just too complex to use callbacks?
Or is it more about locking in to that event loop once you've started to use the tool?

I'm going to take break from reviewing this.
I _really, really, really_ like the end product, but I think there are some points at least discussing before merging.

examples/00-test/index.html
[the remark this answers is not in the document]
JamesAPetts: Done.

src/eventDispatchers/mouseToolEventDispatcher.js
JamesAPetts: Agreed, changed.

src/util/freehand/keysHeld.js
JamesAPetts: Apologies for not making a note of this, but this functionality hasn't actually been ported yet, and isn't in use by the v3 tool. However, to explain: it does require the keyboard input to be enabled to function. This interface is just to highlight all nodes when ctrl is held and config.alwaysShowHandles is false. The ctrl + click to add and remove points uses the eventdata from the mouseDown event to check if ctrl was used. This later functionality is in the PR.

Please let me know if you wish to have any points clarified :).

src/imageTools/freehand.js
JamesAPetts: Yeah, I just left it in for now for quick personal reference. Now I've had a review I'll delete it (y).

src/fancy-tools/angleTool.js
```diff
@@ -307,6 +307,11 @@ export default class extends baseAnnotationTool {
     addToolState(element, this.name, measurementData);
     cornerstone.updateImage(element);
 
+
+    // TEMP
+    this.configuration.drawing = true;
```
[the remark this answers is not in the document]
JamesAPetts: Oops, this was placed here whilst testing, hence the big TEMP flags. I'll remove it for now, but maybe some active flag in config for all tools is something to think about.

src/fancy-tools/freehandTool.js
```diff
@@ -0,0 +1,1143 @@
...
+// Implementation Logic
+import pointInsideBoundingBox from '../util/pointInsideBoundingBox.js';
+import calculateSUV from '../util/calculateSUV.js';
+import dragObject from '../util/freehand/dragObject.js';
+import dropObject from '../util/freehand/dropObject.js';
+import insertOrDelete from '../util/freehand/insertOrDelete.js';
+import freeHandArea from '../util/freehand/freeHandArea.js';
+import calculateFreehandStatistics from '../util/freehand/calculateFreehandStatistics.js';
+import freeHandIntersect from '../util/freehand/freeHandIntersect.js';
+import { FreehandLineFinder } from '../util/freehand/FreehandLineFinder.js';
```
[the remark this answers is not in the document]
JamesAPetts: Sure thing.

src/fancy-tools/freehandTool.js
```diff
@@ -0,0 +1,1143 @@
...
+// Drawing
+import { getNewContext, draw, drawJoinedLines } from './../util/drawing.js';
+import drawLinkedTextBox from './../util/drawLinkedTextBox.js';
+import { clipToBox } from '../util/clip.js';
+
+export default class extends baseAnnotationTool {
+
+  constructor (name) {
+    super({
+      name: name || 'freehand',
```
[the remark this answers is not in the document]
JamesAPetts: Sounds good.

src/fancy-tools/freehandTool.js
JamesAPetts: I see the two relevant PRs are merged, I'll wire this up to all three.

src/fancy-tools/freehandTool.js
JamesAPetts: Aha, that's nifty, I was just being cautious whilst implementing this, I'll remove it.

src/fancy-tools/freehandTool.js
JamesAPetts: Done an pushed.

JamesAPetts:
Hey @dannyrb, I know we already talked on slack, but airing it more publicly here. The main reason of adding and removing listeners is to lock into the event loop once we start using it. This is actually very similar to manipulators that other tools import, which do the same thing, only here I do this explicitly in the class.

There are also some things that could probably use shared functionality (e.g. the textBox), but a lot of the functionality is too bespoke to generalise. I'm happy for any ideas if anyone has them! :)

JamesAPetts:
P.S. All comments here are now addressed and pushed.

dannyrb:
@JamesAPetts just a heads up, I don't need anything else from you on this PR. I'm tidying up the dispatchers and working in some changes to support the flexibility your tool needs. After I'm at a good point, I'll address merge conflicts and pull these changes in.

Didn't want to leave you hanging (: Thanks again for all of your hard work 👍

src/eventDispatchers/mouseToolEventDispatcher.js
```diff
@@ -303,14 +338,36 @@ function mouseDownActivate (evt) {
...
+  let tools;
+  const eventData = evt.detail;
+  const element = eventData.element;
+
+  // Filter out disabled, enabled, and passive
+  tools = getActiveToolsForElement(element, getters.mouseTools());
+  tools = tools.filter((tool) =>
+    isMouseButtonEnabled(eventData.which, tool.options.mouseButtonMask)
+  );
+  tools = tools.filter((tool) => typeof tool.mouseDoublClickCallback === 'function');
```
swederik: mouseDoublClickCallback - missing an 'e'?

dannyrb: Hah. Good callout

JamesAPetts: Oops, this bit isn't used yet. You saved the annotation tool @swederik ! Will change tomorrow.

dannyrb:
Alright, @JamesAPetts. I've cleaned up everything that was blocking me from taking a harder look at this. I think our biggest hurdle for a merge is going to be the best way to handle the `mouseToolEventDispatcher.js` changes. The ones I see are:

- 👍 Adding a callback to `mouseMove`
- 🤷‍♂️ Modify `mouseDown` to identify target/active tool in a new way:
    - Tools with or without data
    - Tools with a `mouseDownCallback`
    - Tools w/ an alternative way to indicate they're a `validTarget` other than `pointNearTool`
- 🤷‍♂️ Modify `mouseDown` to call a tool's `mouseDownCallback` instead of default logic
- 🤖  Add implementation for `mouseDoubleClick`
- 🤖  `setIsAwaitingMouseUp`


🤷‍♂️ - The problem/solution we need to focus on
👍 - I agree, and we should keep this implementation
🤖 - I agree, but have provided an alternative implementation in a different PR. Curious to hear your thoughts.

I'll create a separate PR to add the 👍, and also to address 🤷‍♂️ . We can discuss 🤖 further if you'd like. After we agree on these details, we can merge this PR, but blow out the dispatcher changes.

JamesAPetts:
@dannyrb , thanks for taking the time to go through the mouseToolEventDispatcher changes thoroughly. I look forward to seeing your PR. As for your 🤖 comments, do you mean these will be present in your new PR, or there are already alternatives out in the wild I can look at?

JamesAPetts:
Closing as will rebuild in updated eventDispatch scheme.

PR #497, opened 2018-07-27
Chore/dry up single callback handlers for active tools

dannyrb:


dannyrb:
Having an issue where we `trigger` a touchPress event with a `null` detail (which means we can't find `element`)

```
if (evt.detail === null) {
    console.warn(`event detail is null for ${customFunction}:`, evt);

    return false;
  }
```

dannyrb:
Annnd another issue where if we allow the handle to finish animating, it jerks back to it's original position, and we're unable to continue dragging it.

These are mostly cliff notes for me. Feel free to ignore

dannyrb:
> Annnd another issue where if we allow the handle to finish animating, it jerks back to it's original position, and we're unable to continue dragging it.

Verified that this issue exists on `v2`, so we'll create an issue and consider this "merge safe".

PR #489, opened 2018-07-25
WIP: Migrating freehand tool to v3 (Part 2) 

JamesAPetts:
Please review this after #488.

- Ports the freehandSculpter to v3.
- Minor:
-- Added basic support for onNewImage to the mouseToolEventDispatcher. (If we'd rather the tool does this itself, that'd easy, but thought I'd centralize what I can).
-- Renamed the 'abstract' method onImageRendered -> renderToolData in baseAnnotationTool, as this is what the tools actually use.

PR #499, opened 2018-07-27
WIP V3 brush + KeyboardController

JamesAPetts:
This should be processed after #488 & #489.

Features:
- Fixed @swederik's brush tool fork of master to work correctly on Firefox, Edge and Chrome. Ported to v3 and added support for multiple colors, alpha settings and keyboard interface.

- This PR also contains a Keyboard controller that can be attached to any mouse tool. With support for remapping of keys base on user preferences. An example of this functionality can be found within the brushTool and brushEraserTool.

For Discussion:
- As the paint tools all share the same data, currently the brushTool deals with the rendering, and the brushTool must be at least enabled in order to use other brushes (e.g. the brushEraserTool, and the not-yet-ported adaptiveBrushTool). We may want to rejig where the rendering is done.

Known issues:
- The brushEraseTool works, but cannot render the cursor on mouse move currently, due to limitations of the WIP mouseToolEventDispatcher.

- painting on the far right side creates a line of the left side of the image. This only affects the UI and not the tool data. If anyone has a solution to this I would be very happy!

V3 Update: V3 all functionality of cornerstoneTools master. 

v3 Demo url: https://deploy-preview-499--cornerstone-tools.netlify.com/examples/00-test


JamesAPetts:
Closing as will rebuild in updated eventDispatch scheme.

PR #500, opened 2018-07-27
WIP: WWWCRegion: migration to v3

galelis:


dannyrb:
Test URL: https://deploy-preview-500--cornerstone-tools.netlify.com/examples/00-test/

src/fancy-tools/wwwcRegionTool.js
```diff
@@ -0,0 +1,209 @@
...
+  viewport.voi.windowCenter = minMaxMean.mean;
+
+  external.cornerstone.setViewport(element, viewport);
+  external.cornerstone.updateImage(element);
+};
+
+export default class extends baseTool {
+  constructor (name) {
+    const strategies = {
+      applyWWWCRegion
```
dannyrb: You probably don't need to use `strategies` if there is only one strategy. There's no real harm, but it does add unnecessary complication

PR #501, opened 2018-07-28
added MEASUREMENT_ADDING_END event

gillib:
added trigger for MEASUREMENT_ADDING_END event for moveNewHandle (mouse) moveEndCallback function

feature

When an roi (using a cornerstone tool) is firstly added but not set yet, the MEASUREMENT_ADDED event is triggered.
There is no event triggered when the adding is finished.
example - 
When adding a rectangleRoi, the first mouse click will set one of the rectangle's corners and trigger the MEASUREMENT_ADDED.
There is no event triggered when the second corner is set.

When the adding is done (for now only for mouse) a new event will be triggered 
MEASUREMENT_ADDING_END

There is no breaking change!

* **Other information**:

I think the naming of the original event (MEASUREMENT_ADDED) is wrong because with ROIs it is triggered before it was actually added but when the adding has started.
This event should be used in my case.
Anyway, because I didn't want to create a breaking change of the api so I named the new event a bit different but I believe this should be fine.

### DIFF
--- a/src/events.js
+++ b/src/events.js
@@ -37,6 +37,7 @@ const EVENTS = {
 
   // Measurement / tool events
   MEASUREMENT_ADDED: 'cornerstonetoolsmeasurementadded',
+  MEASUREMENT_ADDING_END: 'cornerstonetoolsmeasurementaddingend',
   MEASUREMENT_MODIFIED: 'cornerstonetoolsmeasurementmodified',
   MEASUREMENT_REMOVED: 'cornerstonemeasurementremoved',
   TOOL_DEACTIVATED: 'cornerstonetoolstooldeactivated',
--- a/src/manipulators/moveNewHandle.js
+++ b/src/manipulators/moveNewHandle.js
@@ -47,7 +47,9 @@ export default function (mouseEventData, toolType, data, handle, doneMovingCallb
     const eventData = e.detail;
 
     if (eventData.measurementData === data) {
-      moveEndCallback();
+      const hasBeenRemoved = true;
+
+      moveEndCallback(hasBeenRemoved);
     }
   }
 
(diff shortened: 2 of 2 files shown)


gillib:
can anyone form the cornerstone team respond to this please?
it has been a few weeks now

dannyrb:
@gillib I understand the problem you're trying to solve, but I'm not sure that this is the best solution. Triggering `MEASUREMENT_MODIFIED` would make more sense in this context, I believe. You may even be able to make the argument that `MEASUREMENT_ADDED` should not be called until `MOUSE_UP` or `TOUCH_DRAG_END` (or some similar "end" event)

For multi-step tools, it might make sense to have events for `started`/`completed` as the data would be in an "incomplete" state, and it might not make sense to attempt to persist it until it's been fully placed.

Feel free to weigh in @swederik? I don't have a strong feeling either way, but I'd appreciate consistency across the tools.

gillib:
Hey thanks for replying!
Your reply makes me understand that the issue you present is definetly arguable/solvable because it seems you understand my persistence problem 🙂

Regarding consistency - full consistency will not be achieved between multi step tools and single step tools especially if we won't want to break backwards compatibility. That's obvious. 

I think the best architecture would have been to have 2 events called measurement_adding_started and measurement_adding_completed for both single and multi step tools alike. 
For example:
-For single step- on mouse down would trigger the first event and on mouse up will trigger the second. 
-For multi step tools the FIRST mouse down would trigger the the first event and the LAST mouse up would trigger the second. 

If I understand the code correctly, the location I triggered the event is where the handling of any tool is ended. (If I understand wrong please let me know)
Therefore the only difference of my code with my suggested architecture is the naming of the events. 

So either I can rename both events and not be backwards compatible. 
Or I can just rename the second event to be a better name. 
Or I can add another event to the start and have 2 events triggering from the same location, and we can remove the old one in a later major release. 

What do you think?

Thanks!!

dannyrb:
@gillib, we're far enough along in the vNext branch that I think it would be a better place to address this issue (we can make breaking changes there). We should take inventory of when/where we are triggering measurement events, identify any inconsistencies or missed cases, and then update appropriately.

Realistically, I don't have the bandwidth to address this for v2; although another maintainer may.

dannyrb:

FYI, there is a related PR/Issue for this now open and being driven by @Zaid-Safadi

PR #504, opened 2018-07-31
[WIP] Feature:freehand dispatcher changes

dannyrb:
I'll tag you as a reviewer when I wrap up @JamesAPetts 

- [Specialty Tools](https://github.com/dannyrb/cornerstoneTools/blob/4e24873ed15e330a220ed5e206cb1e1c4d021aa9/src/eventDispatchers/mouseEventHandlers/mouseDown.js#L43-L59) take priority over other tools; they can choose to `preventPropagation` if they'd like
- For AnnotationTools, [we first check handles](https://github.com/dannyrb/cornerstoneTools/blob/4e24873ed15e330a220ed5e206cb1e1c4d021aa9/src/eventDispatchers/mouseEventHandlers/mouseDown.js#L63-L88):
    - Which tools have handles near our point?
    - Which tool should we use? (For now, first one in array)
    - Use `mouseDownCallback` if it's defined, or `handleMover`
- Then [we check PointNearTool](https://github.com/dannyrb/cornerstoneTools/blob/4e24873ed15e330a220ed5e206cb1e1c4d021aa9/src/eventDispatchers/mouseEventHandlers/mouseDown.js#L91-L127):
    - Which tools have are near our point?
    - Which tool should we use? (For now, first one in array)
    - Use `mouseDownCallback` if it's defined, or `moveAllHandles` handle mover

### DIFF
--- a/src/eventDispatchers/mouseEventHandlers/mouseDown.js
+++ b/src/eventDispatchers/mouseEventHandlers/mouseDown.js
@@ -11,13 +11,14 @@ import isMouseButtonEnabled from './../../util/isMouseButtonEnabled.js';
 import getInteractiveToolsForElement from './../../store/getInteractiveToolsForElement.js';
 import getToolsWithDataForElement from './../../store/getToolsWithDataForElement.js';
 
-// Note: if we find a match, we need to record that we're holding down on a tool
-// So we don't fire the mouse_move event listener
-// On pick-up, we need to "release", so we can re-enable the mouse_move listener
 /**
  * MouseDown is called before MouseDownActivate. If MouseDown
  * finds an existing tool to interact with, it can prevent the
  * event from bubbling to MouseDownActivate.
+ *
+ * TODO: Set that a tool is active to prevent multiple event fires
+ * TODO: Handles should trigger image update when released
+ * TODO: Handles should handle delting out of bound data by setting
  *
  * @param {*} evt
  * @returns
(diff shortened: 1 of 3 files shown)


dannyrb:
Misc. thoughts to munch on:

- We drop `isValidTarget` and let the tool handle that internally. The tool can exit early and the `eventDispatcher` will continue to check other tools.
- We remove the `mouseDownCallback` from the `handle` and `pointNearTool` calls. Annotation tools probably shouldn't deviate, or if they do it should be with a different callback name, or an additional prop passed to the the callback?

In general `mouseDown` is used to move or manipulate existing tool data, and `mouseDownActivate` is used if no existing data can be manipulated to create new data. Does `freehand` follow that convention? If not, why? (Is it mostly due to limitations of a multi-step tool? Can we lock `freehand` into its own loop in `mouseDownActivate` instead?)

JamesAPetts:
@dannyrb, going to review this now so you can wake up to lots of feedback!

Firstly, in regards to your comment:

> Does freehand follow that convention? If not, why? (Is it mostly due to limitations of a multi-step tool? Can we lock freehand into its own loop in mouseDownActivate instead?)

Freehand does work in the same way. Short answer: mouseDownActivate creates freehand data, and mouseDown is used to manipulate existing data. Longer answer: The only issues that arose with the freehand tool are that to manipulate existing data, the user can insert a point in the tool by ctrl+clicking near a line. The tool has its own internal logic to check whether the projected distance from the line is within a certain threshold, yet no handle is actually clicked on. This functionality should easily fit in the proposed framework though.

src/eventDispatchers/mouseEventHandlers/mouseDown.js
```diff
@@ -29,122 +30,193 @@ export default function (evt) {
...
   tools = tools.filter((tool) =>
     isMouseButtonEnabled(eventData.which, tool.options.mouseButtonMask)
   );
-  tools = getToolsWithDataForElement(element, tools);
 
-  // Find tools with handles we can move
-  const toolsWithMoveableHandles = tools.filter((tool) => {
-    const toolState = getToolState(eventData.element, tool.name);
+  // Check if any tool has a special reason to grab the current event
+  const specialTools = tools.filter(
```
JamesAPetts: I'm happy with the way this is done, and understand this is cleaning up what I did, however, upon further reflection, would it not be better to have either:
- "special tools" extend a subclass of specialTool, in which these methods are mandatory, so we don't need to constantly check if all these conditions are true.
- A simple tag, isSpecialTool (or some better name), that implies these methods are present.

I'd prefer the former as its more vigorous, but I'm not sure if it adds another layer of complexity for DX. I think we could convey it quite easily though with appropriate naming and warning functions.

src/eventDispatchers/mouseEventHandlers/mouseDown.js
```diff
@@ -29,122 +30,193 @@ export default function (evt) {
...
+    (tool) =>
+      typeof tool.mouseDownCallback === 'function' &&
+      typeof tool.isValidTarget === 'function' &&
+      tool.mode === 'active' &&
```
JamesAPetts: I think restricting this to active tools makes sense, more of a question than a comment: is there any tool which requires mouseDown functionality whilst passive and not hovering over a handle? I don't think there is, so feel this is correct.

JamesAPetts: Edit: quick mockup if we went with my first solution:

```
const specialTools = tools.filter(
  (tool) => tool instanceof 'SpecialTool' &&
  tool.isValidTarget(evt)
);
```

src/eventDispatchers/mouseEventHandlers/mouseDown.js
```diff
@@ -29,122 +30,193 @@ export default function (evt) {
...
+
+  // HANDLES
+  if (annotationToolsWithMoveableHandles.length > 0) {
+    const firstToolWithMoveableHandles = annotationToolsWithMoveableHandles[0];
+    const toolState = getToolState(element, firstToolWithMoveableHandles.name);
+
+    const hasCustomMouseDownCallback =
+      firstToolWithMoveableHandles.mouseDownCallback === 'function';
+
+    const throwAway = hasCustomMouseDownCallback
```
JamesAPetts: Why do you define a variable here? can't this just be:

```
hasCustomMouseDownCallback
      ? firstToolWithMoveableHandles.mouseDownCallback(evt)
      : findAndMoveHandleNearImagePoint(
        element,
        evt,
        toolState,
        firstToolWithMoveableHandles.name,
        coords
);
```

src/eventDispatchers/mouseEventHandlers/mouseDown.js
```diff
@@ -29,122 +30,193 @@ export default function (evt) {
...
+    const toolState = getToolState(element, firstToolWithPointNearClick.name);
 
-    for (let i = 0; i < toolState.data.length; i++) {
-      const data = toolState.data[i];
-      const distance = 6;
-      const handle = getHandleNearImagePoint(
+    const hasCustomMouseDownCallback =
+      firstToolWithPointNearClick.mouseDownCallback === 'function';
+
+    const throwAway = hasCustomMouseDownCallback
```
JamesAPetts: Same point as previous comment, throwAway needed?

src/eventDispatchers/mouseEventHandlers/mouseDown.js
```diff
@@ -29,122 +30,193 @@ export default function (evt) {
...
+        toolState,
+        firstToolWithMoveableHandles.name,
+        coords
+      );
+
+    return;
+  }
+
+  // POINT NEAR
+  const annotationToolsWithPointNearClick = tools.filter((tool) => {
```
JamesAPetts: :+1:Like this, a lot cleaner.

src/eventDispatchers/mouseEventHandlers/mouseDown.js
```diff
@@ -29,122 +30,193 @@ export default function (evt) {
...
+      ) {
+        return true;
       }
     }
+
+    return false;
   });
+};
+
+const findAndMoveHandleNearImagePoint = function (
```
JamesAPetts: :+1:

src/eventDispatchers/mouseEventHandlers/mouseDown.js
```diff
@@ -29,122 +30,193 @@ export default function (evt) {
...
+      evt.stopImmediatePropagation();
+      evt.stopPropagation();
+      evt.preventDefault();
+
+      return;
+    }
+  }
+};
+
+const findAndMoveAnnotationNearClick = function (
```
JamesAPetts: :+1:

src/eventDispatchers/mouseEventHandlers/mouseMove.js
```diff
@@ -57,6 +57,12 @@ export default function (evt) {
         data.active = !data.active;
         imageNeedsUpdate = true;
       }
+
+      // TODO: We may want to do _only_ this, and provide a default implementation
```
JamesAPetts: :+1::+1::+1:

I think this makes the most sense. All of this can go into BaseAnnotationTool's virtual mouseMoveCallback, cleans up the readability of this routine too. I think more dispatching of events rather than handling of events here is good.

JamesAPetts: Also a note for the future, we still need to think about what to do for cursors here. I think its the main (only?) use case for mouseMoveCallback.

Between lines 28 and 29, we could filter all tools with an updateCursorPosition(evt) function and call them. This function would most likely just be something along the lines of:

```
  updateCursorPosition (evt) {
    const eventData = evt.detail;


    this._lastImageCoords = eventData.currentPoints.image;
  }
```
Such that the position is known by the class for rerender.

JamesAPetts: Actually! (left my previous comment to demonstrate my thought process).

On mouseMove can't we just put the mousePosition into the mouseEventDispatcher as an export?

```
let mousePosition;
export { mousePosition };
```
In mouseEventHandler.mouseMove:
```
mousePosition = evt.detail.currentPoints.image; // or currentPoints.canvas?
```


Job done, any class that wants a custom cursor can pull this info in at render time. We can now use mouseEventDispatcher.mouseMove to track only hover/unhover.

src/eventDispatchers/mouseEventHandlers/mouseDown.js
JamesAPetts: Edit: quick mockup if we went with my first solution:

```
const specialTools = tools.filter(
  (tool) => tool instanceof 'SpecialTool' &&
  tool.isValidTarget(evt)
);
```

PR #508, opened 2018-07-31
WIP: Scale overlay tool

galelis:


galelis:
TODO's: 
- Still missing unit tests. 
- Confirm with Danny if onDisabled and onEnabled callbacks should be create as I did

PR #509, opened 2018-07-31
WIP: ScaleOverlayTool: Basic code

galelis:


galelis:
TODO's:

Still missing unit tests.
Confirm with Danny if onDisabled and onEnabled callbacks should be create as I did

examples/00-scaleOverlay/index.html
```diff
@@ -0,0 +1,121 @@
+<!DOCTYPE HTML>
```
dannyrb: I'm a little torn. On one-hand simple, granular examples are easier to learn from to use a tool. On the other hand, a lot of simple examples can become hard to maintain. For the time being, until we get things hammered out, I think we should slowly add on to the "global" example, and delete the ones that are no longer applicable.

When this becomes the official branch, we can double back and rethink how best to create and maintain examples as a form of documentation.

dannyrb: If you'd like, we can merge with this, and I can use it as a guide to update the primary example, and then delete it? I've also wanted to add a simple way to bind to different mouse buttons in the main example so you can test multiple tools in tandem.

src/fancy-tools/scaleOverlay.js
```diff
@@ -0,0 +1,242 @@
+import external from './../externalModules.js';
+import baseTool from './../base/baseTool.js';
+// Drawing
+import { getNewContext, draw, path } from '../util/drawing.js';
+
+const cornerstone = external.cornerstone;
+
+const drawLine = (context, startPoint, endPoint) => {
```
dannyrb: Can we use the `drawLine` from `util/drawing.js` instead

src/fancy-tools/scaleOverlay.js
```diff
@@ -0,0 +1,242 @@
...
+    });
+  }
+
+  enabledCallback (element) {
+    external.cornerstone.updateImage(element);
+  }
+
+  disabledCallback (element) {
+    external.cornerstone.updateImage(element);
+  }
```
dannyrb: `enabledCallback` and `disabledCallback` look good 👍 

each `setToolMode` method calls a callback with the following formula:
- `tool['${mode}Callback'](element, options);`

src/fancy-tools/scaleOverlay.js
```diff
@@ -0,0 +1,242 @@
...
+
+export default class extends baseTool {
+  constructor (name) {
+    super({
+      name: name || 'scaleOverlay',
+      configuration: {
+        color: 'white',
+        lineWidth: 2,
+        shadowColor: 'black',
+        shadowBlur: 4,
```
dannyrb: We might want to keep track of what common properties we may want to have a global setting/default for, and then allow a local value to override. That way it's easier to set values that affect all tools, and override them as needed.

dannyrb:
Example URL: https://deploy-preview-509--cornerstone-tools.netlify.com/examples/00-scaleoverlay/

src/fancy-tools/scaleOverlay.js
galelis: You mean I should be using the `stateManagement/` to get those values?

src/fancy-tools/scaleOverlay.js
galelis: yeah, I finished this first part, with the idea of just make it work. 
I will refactor :)

examples/00-scaleOverlay/index.html
galelis: hmm, make sense. 
I will figure out a way to add this tool into the global example and delete this example.

dannyrb:
image

PR #511, opened 2018-08-01
[WIP] Feature:freehand dispatcher changes - Round 2

JamesAPetts:
Extension of #504  by @dannyrb .

Added the possibility for 3 optional functions to override default mouseDown functionality:

- activeMouseDownCallback(evt) - Callback that takes priority if the tool is active, in case any special behavior is required.
- handleSelectedCallback(evt, handle) - Callback that overrides handle movement.
- toolSelectedCallback(evt, toolData) - Callback that overrides tool movement.

Also:
- Stores the mouse position on mouseMove, regardless of what happens, in case any tool's renderToolData method requires it.
- On mouseMove, if activeTool.hasCursor, then call a re-render. This should remove any need for  tools to have their own mouseMoveCallback.


src/base/baseAnnotationTool.js
```diff
@@ -15,25 +15,37 @@ export default class extends baseTool {
       configuration,
       supportedInteractionTypes
     });
-
-    this.isAnnotationTool = true;
```
dannyrb: Good call on removing this 👍 
It's one of those things I put in "just because", but never ended up using anywhere.

src/base/baseAnnotationTool.js
```diff
@@ -42,15 +54,48 @@ export default class extends baseTool {
...
+   * @param  {*} evt
+   * @param  {*} tool The selected tool.
+   * @return {boolean} stopImmediatePropagation - True if function consumed the event.
+   */
+  /*
+  toolSelectedCallback (evt, tool) {
+    // Implementation
+    return stopImmediatePropagation;
+  }
+  */
```
dannyrb: Really like the separation of `abstract` and `optional`. Really easy to follow 👍

src/eventDispatchers/mouseEventHandlers/mouseDown.js
```diff
@@ -29,31 +33,133 @@ export default function (evt) {
...
+    const firstActiveTool = activeTools[0];
+
+    if (typeof firstActiveTool.activeMouseDownCallback === 'function') {
+      const shouldPreventPropagation = firstActiveTool.activeMouseDownCallback(evt);
+
+      if (shouldPreventPropagation) {
+        preventPropagation(evt);
+
+        return;
+      }
```
dannyrb: I think we may need to separate these two bits of logic.

if `firstActiveTool.activeMouseDownCallback(evt);` returns true, then we should `return;` so the dispatcher doesn't dispatch the event multiple times; however, the callback itself is handed the event, so it can choose to stopPropagation if it would like to?

I don't want to lock users into stopping propagation if they want the eventDispatcher to exit.

(I hope that makes sense)

src/eventDispatchers/mouseEventHandlers/mouseDown.js
```diff
@@ -29,31 +33,133 @@ export default function (evt) {
...
+  const activeTools = tools.filter((tool) => tool.mode === 'active');
+
+  // If any tools are active, check if they have a special reason for dealing with the event.
+  if (activeTools.length > 0) {
+    // TODO: If length > 1, you could assess fitness and select the ideal tool
+    // TODO: But because we're locking this to 'active' tools, that should rarely be an issue
+    const firstActiveTool = activeTools[0];
+
+    if (typeof firstActiveTool.activeMouseDownCallback === 'function') {
+      const shouldPreventPropagation = firstActiveTool.activeMouseDownCallback(evt);
```
dannyrb: is `activeMouseDownCallback` to indicate it's only called for `active` tools?
That makes sense, but I initially wondered if it was for the `mouseDownActivate` event

src/eventDispatchers/mouseEventHandlers/mouseDownActivate.js
```diff
@@ -30,7 +31,7 @@ export default function (evt) {
   // Note: custom `addNewMeasurement` will need to prevent event bubbling
   if (activeTool.addNewMeasurement) {
     activeTool.addNewMeasurement(evt, 'mouse');
-  } else if (activeTool.isAnnotationTool) {
+  } else if (activeTool instanceof baseAnnotationTool) {
```
dannyrb: much cleaner 👍

src/eventDispatchers/mouseEventHandlers/mouseMove.js
```diff
@@ -22,10 +22,28 @@ export default function (evt) {
   let tools;
   const { element, currentPoints } = evt.detail;
 
+  // Set the mouse position incase any tool needs it.
+  setters.mousePositionImage = currentPoints.image;
```
dannyrb: I was winging the implementation for `getters`, `state`, `mutations`, but we can switch to `setters` if that makes more sense? Honestly, I was stealing some nomenclature from Vue/React

src/eventDispatchers/mouseEventHandlers/mouseDown.js
JamesAPetts: Hey, you're correct that that's what I was trying to dictate, but I didn't think of that, maybe the naming is a little confusing.

src/eventDispatchers/mouseEventHandlers/mouseMove.js
JamesAPetts: I'm easy on this one, I just added some setting function as I wasn't so sure this counted as a mutation (perhaps it does?), and needed somewhere to put it.

src/eventDispatchers/mouseEventHandlers/mouseMove.js
```diff
@@ -57,6 +75,12 @@ export default function (evt) {
         data.active = !data.active;
         imageNeedsUpdate = true;
       }
+
+      // TODO: We may want to do _only_ this, and provide a default implementation
+      // TODO: On each tool?
+      if (typeof tool.mouseMoveCallback === 'function') {
+        tool.mouseMoveCallback(evt);
```
dannyrb: Do we need an `active` version of this callback as well? One that tools without data can utilize? 🤔

JamesAPetts: Feel free to disagree, but I'm an advocate of not adding functionality that isn't currently needed. If a current tool needs this then sure. If its needed for a future tool we can make a new PR for it.

I'm even inclinded to remove these few lines, as I don't think we need it now? mouseMove is only used for hover activation, and custom cursors are now supported without the tool needing a callback.

src/eventDispatchers/mouseEventHandlers/mouseMove.js
dannyrb: Sounds good. We can merge this as is, and I can update it to match the current method?

JamesAPetts: If you check the store, I've already done this :).

src/eventDispatchers/mouseEventHandlers/mouseDown.js
dannyrb: I think this is the only "blocker" for a merge. Can we update this so that:

```
const claimEvent = firstActiveTool.activeMouseDownCallback(evt);
if(claimEvent){ return; }
``` 

And `activeMouseDownCallback(evt)` can call

`evt.stopImmediatePropagation();` and other actions on the event if it wants to?

dannyrb: We can talk about this more if you'd like; that's just the preference that gives me the good feels; but I can be swayed :P

JamesAPetts: Yeah, that seems fine to me, simple actually!

PR #514, opened 2018-08-02
vNext freehandTool

JamesAPetts:
Now the new event dispatching framework is pulled, I've refactored the freehandTool.

The textbox movement now uses the default textbox movement to remove duplication.
I've also refactored to make the private event loops more clear.

I've exported (rather brashly) findAndMoveHandleNearImagePoint from eventDispatchers/mouseEventHandles/mouseDown.js. Do we want to stick all these helper functions in another shared file? Other tools might want to use them like I have. Thought I'd get a second opinion rather than just do it.

Thanks,
James


src/fancy-tools/freehandMouseTool.js
```diff
@@ -0,0 +1,1080 @@
...
+// Drawing
+import { getNewContext, draw, drawJoinedLines } from './../util/drawing.js';
+import drawLinkedTextBox from './../util/drawLinkedTextBox.js';
+import { clipToBox } from '../util/clip.js';
+
+export default class extends baseAnnotationTool {
+
+  constructor (name) {
+    super({
+      name: name || 'freehandMouse',
```
swederik: You can give name a default value in the constructor arguments

name = 'freehandMouse'

and then super { name, ... }

src/fancy-tools/freehandMouseTool.js
```diff
@@ -0,0 +1,1080 @@
...
+   * @param {*} eventData
+   * @returns
+   */
+  createNewMeasurement (eventData) {
+
+    const goodEventData =
+      eventData && eventData.currentPoints && eventData.currentPoints.image;
+
+    if (!goodEventData) {
+      console.error(
```
swederik: Any reason why we shouldn't just throw an error?

src/fancy-tools/freehandMouseTool.js
```diff
@@ -0,0 +1,1080 @@
...
+
+    const measurementData = {
+      visible: true,
+      active: true,
+      invalidated: true,
+      color: undefined,
+      handles: []
+    };
+
+    measurementData.handles.textBox = {
```
swederik: Isn't 'handles' an array? Is it an array with a 'textBox' property? Seems a bit weird, no?

In other tools, handles is an object.

src/fancy-tools/freehandMouseTool.js
JamesAPetts: Nope, that would make more sense 👍 .

src/fancy-tools/freehandMouseTool.js
```diff
@@ -0,0 +1,1080 @@
...
+      );
+    }
+
+    if (!validParameters || data.visible === false) {
+      return false;
+    }
+
+    const isPointNearTool = this._pointNearHandle(element, data, coords);
+
+    if (isPointNearTool !== null) {
```
swederik: You can just do:

```
return isPointNearTool !== null;
```

or !!isPointNearTool, or return this._pointNearHandle...

src/fancy-tools/freehandMouseTool.js
JamesAPetts: Yup, that's a lot cleaner, thanks.

src/fancy-tools/freehandMouseTool.js
```diff
@@ -0,0 +1,1080 @@
...
+   * @param {*} coords
+   * @returns
+   */
+  pointNearTool (element, data, coords) {
+    const validParameters =
+      data && data.handles;
+
+    if (!validParameters) {
+      console.warn(
+        `invalid parameters supplieed to tool ${this.name}'s pointNearTool`
```
swederik: typo in supplied

Again, I think it seems more sensible to just throw an error.

src/fancy-tools/freehandMouseTool.js
```diff
@@ -0,0 +1,1080 @@
...
+    return measurementData;
+  }
+
+  /**
+   *
+   *
+   * @param {*} element
+   * @param {*} data
+   * @param {*} coords
+   * @returns
```
swederik: returns a boolean

src/fancy-tools/freehandMouseTool.js
JamesAPetts: Handles is an array with a textBox property, I had to do this so it mouseEventDispatcher would enumerate properly. In most circumstances internally I want to cycle through an array, but I need to handle mouse clicks on the textBox from the mouseEventDispatcher.

I'm open for a neater solution if you can think of one.

src/fancy-tools/freehandMouseTool.js
```diff
@@ -0,0 +1,1080 @@
...
+    }
+
+    // Add the new handle
+    data.handles.push(newHandleData);
+
+    // Increment the current handle value
+    config.currentHandle += 1;
+
+    // Force onImageRendered to fire
+    external.cornerstone.updateImage(eventData.element);
```
swederik: Is this necessary? Isn't this caused by mouse click, which would already force it to update?

JamesAPetts: It's within the internal drawing loop, which blocks the main eventDispatcher, so the rendering needs to be dealt with manually. Perhaps it makes sense to move that the render to the end of _drawingMouseDownCallback though.

src/fancy-tools/freehandMouseTool.js
```diff
@@ -0,0 +1,1080 @@
...
+  *
+  * @private
+  * @param {Object} eventData - Data object associated with an event.
+  * @param {Object} dataHandles - Data object associated with the tool.
+  */
+  _addPointPencilMode (eventData, dataHandles) {
+    const config = this.configuration;
+    const mousePoint = config.mouseLocation.handles.start;
+
+    for (let i = 0; i < dataHandles.length; i++) {
```
swederik: Might be more clear with Array.some or Array.every? Took me a minute to understand what is supposed to be happening here. You want to ensure that the mouse is farther than 'spacing' away from all points?

if (Array.every(handleFurtherAwayThan(config.spacing)) { this._addPoint }

or 

if (Array.some(handleCloserThanMinimumSpacing)) { return; }

src/fancy-tools/freehandMouseTool.js
```diff
@@ -0,0 +1,1080 @@
...
+    }
+
+    // Check to see if mouse in bounding box of textbox
+    if (data.handles.textBox) {
+      if (pointInsideBoundingBox(data.handles.textBox, coords)) {
+        return data.handles.textBox;
+      }
+    }
+
+    return null;
```
swederik: Necessary? Is returning undefined not good enough?

src/fancy-tools/freehandMouseTool.js
```diff
@@ -0,0 +1,1080 @@
...
+  /**
+  * Adds a point on mouse click in polygon mode.
+  *
+  * @private
+  * @param {Object} eventData - data object associated with an event.
+  */
+  _addPoint (eventData) {
+    const toolState = getToolState(eventData.element, this.name);
+
+    if (toolState === undefined) {
```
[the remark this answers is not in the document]
JamesAPetts: Stuck with !toolState, and removed a load of checks where the toolState is now guaranteed to exist due to being inside a drawing loop.

src/fancy-tools/freehandMouseTool.js
```diff
@@ -0,0 +1,1080 @@
...
+  }
+
+  /**
+  * Removes modify loop event listeners.
+  *
+  * @private
+  * @param {Object} element - The viewport element to add event listeners to.
+  * @modifies {element}
+  */
+  _deactivateModifty (element) {
```
swederik: Modifty? I guess this is meant to be Modify?

src/fancy-tools/freehandMouseTool.js
```diff
@@ -0,0 +1,1080 @@
...
+  }
+
+  /**
+  * Adds modify loop event listeners.
+  *
+  * @private
+  * @param {Object} element - The viewport element to add event listeners to.
+  * @modifies {element}
+  */
+  _activateModifty (element) {
```
swederik: Same here.

dannyrb: Consistency 😉

src/fancy-tools/freehandMouseTool.js
JamesAPetts: It actually makes for the call to happen in addPoint, as it needs to occur when using the pencil mode and a node is autodropped also.

PR #517, opened 2018-08-03
V3 - Brush

JamesAPetts:
- Added brush back to new framework.

Edit: updated to the #521 api changes.


JamesAPetts:
Fixed most of tests that I broke by adding the KEY_DOWN event to the mouseToolEventDispatcher. Confused as to why the brushTool is interfering with the other tool's tests?

package.json
```diff
@@ -85,6 +85,7 @@
     "webpack-dev-server": "^3.1.4"
   },
   "dependencies": {
+    "babel-plugin-transform-runtime": "^6.23.0",
```
dannyrb: I think I know why, but just curious, why this and the linter change?

src/base/baseBrushTool.js
```diff
@@ -0,0 +1,367 @@
...
+  }) {
+    super({
+      name,
+      strategies,
+      defaultStrategy,
+      configuration,
+      supportedInteractionTypes
+    });
+
+    this.isBrushTool = true;
```
dannyrb: similar to `isAnnotationTool`, we can probably drop this and use `instanceof`?

src/base/baseBrushTool.js
```diff
@@ -0,0 +1,367 @@
...
+    this.hasCursor = true;
```
dannyrb: Do we want _all_ brush tools to have a cursor (and in all states for that brush)? Is there a way to change this? Would it be worth separating out the `render` so that tools with a `renderCursor` method trigger the redraw and handle the rendering of the cursor so we can keep the logic separate from the primary tool rendering?

src/eventDispatchers/mouseToolEventDispatcher.js
```diff
@@ -30,6 +31,7 @@ const enable = function (element) {
   element.addEventListener(EVENTS.MOUSE_MOVE, mouseMove);
   element.addEventListener(EVENTS.MOUSE_UP, mouseUp);
   element.addEventListener(EVENTS.MOUSE_WHEEL, mouseWheel);
+  element.addEventListener(EVENTS.KEY_DOWN, keyDown);
```
dannyrb: We have a lot of `misc` listeners like this popping up (IE. some of the `cornerstone` specific ones that aren't tied to touch/mouse). I'm still noodling on the best place for them to live.

src/base/baseBrushTool.js
```diff
@@ -0,0 +1,367 @@
...
+  // Abstract Methods - Must be implemented.
+  //===================================================================
+
+  /**
+  * Helper function for rendering the brush.
+  *
+  * @abstract
+  * @param {Object} evt - The event.
+  */
+  renderBrush (evt) {
```
dannyrb: I know very little about the brush tool. Is there a reason to provide this method AND the `renderToolData` method? Is it a way to separate brush rendering logic from annotations the tool may also create (IE. textboxes holding calculations related to what's been painted)?

src/base/baseBrushTool.js
```diff
@@ -0,0 +1,367 @@
...
+
+  /**
+   * Callback for when the tool is activated.
+   *
+   * @virtual
+   */
+  activeCallback () {
+    const configuration = this.configuration;
+
+    this._changeDrawColor(configuration.draw);
```
dannyrb: Can this be moved to the constructor? And then if you modify the configuration, or expose a method to change this property, it's applied when drawing?

src/base/baseBrushTool.js
```diff
@@ -0,0 +1,367 @@
...
+  }
+
+  /**
+  * Event handler for MOUSE_DOWN event.
+  *
+  * @virtual
+  * @event
+  * @param {Object} evt - The event.
+  */
+  postMouseDownCallback (evt) {
```
dannyrb: Interesting choice. I wondered if you'd favor `brush` or `annotations` when trying to paint near/over an existing annotation.

src/base/baseBrushTool.js
```diff
@@ -0,0 +1,367 @@
...
+  }
+  /**
+  * Adds modify loop event listeners.
+  *
+  * @protected
+  * @param {Object} element - The viewport element to add event listeners to.
+  * @modifies {element}
+  */
+  _startListeningForMouseUp (element) {
+    element.removeEventListener(EVENTS.MOUSE_UP, this._drawingMouseUpCallback);
```
dannyrb: Makes sense. We're locking ourselves to this tool while drawing, and this works like the handles in that an up/click will end the loop/lock?

src/base/baseBrushTool.js
```diff
@@ -0,0 +1,367 @@
...
+  /**
+   * Sets the keybinds for the tool. // TODO: If this will be used in other
+   * tools, maybe it should be stored in a different module and imported?
+   *
+   */
+  set keyBinds (keyBinds) {
+    const configuration = this.configuration;
+
+    configuration.keyBinds = keyBinds;
+    this._keyboardController = new KeyboardController(this, keyBinds);
```
dannyrb: Oooh. Interesting. I'll have to mull this over.

src/eventDispatchers/mouseEventHandlers/keyDown.js
```diff
@@ -0,0 +1,34 @@
+import external from './../../externalModules.js';
+// State
+import { getters, state } from './../../store/index.js';
+// Todo: Where should these live?
+import getInteractiveToolsForElement from './../../store/getInteractiveToolsForElement.js';
+
+const cornerstone = external.cornerstone;
+
+/**
+ * KeyDown is used for keyboard input for mouse tools.
```
dannyrb: You might be able to delete this and just create a new function by calling `.bind()` on `customCallbackHandler.js`

src/fancy-tools/brushEraserTool.js
```diff
@@ -0,0 +1,144 @@
...
+  constructor (name = 'brushEraser') {
+    super({
+      name,
+      supportedInteractionTypes: ['mouse'],
+      configuration: defaultBrushConfiguration()
+    });
+
+    const keyBinds = this.configuration.keyBinds;
+
+    this._keyboardController = new KeyboardController(this, keyBinds);
```
dannyrb: If we go this route, we can probably have the `super`/`base` take care of instantiating the KeyboardController if keyBinds are specified?

src/fancy-tools/brushTool.js
```diff
@@ -0,0 +1,295 @@
...
+     });
+   };
+ }
+
+export default class extends baseBrushTool {
+  constructor (name = 'brush') {
+    super({
+      name,
+      supportedInteractionTypes: ['mouse'],
+      configuration: defaultBrushToolConfiguration()
```
dannyrb: If it's a default, we can have it set in the `baseBrushTool` and use `Object.assign({}, defaultBrushToolConfiguration, customBrushToolConfiguration)` so long as we keep the object flat?

src/fancy-tools/brushTool.js
```diff
@@ -0,0 +1,295 @@
...
+   *
+   * @override
+   * @param {Object} element - The element to add/remove event listners.
+   */
+  activeCallback(element) {
+    const config = this.configuration;
+    this._changeDrawColor(config.draw);
+
+    element.removeEventListener(EVENTS.NEW_IMAGE, this._onNewImageCallback);
+    element.addEventListener(EVENTS.NEW_IMAGE, this._onNewImageCallback);
```
dannyrb: We check for a `newImageCallback` on each tool now. Use that instead, and then check your tool's `mode` property to see if you should execute the logic in that hook?

src/fancy-tools/shared/isToolActive.js
```diff
@@ -0,0 +1,7 @@
+import getTool from '../../store/getTool.js';
```
dannyrb: Not 100% sure where want this to live; it would make sense in the `getters` for the `store`?

dannyrb:
Done with a first pass at reviewing. I'm still 🤔 over the keyboard changes. I'm sure @swederik will have better `brush` specific feedback.

package.json
JamesAPetts: This is so we can transpile native ES6 promises and async functions to ES5. The window.createImageBitmap returns a promise, so I added these babel plugins such that I could write a polyfill for brush for IE/Edge/Safari.

src/base/baseBrushTool.js
JamesAPetts: Good catch, this was pretty copy-paste before I decided on that. Its not used anywhere, so I'll just remove it.

src/base/baseBrushTool.js
JamesAPetts: I can't see a brush tool that wouldn't require some form of cursor so that you can see what you intend to paint, but there's always an exception, which I'd be happy to hear about if you have any idea.

As for the latter part of your question, I do this with renderBrush, but you're right it should be renamed renderCursor for consistency.

src/eventDispatchers/mouseToolEventDispatcher.js
JamesAPetts: I agree there must be a neater way to store the keyboard events. Both mouseTools  and keyboardTools can utilize them, so they are a bit in limbo. The solution will likely become clearer when we think about putting keyboard only tools back in v3, are there any on our current roadmap?

src/base/baseBrushTool.js
JamesAPetts: I agree should be refactored. My original rationale for doing it this way: trying to fit a rather new methodology into the current framework.

I only wanted one brush tool to do all the rendering of the actual mask. There is one common data file per slice that "all" brushes manipulate (currently only 2 tools, but we envisage at least 2 more). Currently the brushTool overrides renderToolData to draw the actual mask, then calls renderBrush. The eraser uses the default renderToolData which just wraps renderBrush.

The downside of this approach is that currently the brushTool needs to be enabled for the brushEraserTool to function, which isn't too nonsensical for now. However, once we have an adaptive brush, perhaps a user would like to enable e.g. only adaptiveBrushTool and brushEraserTool.

I think the solution for this is as follows:
- renderToolData just renders the cursor (leave the wrapper for easy to follow naming? idk).
- in eventDispatchers/onImageRendered:
  - filters tools by (instanceof baseBrushTool).
  - if # baseBrushTools > 0, render brush data.

I'm happy to work on this refactor, but will wait for an opinion on this idea first.

src/base/baseBrushTool.js
JamesAPetts: Yeah I'm still unsure of this.The two options are:
- Completely ignore the passive status of all other tools.
- Make painting potentially more fiddly on a densely annotated image.

This is something we should probably discuss on Thursday.

src/base/baseBrushTool.js
JamesAPetts: Yeah I think that'd make more sense, done.

The _changeDrawColor method is already wrapped and exposed by the keyboard interface. Do you intend to have a touch interface for the painting tools? We may need to expose it in another way if you wish to switch stuff via a menu on a phone. I feel like the UX of a brushTool interface may be tricky to get right on touch devices though.

src/base/baseBrushTool.js
JamesAPetts: Yup :).

src/eventDispatchers/mouseEventHandlers/keyDown.js
JamesAPetts: This is a good idea. I can do this if we don't allow multiple functions to be bound to the same key, otherwise we need the bespoke routine. This is something to discuss.

src/fancy-tools/brushEraserTool.js
JamesAPetts: That's a good idea. I tried to be less intrusive to core whilst these were the only tools using hotkeys. I guess I could implement this now.

JamesAPetts: That's a good idea. I tried to be less intrusive to core whilst these were the only tools using hotkeys. I guess I could implement this now.

src/fancy-tools/brushTool.js
JamesAPetts: Its the default for the brushTool, the brushEraserTool has distinct defaults. Also the keybinds are in the configuration, which i'd rather not flatten, maybe we need to discuss the best approach for this.

src/fancy-tools/brushTool.js
JamesAPetts: Aha, didn't realise a dispatcher for this method was in now. I'll fix this for the brushTool for now, but as I mentioned elsewhere, perhaps this rendering functionality needs to be centralized somewhere away from a specific brush tool.

src/fancy-tools/shared/isToolActive.js
JamesAPetts: Yeah, I think getters.tool(toolName) would make more sense now.

src/fancy-tools/brushTool.js
JamesAPetts: EDIT: I need to use newImageCallback even when the tool is passive/enabled. This is because the tool uses this hook in order to redraw the element to use the new slices toolData. I'm not sure if there is an easier way around this now, this is how we had to do it in the v2 implementation.

JamesAPetts: I won't change anything about this comment until we have discussed this.

src/fancy-tools/brushEraserTool.js
JamesAPetts: I've put this in baseTool. Now all a tool actually has to do to hook up a keybinds is attach a keybind object to config ala:
```
keyBinds: {
  methodOne: '+',
  methodTwo: '-'
}
```

that will hookup methodOne() and methodTwo() to + and -, respectively.

Through the use of pretty new setters, setting new keybinds is just this:
```
const newKeyBinds: {
  methodOne: 'a',
  methodTwo: 'b'
}

tool.keyBinds = newKeyBinds;
```

ofc if no config.keyBinds object is found, no KeyboardController is made.

JamesAPetts:
@dannyrb , improved the keyboard interfacing a bit and put in base tool, see other comments.

src/eventDispatchers/mouseToolEventDispatcher.js
dannyrb: The only one I'm aware of is `keyboardStackScroll`. I "think" it's on @galelis's hit list. This is one of those discussion items where we're not sure if a `keyboard` tool is really just another `interactionParadigm` for a mouse/touch tool. IE.

`wwwcKeyboardTool` could be the `wwwcTool` with keybindings that apply a strategy. It's a question of: what's most intuitive? how would this impact instantiating and activating tools?

src/base/baseBrushTool.js
dannyrb: I "think" I follow, but @swederik may have better thoughts about how to proceed. Times like this I wish we shared an office so we could whiteboard.

src/base/baseBrushTool.js
dannyrb: > Do you intend to have a touch interface for the painting tools?

Eventually? Yes. I could see some of our clinics using tablets or touch screen laptops like the Microsoft Surface. I'm not sure there's an _immediate_ need for Asteris.

src/base/baseBrushTool.js
dannyrb: Feel free to create a discussion item! 🎉

src/base/baseBrushTool.js
dannyrb: @JamesAPetts key collision is a good call out. If we limit ourselves to one active tool per `interaction paradigm` + `mask`, then having `modifier` keys shouldn't produce collisions. Only passive actions with duplicate bindings would be an issue, right?

src/fancy-tools/brushEraserTool.js
dannyrb: I think that makes sense and is a good solution. If we want to account for collisions, we can just modify the `KeyboardController`, and if _any tool_ wants to leverage keybinding, they can do so with simple configuration.

src/fancy-tools/brushTool.js
dannyrb: @JamesAPetts I'm okay with updating the `newImageEventDispatcher` to include tools with `passive`, and `enabled` modes (like the `render` dispatcher). Then we can allow each tool that implements the callback decide which modes it will care about that event for.

Unlike the other callbacks, this one isn't traditionally prevented/stopped from reaching other tools, so that should be a safe and reasonable approach.

src/base/baseBrushTool.js
JamesAPetts: I can explain it on Thursday more clearly :).

src/fancy-tools/brushTool.js
JamesAPetts: Alrighty, I'll make this update tomorrow then.

src/base/baseBrushTool.js
JamesAPetts: I reckon you could make it work, but the minimum brush size would have to be ~size of fingertip. This would be fine for quick drawing for demonstration etc. I'm assuming anyone who wants to draw out very detailed regions would use a computer?

dannyrb: More than likely. The use case would be less diagnostic, and more demonstration. Our software is used in a few classrooms, and sometimes for patient education.

package.json
swederik: Shouldn't we use babel-runtime as the dependency, and transform-runtime as the dev dependency? https://babeljs.io/docs/en/babel-plugin-transform-runtime.html#installation

src/base/baseBrushTool.js
```diff
@@ -0,0 +1,331 @@
+import external from './../externalModules.js';
+import EVENTS from './../events.js';
+import baseTool from './../base/baseTool.js';
+// Utils
+import isToolActive from '../fancy-tools/shared/isToolActive.js';
+import { getNewContext } from '../util/drawing.js';
+import { COLOR_MAP_ID } from '../stateManagement/brushToolColors.js';
+
+export default class extends baseTool {
```
swederik: For searchability, can we include the name baseBrushTool when exporting?

src/base/baseBrushTool.js
```diff
@@ -0,0 +1,331 @@
...
+
+  /**
+   * Paints the data to the canvas.
+   *
+   * @protected
+   * @abstract
+   * @param  {Object} eventData The data object associated with the event.
+   */
+  _paint (eventData) {
+    throw new Error(`Method renderBrush not implemented for ${this.toolName}.`);
```
swederik: Shouldn't this be "Method _paint..."?

src/base/baseTool.js
```diff
@@ -24,6 +28,13 @@ export default class {
 
     // True if tool has a custom cursor, causes the frame to render on every mouse move when the tool is active.
     this.hasCursor = false;
+
+    // Setup keybinds if present.
+    const keyBinds = this.keyBinds;
+
+    if (keyBinds) {
```
swederik: I don't understand these five lines. Looks like we're just copying one variable into another and back again?

package.json
JamesAPetts: Hmm, you're right.

src/base/baseBrushTool.js
JamesAPetts: Copy + paste. Yup.

src/fancy-tools/brushTool.js
```diff
@@ -0,0 +1,289 @@
...
+function defaultBrushToolConfiguration () {
+  return {
+    draw: 1,
+    radius: 10,
+    minRadius: 1,
+    maxRadius: 50,
+    brushAlpha: 0.4,
+    hoverColor: toolColors.getToolColor(),
+    dragColor: toolColors.getActiveColor(),
+    drawing: false,
```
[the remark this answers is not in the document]
JamesAPetts: would make sense as just a member of the class, I'll change this.

src/fancy-tools/shared/KeyboardController.js
```diff
@@ -0,0 +1,58 @@
+import { getKey } from './keyCodes.js';
+
+export default class {
+  constructor (tool, keyBinds) {
+    const keyInterface = this.constructor._generateKeyInterface(tool, keyBinds);
+
+    Object.keys(keyInterface).forEach((key) => {
+      if (!(typeof keyInterface[key] === 'function')) {
```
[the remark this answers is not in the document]
JamesAPetts: Indeed.

src/fancy-tools/shared/KeyboardController.js
```diff
@@ -0,0 +1,58 @@
...
+    const keyInterface = this._keyInterface;
+
+    let imageNeedsUpdate = false;
+
+    Object.keys(keyInterface).forEach((key) => {
+      if (keyPressed === key) {
+        keyInterface[key]();
+        imageNeedsUpdate = true;
+
+        return;
```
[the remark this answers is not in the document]
JamesAPetts: Yeah I was intending to break out the loop. Array.some seems the way to go, thanks.

src/base/baseBrushTool.js
JamesAPetts: I'll go through and do this for all the tools as we approach the end, as well as change classes to be capitalized.

PR #518, opened 2018-08-03
V3 next round of mouseToolEventDispatcher

JamesAPetts:
I thought the checking of functions in the eventDispatcher was a bit confusing in strange. So instead I gave the base classes default implementations of activeMouseDownCallback, handleSelectedCallback and toolSelectedCallback.

All current mouse tools in vNext branch are opperational as prior to this PR.

The largest changes are to the control flow of the mouseDown and mouseMove controllers of mouseToolEventDispatcher.


src/base/baseAnnotationTool.js
```diff
@@ -1,5 +1,13 @@
 import baseTool from './baseTool.js';
 
+import { getToolState } from './../stateManagement/toolState.js';
+import handleActivator from './../manipulators/handleActivator.js';
+
+import {
+  moveHandleNearImagePoint,
+  moveAnnotationNearClick
+} from './baseAnnotationToolHelpers.js';
```
dannyrb: We may need to expose `moveHandleNearImagePoint` and `moveAnnotationNearClick` in a better way for custom tools built outside of this framework; although; the current implementation will probably work fine for the majority of them 🤔

src/eventDispatchers/mouseEventHandlers/mouseDown.js
```diff
@@ -2,12 +2,13 @@ import external from './../../externalModules.js';
...
-import getHandleNearImagePoint from './../../manipulators/getHandleNearImagePoint.js';
-import moveAllHandles from './../../manipulators/moveAllHandles.js';
-import moveHandle from './../../manipulators/moveHandle.js';
 // Util
 import isMouseButtonEnabled from './../../util/isMouseButtonEnabled.js';
+import {
+  getToolsWithMovableHandles,
+  findHandleDataNearImagePoint,
+  findAnnotationNearClick
+} from '../../base/baseAnnotationToolHelpers.js';
```
dannyrb: If we're going to use these here, maybe a different shared/file name?

src/eventDispatchers/mouseEventHandlers/mouseDown.js
```diff
@@ -48,14 +49,12 @@ export default function (evt) {
   if (activeTools.length > 0) {
     // TODO: If length > 1, you could assess fitness and select the ideal tool
     // TODO: But because we're locking this to 'active' tools, that should rarely be an issue
+    // Super-Meta-TODO: ^ I think we should just take the approach of one active tool per mouse button?
```
dannyrb: I mean... Probably, haha. It's still a little early to put in the constraint, but I imagine that's what will shake out. Having more than one active tool per mouse button, touch event type, keybinding, etc. feels messy.

src/base/baseAnnotationTool.js
JamesAPetts: We could also expose them in some clever way to be utilised by tools that extend this framwork, maybe even just a reference would suffice to them in e.g. baseAnnotationTool?

src/eventDispatchers/mouseEventHandlers/mouseDown.js
JamesAPetts: Yeah you are correct, we should probably go on a renaming frenzy nearer completion of all the tools.

src/eventDispatchers/mouseEventHandlers/mouseDown.js
JamesAPetts: ;). Time will tell.

PR #519, opened 2018-08-06
Fix/v3 wwwc region

dannyrb:
@galelis did you want the "two click" method to work? Or only click + drag?

### DIFF
--- a/src/base/baseAnnotationTool.js
+++ b/src/base/baseAnnotationTool.js
@@ -25,9 +25,9 @@ export default class extends baseTool {
     });
   }
 
-  //===================================================================
+  // ===================================================================
   // Abstract Methods - Must be implemented.
-  //===================================================================
+  // ===================================================================
 
   /**
    * @abstract Creates a new annotation.
@@ -36,7 +36,9 @@ export default class extends baseTool {
    * @return {type}     description
    */
   createNewMeasurement (evt) {
-    throw new Error(`Method createNewMeasurement not implemented for ${this.toolName}.`);
+    throw new Error(
+      `Method createNewMeasurement not implemented for ${this.toolName}.`
+    );
   }
 
   /**
@@ -48,7 +50,9 @@ export default class extends baseTool {
    * @returns {boolean} If the point is near the tool
    */
   pointNearTool (element, data, coords) {
-    throw new Error(`Method pointNearTool not implemented for ${this.toolName}.`);
+    throw new Error(
+      `Method pointNearTool not implemented for ${this.toolName}.`
+    );
   }
 
   /**
@@ -62,7 +66,9 @@ export default class extends baseTool {
    * calculated.
    */
   distanceFromPoint (element, data, coords) {
-    throw new Error(`Method distanceFromPoint not implemented for ${this.toolName}.`);
+    throw new Error(
+      `Method distanceFromPoint not implemented for ${this.toolName}.`
+    );
   }
 
   /**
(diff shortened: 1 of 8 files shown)


dannyrb:
Still need to fix the tests I broke, but this should be functioning. @galelis please weigh in on tool changes. @JamesAPetts would like your thoughts on my dispatcher/baseTool changes.

src/fancy-tools/wwwcRegionTool.js
```diff
@@ -2,102 +2,33 @@
...
-};
-
 export default class extends baseTool {
   constructor (name) {
-    const strategies = {
-      applyWWWCRegion
-    };
-
     super({
       name: name || 'wwwcRegion',
```
JamesAPetts: @swederik  gave me feedback on a previous PR that I think we should follow everywhere: replace
```name: name || 'wwwcRegion',```
with
```name,```
and then use a default parameter in the constructor:
```constructor (name = 'wwwcRegion) ```.

File: src/fancy-tools/wwwcRegionTool.js [hunk 2]
```diff
@@ -2,102 +2,33 @@
...
         minWindowWidth: 10
       }
     });
 
-    this._mouseUpCallback = this._mouseUpCallback.bind(this);
-    this._mouseMoveCallback = this._mouseMoveCallback.bind(this);
-
     this._resetHandles();
+    // Touch
+    this.activeTouchStartCallback = this._startOutliningRegion.bind(this);
```
JamesAPetts: I'm in two minds about this. Obviously we are just starting to flesh out all the tools with touch interfaces. I wonder if its best to have everything in one class, like here. Or two classes e.g wwwcMouseTool and wwwcTouchTool. This particular case isn't too heavy, but I'm wondering if it makes the class difficult to parse by eye. Not sure 🤷‍♂️ .

src/fancy-tools/wwwcRegionTool.js
```diff
@@ -147,63 +73,135 @@ export default class extends baseTool {
...
-  *
-  * @private
-  * @param {Object} element - The viewport element to add event listeners to.
-  * @modifies {element}
-  */
-  _activateModify (element) {
-    element.addEventListener(EVENTS.MOUSE_UP, this._mouseUpCallback);
-    element.addEventListener(EVENTS.MOUSE_CLICK, this._mouseUpCallback);
-    element.addEventListener(EVENTS.MOUSE_MOVE, this._mouseMoveCallback);
+/**
```
JamesAPetts: I agree helpers look nicer at the bottom, so you see the class upon opening the file. I've wondered though (and I'm guilty of the same), if we aren't reusing these somewhere, should they not just be class methods?

dannyrb: I could see them being private class methods. I wasn't sure if they were generic enough to be pulled into a shared module as potentially reusable code though?

src/base/baseAnnotationTool.js
```diff
@@ -36,7 +36,9 @@ export default class extends baseTool {
    * @return {type}     description
    */
   createNewMeasurement (evt) {
-    throw new Error(`Method createNewMeasurement not implemented for ${this.toolName}.`);
+    throw new Error(
```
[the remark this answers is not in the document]
dannyrb: Same thing here. Autostyled on save. Let me know your preference and I can update our defaults.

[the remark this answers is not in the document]
JamesAPetts: 🤷‍♂️ , this is fine and I'm happy to follow suit.

src/fancy-tools/wwwcRegionTool.js
JamesAPetts: I think we should leave it as you have it now for this PR. Nearer the end we can see if this functionality should be pulled out or if we should make it private.

src/fancy-tools/wwwcRegionTool.js
JamesAPetts: Yeah, I think this is still better currently. I'm not too hot on touch controls, is there effectively a 1:1 analogue with e.g. drag/click, or is the touch functionality more bespoke than that? If their is an analogue we could probably get rid of a load of boilerplate with some clever interfacing and inheritance.

dannyrb: @JamesAPetts, I think there's definitely a path for some consolidation. I'm hoping as well wrap up the remaining "baked in tools" that the path becomes a little clearer.

PR #524, opened 2018-08-07
Feature/eraser tool

dannyrb:
@dlwire's `eraserTool` ported over. Adds highlighting 👍 
Had to create `pre` and `post` `mouseDownCallback` hooks to accomplish the behavior. I think `mouseDown` may be the only event that needs it because of our default annotation ranking and selection logic?


dannyrb:
0001-example-eraser

PR #534, opened 2018-08-13
Crosshairtool

galelis:


galelis:
Test: https://deploy-preview-534--cornerstone-tools.netlify.com/examples/crosshairs/

Yet to figured out whats the best way to have an example for tool that interact between viewports

dannyrb:
@galelis can you create a new [project board](https://github.com/cornerstonejs/cornerstoneTools/projects/1) item, convert it to an issue, and assign it to me? Just a simple list of things you'd need the primary example to be able to support so we can continue to use it as a quick place to check one or more tools?

PR #535, opened 2018-08-14
MultiTouch Pan Tool: Inicial changes

galelis:


dannyrb:
Two finger pan works well for me. Has the "number of touch points" configuration been ported as well? I couldn't find it:

https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/imageTools/panMultiTouch.js#L18-L22

I believe this allowed you to require 3, 4, 5+ points of contact for the tool to apply it's strategy. It sounds silly, but this allows you to have a different tool applied depending on the number of points of contact.

PR #539, opened 2018-08-16
StackScroll Multi Touch tool

galelis:


galelis:
I have create another PR (#571) that will fix this broken unit tests.

PR #543, opened 2018-08-17
Updating usage of external.cornerstone

galelis:
## Changes
CornerstoneTools files are loaded before having cornerstone into window scope, so, if we store external.cornerstone as a variable outside of the scope of the class/function it will get undefined. 
To have consistency I've changed all new files to always use external.cornerstone, then we will assure that it will aways run the get and pickup the currently window.cornerstone object.


dannyrb:
Ahh. That makes sense. Good catch, @galelis :+1:

PR #552, opened 2018-08-21
DoubleTapZoom tool migration to v3

galelis:
## TODO
- Test in a touch device using netlify (DONE)


galelis:
Unable to test.

src/tools/doubleTapZoomTool.js
```diff
@@ -0,0 +1,18 @@
...
+    super({
+      name,
+      supportedInteractionTypes: ['touch']
+    });
+  }
+
+  doubleTapCallback (evt) {
+    const eventData = evt.detail;
+
+    external.cornerstone.fitToWindow(eventData.element);
```
dannyrb: If I zoom out, then activate this tool, double tapping does fit the image to the window :+1:

It confused me at first, because the name implies that `doubleTap` zooms in? At least in my head. But the code does fire on double tap 👍

swederik: Yeah that's weird, we should probably rename it.

galelis: You guys have any idea of a new name? 
I agreed of changing it, maybe even 'Fit to window" tool.

swederik: doubleTapFitToWindow seems fine

galelis:
I have create another PR (#571) that will fix this broken unit tests.

PR #553, opened 2018-08-22
Add fill the region of brush tool

sisobus:
I've implemented fill inside of segmentation with brush tools.

In the case of the closed space first, double-clicking will color the region.
If it is not a closed space, it will be colored by connecting both ends.
Please refer to attached images.

### DIFF
--- a/src/paintingTools/brush.js
+++ b/src/paintingTools/brush.js
@@ -336,9 +336,6 @@ function fill (eventData) {
         if (ny < 0 || ny >= rows || nx < 0 || nx >= columns || isSameColor(nx, ny, brushPixelValue)) {
           continue;
         }
-        if (!isSameColor(nx, ny, 0)) {
-          continue;
-        }
         q.enqueue({
           x: nx,
           y: ny
--- a/src/paintingTools/brush.js
+++ b/src/paintingTools/brush.js
@@ -111,6 +111,7 @@ function onImageRendered (e) {
 
 function fill (eventData) {
   const configuration = brush.getConfiguration();
+  const radius = configuration.radius;
   const element = eventData.element;
   const layer = external.cornerstone.getLayer(element, configuration.brushLayerId);
   const { rows, columns } = layer.image;
@@ -130,13 +131,7 @@ function fill (eventData) {
   const getPixelIndex = (x, y) => (y * columns) + x;
   const isSameColor = (x, y, color = brushPixelValue) => pixelData[getPixelIndex(x, y)] === color;
   const draw = (x, y, brushPixelValue) => drawBrushPixels([[x, y]], pixelData, brushPixelValue, columns);
-  const drawRange = (range, color) => {
-    range.forEach((point) => {
-      const spIndex = getPixelIndex(point[0], point[1]);
-
-      pixelData[spIndex] = color;
-    });
-  };
+  const drawRange = (range, color) => drawBrushPixels(range, pixelData, color, columns);
 
   x = Math.round(x);
   y = Math.round(y);
@@ -154,12 +149,20 @@ function fill (eventData) {
 
     return null;
   };
+  // This thisCircle = [(x, y - radius), (x + radius, y), (x, y + radius), (x - radius, y)]
   const thisCircle = [
     find(y, 0, (x, y) => !isSameColor(x, y - 1), x, true),
     find(x, columns, (x, y) => !isSameColor(x + 1, y), y, false),
     find(y, rows, (x, y) => !isSameColor(x, y + 1), x, true),
     find(x, 0, (x, y) => !isSameColor(x - 1, y), y, false)
   ];
+
+  if (Math.abs(thisCircle[0] - thisCircle[2]) > radius * 2 && Math.abs(thisCircle[1] - thisCircle[3]) > radius * 2) {
+    // If the position you clicked is aleady in same color, this method is not necessary.
+    return;
+  }
+
+  // Find a direction the closest point where same color from (x, y)
   let [minDistance, idx, base, low, high] = [2147483647, -1];
 
   for (let i = 0; i < thisCircle.length; i += 1) {
@@ -197,7 +200,6 @@ function fill (eventData) {
     return ret;
   };
 
-  drawRange(getPointList(low, high, base, !(idx % 2)), 0);
 
   class Queue {
     constructor () {
@@ -225,7 +227,7 @@ function fill (eventData) {
     }
   }
 
-  const bfs = (x, y, columns, rows) => {
+  const getTheFarthestPoint = (x, y, columns, rows) => {
     let [rx, ry, maxDepth] = [x, y, 0];
     const q = new Queue();
     const d = [];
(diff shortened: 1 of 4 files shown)


JamesAPetts:
Live demo: https://deploy-preview-553--cornerstone-tools.netlify.com/examples/segmentationbrush/

swederik:
I agree with @JamesAPetts, I would also expect it to fill over previous segments.

Very cool addition! Thanks @sisobus!

sisobus:
I agree with @JamesAPetts , I modify some source code and commit.

Thanks for quick review.

src/paintingTools/brush.js
```diff
@@ -107,13 +107,265 @@ function onImageRendered (e) {
   }
 }
 
+// This method is for fill region of segmentation overlays
+
+function fill (eventData) {
```
JamesAPetts: This function is a bit long and has a lot going on (codeclimate gives it a cognitive complexity of 105). For the sake of maintainability, would it be possible to split this function down into a sequence of appropriately named smaller functions, so that it is easier to follow?

src/paintingTools/brush.js
```diff
@@ -107,13 +107,265 @@ function onImageRendered (e) {
...
+
+  const getPixelIndex = (x, y) => (y * columns) + x;
+  const isSameColor = (x, y, color = brushPixelValue) => pixelData[getPixelIndex(x, y)] === color;
+  const draw = (x, y, brushPixelValue) => drawBrushPixels([[x, y]], pixelData, brushPixelValue, columns);
+  const drawRange = (range, color) => drawBrushPixels(range, pixelData, color, columns);
+
+  x = Math.round(x);
+  y = Math.round(y);
+
+  const find = (low, high, condition, base, isY = false) => {
```
JamesAPetts: This function probably doesn't need to be an inner function. Pulling it out may make the logical flow of fill() easier to follow.

src/paintingTools/brush.js
```diff
@@ -107,13 +107,265 @@ function onImageRendered (e) {
...
+    find(x, 0, (x, y) => !isSameColor(x - 1, y), y, false)
+  ];
+
+  if (Math.abs(thisCircle[0] - thisCircle[2]) > radius * 2 && Math.abs(thisCircle[1] - thisCircle[3]) > radius * 2) {
+    // If the position you clicked is aleady in same color, this method is not necessary.
+    return;
+  }
+
+  // Find a direction the closest point where same color from (x, y)
+  let [minDistance, idx, base, low, high] = [2147483647, -1];
```
JamesAPetts: This is more stylistic preference, but I find this line a bit cryptic.
May I suggest you replace this declaration with:

```
let minDistance = Number.MAX_SAFE_INTEGER;
let idx = -1;
let base, low, high;
```

src/paintingTools/brush.js
```diff
@@ -107,13 +107,265 @@ function onImageRendered (e) {
...
+
+  for (let i = 0; i < thisCircle.length; i += 1) {
+    let b, l, h;
```
JamesAPetts: I understand you wish to keep the variable names short here as you are doing a lot of arithmetic, but could we have a comment to show more clearly that b, l, h means base, low, high?

Generally more instructive comments in this for loop would be very much appreciated.

src/paintingTools/brush.js
```diff
@@ -107,13 +107,265 @@ function onImageRendered (e) {
...
+
+    for (let t = low; low < high ? (t <= high) : (t >= high); t += (low < high ? 1 : -1)) {
+      ret.push(isY ? [base, t] : [t, base]);
+    }
+
+    return ret;
+  };
+
+
+  class Queue {
```
JamesAPetts: I think this class should not be an inner class, and should be pulled out of the fill() function. Again, the fill() function could be easier to follow if it just contained the logical flow.

src/paintingTools/brush.js
```diff
@@ -107,13 +107,265 @@ function onImageRendered (e) {
...
+      this.front++;
+
+      return data;
+    }
+    size () {
+      return this.end - this.front;
+    }
+  }
+
+  const getTheFarthestPoint = (x, y, columns, rows) => {
```
JamesAPetts: Again, I think it would be a lot easier to follow if all these functions that work in a functional manner (i.e. that take an input and return a result without changing the state of the input) were pulled out of fill() and placed at the root of the file. Perhaps with appropriate renaming if necessary.

JamesAPetts:
Another small issue, if you double click anywhere outside some regions, the whole screen will be filled with that color, potentially 'erasing' many segmentations.

E.g.:
screen shot 2018-08-22 at 12 10 55
..double click:
screen shot 2018-08-22 at 12 11 06

src/paintingTools/brush.js
sisobus: I think so! I separated fill function to smaller functions named functionally. 
and these are moved to fill.js.

src/paintingTools/brush.js
sisobus: I agree. find function is moved to fill.js

src/paintingTools/brush.js
sisobus: copy & paste :) thanks

src/paintingTools/brush.js
sisobus: This class is moved to queue.js

src/paintingTools/brush.js
sisobus: I took this part and created it as a function (findTheClosestDirection)

sisobus:
Hi, thanks to refactoring some code. It has a bug, and i fixed the bug. :)

JamesAPetts:
Hey, thanks so much for the quick response, the fix works great!

Finally, what do you think about this issue:
https://github.com/cornerstonejs/cornerstoneTools/pull/553#issuecomment-414997154

Perhaps in general it would be better if we only allow the user to fill in closed shapes?
Unless you can find a way to make filling in partially filled shapes without the side effect of painting the whole screen?

Thanks for all the hard work :)!

sisobus:
https://github.com/cornerstonejs/cornerstoneTools/pull/553#issuecomment-414997154

I fixed this issue this commit https://github.com/cornerstonejs/cornerstoneTools/pull/553/commits/26fbd9eaffc7000a78bb3b1c236329e0b29773cb
So, if i double click on the first image, then It will not be painted (refer to second image).



But Because of the spatialy painted problem of the web canvas when draws quickly, there is a problem that it does not work as I intended. (situation of below image)



currently, There is no idea to solve this problem, i think. 

P.S. If the brush tool is drawn continuously, this is a problem to be solved. Do you have any plans for this with the brush tool?

JamesAPetts:
Hey @sisobus, thanks a lot for the fix. I think the edge case you show is relatively minor, and also depends on how the brush tool itself functions. If a clean line is drawn, your addition works as intended. And the consequence of this edge case aren't too severe (a single line at the worst case).

The issue of drawing continuously is an interesting problem, as we are limited by how long the draw call takes currently. We could probably set up some kind of draw queue, but that would be a whole other potential PR/issue to this one.

You can make the tool do strange things if you draw strange shapes, but most of it seems relatively harmless. I'm quite happy with the current state, but I would like to see what @swederik thinks of the changes.

JamesAPetts:
@sisobus, just an update, haven't forgot about this PR, I just would like @swederik to have a look at the most recent changes, and he's currently unavailable for a couple of weeks.

swederik:
I didn't mean to hold this up. I was actually fine with the last set of changes.

Thanks for your hard work @sisobus and you too @JamesAPetts!

Issue #554, opened 2018-08-22
Annotation is not working properly


mpkumar87:
HI,
I am checking cornerstoneTool package in local and tested examples. In Annotation example, it create first annotation with text displaying below the line. But when try to create another annotation and text entered is not displayed for new annotation draw, instead it update the value of first annotation line. Please guide me to solve this problem.

Thanks.

dannyrb:
Hello @mpkumar87, please include a screenshot and code example? Bonus points for an example on codepen that demonstrates the issue.

mpkumar87:
Thanks for ur reply.. Example which is in github also not working for annotation.

https://rawgit.com/cornerstonejs/cornerstoneTools/master/examples/annotation/index.html

First annotation mark display value which i entered in dialog box. But further annotation mark not displaying value, it is getting update first annotation mark.

Thanks.

swederik:
Looks like you're right. I wonder when that broke...

galelis:
I have created a PR to fix this issue
https://github.com/cornerstonejs/cornerstoneTools/pull/557

@mpkumar87 Hey, I have analysed the issue and it was actually an issue only with example code that was not removing correctly the dialog events, so it was being called as many times as we have arrowAnnotations in the app.

param88singh:
Helle swederik,

First of all thanks for your great effort for medical software(Dicom Viewer). I configured your latest versions of  cornerstone library. cornerstone-wado-image-loader - 2.0.0, cornerstone-tools - 2.3.3,
Here i seen annotation is not working accurately. It is almost difference of hundreds in length. But when i tried with yous oldest library(Augast 2017 versions). It is working very fine. Let me know what is the bug in your latest library. It is working good for almost all dicom files. But there is some dicom on which not working fine. I have attached the Example dicom file of DX modality. Check and confirm..
[DicomAnnotationIssue.zip](https://github.com/cornerstonejs/cornerstoneTools/files/2501313/DicomAnnotationIssue.zip)

dannyrb:
@param88singh can you create a new issue with this information? We try to avoid re-opening issues that have merged work.

PR #557, opened 2018-08-23
Fix arrowAnnotate example

galelis:
## Bug Fix
- Click events was not being removed correctly. 
- Fix: Make sure to remove all events on dialog before close it. 

(Closes issue #554)


dannyrb:
@galelis you can almost always use your own discretion when choosing to update an example :+1:

This looks good to me. Thanks for taking the time to knock out a fix 🎉

PR #559, opened 2018-08-29
SaveAs tool v3

galelis:


swederik:
Clicked a button called 'Update Branch'. Not sure if we prefer this approach (Merge vNext into feature branch) versus rebasing the feature branch on top of vNext, though. Not sure if it matter if we squash the commits anyway.

galelis:
I think thats fine, I usually rebase, but this merge will work similarly.

PR #564, opened 2018-09-04
vNext brush rendering and image caching changes - WIP

JamesAPetts:
In this PR I've centralized the rendering of labelmap data, so that it is rendered if any brush tool is active/passive/enabled (i.e. any tool which extends baseBrushTool).

The labelmap is always rendered first in the queue, such that all annotations appear in front of it.

WIP:
The bitmap generation is asynchronous, so I need a cache of the previous image before a new one is drawn. I'm currently caching once per toolData, which allows for seemless usage whilst working with multiple cornerstone enabled elements, however I'm going to look if I can cache just one slice per element. (i.e. one x*y image, rather than a potential whole data cube).

Demo: https://deploy-preview-564--cornerstone-tools.netlify.com/

### DIFF
--- a/src/base/baseBrushTool.js
+++ b/src/base/baseBrushTool.js
@@ -3,7 +3,6 @@ import EVENTS from './../events.js';
 import baseTool from './../base/baseTool.js';
 // Utils
 import isToolActive from '../tools/shared/isToolActive.js';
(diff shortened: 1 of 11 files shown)


JamesAPetts:
Note, moved to #569.

PR #571, opened 2018-09-13
Refactor brushStore and unit tests

galelis:
## Changes
- Make sure that external.cornerstone.colors exist before using 
- It will avoid breaking unit tests that does not mock this functions and also we can remove a lot of mock functions inside unit tests that are not being tested directly. 

NOTE: In case we want to test the colors/colorMap from Brush, we can mock the function inside the unit test and perform the tests desired.

### DIFF
--- a/src/store/brushStore.js
+++ b/src/store/brushStore.js
@@ -65,28 +65,29 @@ export default {
   mutations
 };
 
-
 // DEFAULT BRUSH COLOR MAP
-const colormap = external.cornerstone.colors.getColormap(state.colorMapId);
+if (external.cornerstone && external.cornerstone.colors) {
+  const colormap = external.cornerstone.colors.getColormap(state.colorMapId);
 
-colormap.setNumberOfColors(20);
-colormap.setColor(0, [255, 255, 255, 0]);
-colormap.setColor(1, [230, 25, 75, 255]);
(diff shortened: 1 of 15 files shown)


JamesAPetts:
Sorry I didn't get round to this yet, I'll get on it on Monday.

src/store/brushStore.js
```diff
@@ -65,28 +65,29 @@ export default {
   mutations
 };
 
-
 // DEFAULT BRUSH COLOR MAP
-const colormap = external.cornerstone.colors.getColormap(state.colorMapId);
+if (external.cornerstone && external.cornerstone.colors) {
```
JamesAPetts: Simple and elegant solution to the the mocking issues 👍.

src/tools/PanTool.test.js
```diff
@@ -2,19 +2,8 @@ import PanTool from './PanTool.js';
...
-    colors: {
-      getColormap: jest.fn().mockImplementation(() => {
-        return {
-          setNumberOfColors: jest.fn(),
-          setColor: jest.fn()
-        }
-      })
-    }
-  }
+  cornerstone: {}
```
JamesAPetts: Perfect 👍 .

PR #572, opened 2018-09-13
SaveAs tool fix after latest merges

galelis:
## Changes 
- Adding back SaveAs to cornerstoneTools
- Fix global example


galelis:
@swederik Seems that one of the latest PRs removed some of the code for SaveAs tool and the pattern of the file names changed. This is just an update for it.

JamesAPetts:
Doh, this was my fault, checked all the 'tools' were working, ommited save on the refactor, thanks @galelis.

PR #575, opened 2018-09-19
v3 Docs and support for multiple versions

dannyrb:
### Supporting Multiple Versions

image

- Used the `versions` plugin, which adds a drop down and default version in the top left of the menu.
- The `versions` plugin also takes an endpoint to `master`'s `book.json`, so old versions will update their list if it is outdated
- Store _markdown_ versioned docs side by side in `/docs` ( `/docs/v2`, `/docs/latest`)
- Update the `update-docs.sh` script to...
   - Iterate over folders in `/docs`
   - Do normal gitbook install and build per version
   - Use `/docs/latest` as root, move all other versions' output to a `/history` folder
   - Publish `/docs/latest` using same method as previously

All doc links are relative, so linking is preserved. Version dropdown allows you to hop from version to version. Adding a new version is a simple as updating `book.json`, and moving `/latest` to a `/vNumber` folder.


dannyrb:
image

^ Example of updated `publish` script deploying

dannyrb:
Screenshot of docs w/ new styling and hierarchy:
image

Issue #570, opened 2018-09-13
vNext - Should we allow tools to limit the number of states they have? Or an easy pattern to indicate that they can only be toggled on/off?


JamesAPetts:
For example, `ScaleOverlayTool` only makes sense in `enabled` and `disabled`.

For tools like the scale overlay that should be e.g. on/off, can't we just throw an error in the `passiveCallback` and `activeCallback` which states that the tool isn't intended to be used that way?

We could do this by having an intermediate class such as `BinaryTool`, which extends `BaseTool` and has these error messages baked in the callbacks.

JamesAPetts:
Addressed in #585 :).

PR #585, opened 2018-09-20
BinaryTools

JamesAPetts:
Simple solution to https://github.com/cornerstonejs/cornerstoneTools/issues/570.


JamesAPetts:
Hey @dannyrb , thanks for the detailed feedback!

Yeah, as you know I just tried to bash out an initial idea in the last hour before I left the office ^^.

I agree with your points about diverting default behaviour if we do something wrong and your potential strategies look similar to what I had in mind whilst running to the train! I'll take a deeper thinkg about this when I'm in the office tomorrow and try to write up a more elegant solution :).

JamesAPetts:
> baseTool takes a behaviors array with string names that find and Object.assign matching behaviors from the behaviors folder?

I like this solution, because:
- it gives us the power to check if such a tool can use a mixin (i.e. I could foresee some mixins being `BaseAnnotationTool` specific, for instance.
- It allows us to abstract Object.assign(), which is frankly looks a little obtuse and confusing when mixed in with pretty ES6 syntax.

JamesAPetts:
Took a stab at a having a `mixins` argument for `BaseTool`, and addressed your other points. Note I've left the directory name as `mixins` for now, but I have no real preference for  `mixins` or `behaviors`.

src/base/BaseTool.js
```diff
@@ -94,6 +101,26 @@ export default class BaseTool {
...
+   * Applys the requested mixins to the class.
+   *
+   * @param  {Array} mixinArray An array of mixin names (strings).
+   */
+  _applyMixins (mixins) {
+    for (let i = 0; i < mixins.length; i++) {
+      const mixinName = mixins[i];
+
+      if (typeof mixinCollection[mixinName] === 'object') {
+        Object.assign(this, mixinCollection[mixinName]);
```
dannyrb: Only because I'm way overthinking it, but what if the user has two mixins that want to add a behavior that's called in the `passiveCallback`? The last one applied would override. For each callback, do we want to allow an _array of functions_, where each one is called after the other?

I mean.. Probably, maybe, eventually.
Probably not worth adding in just yet as we don't have any use cases for it yet.
Just food for thought ^_^

PR #589, opened 2018-09-23
Feature: Docs & Implementation for Global Configuration

dannyrb:
# WIP

- [x] This breaks the `brush` module. Refactor it to work with the new setup.
- [x] Not in love w/ the Getters/Setters/Mutations/Modules; brainstorm a better approach


dannyrb:
Hmm... Starting to wonder if a simple JS store would make a lot more sense. Adding in the semantics of state/mutations/getters doesn't make as much sense without some of the benefits afforded by a more mature framework.

I'll need to think on this, and maybe annoy @JamesAPetts for some thoughts. I want to be sure that the approach we take for the store works well enough for our current needs without over-engineering it, but also be able to support new "modules" that can be dropped in by 3rd party tools or as the library grows.

JamesAPetts:
It does feel slightly overkill at the moment, but I was waiting to see if it its complexity would would make sense as vNext matured. I agree it can probably be simpler for now, feel free to DM on slack at some point if you want to arange a chat 👍.

src/eventListeners/onImageRenderedBrushEventHandler.js
```diff
@@ -5,7 +5,7 @@ import external from '../externalModules.js';
 import BaseBrushTool from '../base/BaseBrushTool.js';
 import { getNewContext } from '../util/drawing.js';
 
-const brushState = store.modules.brush;
+const { state, getters, setters } = store.modules.brush;
```
JamesAPetts: Like the use of destructors here. I think going forwards we should follow this pattern in any file that only refers to one store module 👍 .

src/eventListeners/onImageRenderedBrushEventHandler.js
```diff
@@ -16,18 +16,25 @@ export default function (evt) {
...
     for (let i = 0; i < maxSegmentations; i++) {
       addToolState(element, BaseBrushTool.getReferencedToolDataName(), {});
     }
 
     toolData = getToolState(element, BaseBrushTool.getReferencedToolDataName());
 
     // TEMP: HACK: Create first pixel data such that the tool has some data and the brush
-    // cursor can be rendered. Can be replaced once we have a mechanism for SVG cursors.
-    const newPixelData = new Uint8ClampedArray(eventData.image.width * eventData.image.height);
+    // Cursor can be rendered. Can be replaced once we have a mechanism for SVG cursors.
```
JamesAPetts: Keeping the linter happy 👍 .

src/eventListeners/onImageRenderedBrushEventHandler.js
```diff
@@ -54,27 +63,29 @@ function shouldRenderSegmentation (evt, segIndex, segData) {
...
   }
 
-  if (visibleSegmentations[segIndex]) { // Has data and marked as visible, render!
+  if (visibleSegmentations[segIndex]) {
+    // Has data and marked as visible, render!
     return true;
   }
 
-  const currentColor = brushState.getters.draw();
+  const currentColor = state.drawColorId;
```
JamesAPetts: I'm a bit torn on this. Part of me feels like we should call a getter/setter for any interaction with state, but internally there isn't any problem with accessing the state directly. I think we can access the state directly from inside `cornerstoneTools`, but should definitely make sure everything that may changed via an external call has get/set api endpoints

We should probably wrap this in getters, no? I think anything that could be public API (change color surely is, should be changed through a api method we can easily doc, rather than modifying

src/store/addEnabledElement.js
```diff
@@ -12,21 +12,25 @@ import {
...
   // NOTE: the 'enabledElement' argument here is actually the DOM element...
-  const cornerstoneEnabledElement = external.cornerstone.getEnabledElement(enabledElement);
+  const cornerstoneEnabledElement = external.cornerstone.getEnabledElement(
+    enabledElement
+  );
 
   // TEMP
   if (cornerstoneEnabledElement.uuid) {
-    console.error('uuid has gone core! remove this block and the generateGUID function.');
+    console.error(
```
JamesAPetts: This is unrelated to your PR specifically, but lets change this to warn.

src/store/addEnabledElement.js
```diff
@@ -36,13 +40,16 @@ export default function (enabledElement) {
...
   mouseEventListeners.enable(enabledElement);
   mouseWheelEventListeners.enable(enabledElement);
-  touchEventListeners.enable(enabledElement);
 
   // Dispatchers
   imageRenderedEventDispatcher.enable(enabledElement);
   mouseToolEventDispatcher.enable(enabledElement);
   newImageEventDispatcher.enable(enabledElement);
-  touchToolEventDispatcher.enable(enabledElement);
+
```
JamesAPetts: 👍  I like how you've dealt with global config on init.

src/store/index.js
```diff
@@ -38,15 +40,18 @@ export const mutations = {
     state.enabledElements.push(enabledElement);
 
     if (brush) {
-      brush.mutations.SET_ELEMENT_VISIBLE(enabledElement);
+      brush.setters.setElementVisible(enabledElement);
     }
   }
 };
 
+export const modules = {
```
JamesAPetts: For 3rd party support, perhaps we could do with an api hook that adds a module to `modules`. The 3rd party plugin would then be able to change its own state, which can then also be get/set by other plugins that depend on the plugin (if that ever ends up being a thing!).

(I hope this made sense).

src/store/modules/brushModule.js
```diff
@@ -1,7 +1,7 @@
-import external from '../externalModules.js';
+import external from './../../externalModules.js';
 
 const state = {
-  draw: 0,
+  drawColorId: 0,
```
JamesAPetts: Much clearer :+1:.

src/store/modules/brushModule.js
```diff
@@ -13,35 +13,24 @@ const state = {
...
+   * Sets the brush radius, account for global min/max radius
+   *
+   * @param {*} radius
+   */
+  setRadius: (radius) => {
+    state.radius = Math.min(Math.max(radius, state.minRadius), state.maxRadius);
   },
 
   /**
+   * TODO: Should this be a init config property?
```
JamesAPetts: Yes I think so. Hmm, what is a good way to account for 3rd party module configuration during the init call 🤔 ?

src/store/modules/brushModule.js
```diff
@@ -13,35 +13,24 @@ const state = {
...
-  SET_HIDDEN_BUT_ACTIVE_ALPHA: (alpha) => {
-    state.hiddenButActiveAlpha = alpha;
+const setters = {
+
+  /**
+   * Sets the brush radius, account for global min/max radius
+   *
+   * @param {*} radius
+   */
+  setRadius: (radius) => {
```
JamesAPetts: Thanks for filling in the jsdocs! 💯

src/store/modules/brushModule.js
```diff
@@ -66,34 +57,30 @@ const mutations = {
...
+    visible = true
+  ) => {
     if (!state.visibleSegmentations[enabledElementUID]) {
       state.imageBitmapCache[enabledElementUID] = [];
     }
 
     state.visibleSegmentations[enabledElementUID][segIndex] = visible;
   },
-  SET_ELEMENT_IMAGE_BITMAP_CACHE: (enabledElementUID, segIndex, imageBitmap) => {
+  setImageBitmapCacheForElement: (enabledElementUID, segIndex, imageBitmap) => {
```
JamesAPetts: Since we are going back to camelCase, we can safely turn the new-cap eslint rule on, as it should now only be for constructors.

src/store/modules/brushModule.js
```diff
@@ -66,34 +57,30 @@ const mutations = {
...
   },
-  CLEAR_ELEMENT_IMAGE_BITMAP_CACHE: (enabledElementUID) => {
+  clearImageBitmapCacheForElement: (enabledElementUID) => {
     state.imageBitmapCache[enabledElementUID] = [];
   }
-
 };
 
 const getters = {
-  draw: () => state.draw,
```
JamesAPetts: I think we still need these api endpoints so that the parent app has easy accessibility to them?

src/store/modules/brushModule.js
```diff
@@ -113,7 +100,7 @@ const getters = {
 export default {
   state,
   getters,
-  mutations
+  setters
 };
 
 const DISTINCT_COLORS = [
```
JamesAPetts: Lets get rid of the caps here too? distanceColors is fine, I think.

src/tools/BrushTool.js
```diff
@@ -25,10 +20,10 @@ export default class BrushTool extends BaseBrushTool {
   }
 
   /**
-  * Called by the event dispatcher to render the image.
-  *
-  * @param {Object} evt - The event.
-  */
+   * Called by the event dispatcher to render the image.
```
JamesAPetts: Ahh good spot, I've only just started auto-generating these.

src/tools/BrushTool.js
```diff
@@ -50,17 +45,16 @@ export default class BrushTool extends BaseBrushTool {
     const { rows, columns } = eventData.image;
     const { x, y } = mousePosition;
 
-    if (x < 0 || x > columns ||
-      y < 0 || y > rows) {
+    if (x < 0 || x > columns || y < 0 || y > rows) {
```
JamesAPetts: I don't have strong feelings about this either way, but I put this on two lines just because I think the triple `||` is a little confusing. Eh ¯\\_(ツ)\_/¯.

src/tools/BrushTool.js
```diff
@@ -105,26 +105,36 @@ export default class BrushTool extends BaseBrushTool {
       shouldErase = true;
     }
 
-    const segmentationIndex = brushState.getters.draw();
+    const segmentationIndex = state.drawColorId;
 
     if (!toolData.data[segmentationIndex].pixelData) {
-      console.log(`creating new pixelData for segmentation ${segmentationIndex}`);
-      const newPixelData = new Uint8ClampedArray(eventData.image.width * eventData.image.height);
+      console.log(
```
JamesAPetts: We can probably get rid of this log now :).

src/tools/BrushTool.js
dannyrb: That's my auto-formatter being too opinionated. I think `prettier` is just enforcing a line length here.

JamesAPetts: > Prettier is an opinionated code formatter.

JamesAPetts: But yeah, its not really an issue 👍 .

src/store/index.js
dannyrb: I think we'll get to that. Being able to "add shared state" that's namespaced for your tool or suite of tools sounds like a good value add.

src/tools/BrushTool.js
dannyrb: Our generated `api` docs are pretty sad for `cornerstone tools`, haha. I think we'll need to make a strong effort to add a lot of the doc tags recommended by `documentation.js`

src/store/modules/brushModule.js
dannyrb: Maybe not for `vNext`, but we can add a setup/config method as part of the "module interface" that gets called when we first create an instance of the store? 🤷‍♂️

src/eventListeners/onImageRenderedBrushEventHandler.js
dannyrb: > should definitely make sure everything that may changed via an external call has get/set api endpoints

That makes sense to me. Anything that we want to expose as an external API call should have:

- JS Doc comments
- Be unit tested (eventually... haha)
- Use getters/setters

What's a little difficult is that the entire store is a "Public API" from the perspective of 3rd party tool authors. 🤔 

So should `state` only be `private`, and used exclusively by `getters`/`setters`? FWIW, Vue uses:

- state: can be imported and used by anything, but only updated by `mutations`
- getters: for computed/formatted/filtered/mapped values from `state`
- mutations: the only way to change a state property's value

So some a given module might use `state` or `getters` to retrieve a value, but only `mutations` to modify.

src/store/modules/brushModule.js
JamesAPetts: Sounds like a plan, following recent slack conversation, lets also leave this as is for now 👍 .

src/store/modules/brushModule.js
JamesAPetts: Think we resolved this in slack :).

Issue #584, opened 2018-09-20
Problem when using cornerstoneTools and cornerstone as a module in a node.js project


vsaase:
https://github.com/cornerstonejs/cornerstoneTools/blob/1f7d41d15c795dc99f8e395713cf9d63094f49b0/src/externalModules.js#L1

Hey, I am encountering some problems here when loading cornerstone and cornerstoneTools as a module with `require` or `import`. Then `window.cornerstone` is not available to cornerstoneTools.

vsaase:
`window.Hammer` is working, but `window.cornerstone` not. Maybe something with how webpack creates the module file for cornerstone?
 
Thats how I import:
```
import * as Hammer from 'hammerjs';
import * as cornerstone from 'cornerstone-core/dist/cornerstone.js';
import * as cornerstoneMath from 'cornerstone-math/dist/cornerstoneMath.js';
import * as cornerstoneTools from 'cornerstone-tools/dist/cornerstoneTools.js';
import * as cornerstoneWADOImageLoader from 'cornerstone-wado-image-loader/dist/cornerstoneWADOImageLoader.js';

cornerstoneTools.external.Hammer = Hammer;
cornerstoneTools.external.cornerstone = cornerstone;
cornerstoneTools.external.cornerstoneMath = cornerstoneMath;
```

PS: I am trying to make this work for an Ionic app, this is also inside another module
Everything else works just fine.

dannyrb:
@vsaase to @JamesAPetts's point, just make sure you're depending on an _exact_ version, and not loosely in your NPM dependencies. We'll cut a more stable release before too long, but until then, every new release has a chance to introduce a breaking change.

That being said, feedback on what is and is not working for the latest build is always appreciated ^_^

galelis:
@vsaase Hey Victor,
Would you mind to provide me any error log with errors or any other information that could help me to understand your issue? 

I have the @vnext version set up with OHIF Standalone viewer and everything is working fine there as well.  

Ideally, I added that line to prevent the code from breaking in case cornerstone is not present into external (and window)

vsaase:
So I import cornerstone like this:
```
import * as Hammer from 'hammerjs';
import * as cornerstone from 'cornerstone-core/dist/cornerstone.js';
import * as cornerstoneMath from 'cornerstone-math/dist/cornerstoneMath.js';
import * as cornerstoneTools from 'cornerstone-tools/dist/cornerstoneTools.js';
[...]

PR #591, opened 2018-09-24
removing the dependency on cornerstone.colors for the brush colormaps.

vsaase:
this resolves issue #584

I think having an own set of colormaps for cornerstoneTools and brush makes sense, as colormaps for segmentation are very different from colormaps for images.

let me know what you think. also as this is my first pull request let me know what I missed to do

what should I do with the version.js and the package-lock.json files?


dannyrb:
@JamesAPetts this is your baby, and outside my realm of knowledge. Feel free to take charge of any feedback/discussion/merging.

JamesAPetts:
Thanks for the contribution @vsaase! I like the approach, but just having a chat with a few colleagues about the best way to go about this :+1:.

I think I agree that we should have a separate implementation of colormap for `cornerstoneTools`, because (as you say) the colormap in `cornerstone-core` is more designed for image data and may change in the future to be more focused to that regard.

JamesAPetts:
@vsaase: Thank you so much for your hard work, I'm closing this for now, if you'd see my response in #584. Thanks for bringing this issue to our attention.

PR #593, opened 2018-09-24
feature: automatic tracking of enabledElements; keyed off `ELEMENT_ENABLED/DISABLED` events

dannyrb:
# Automatic adding/removal of enabledElements from store

- Adds `removeTool` and `removeToolForElement` API calls
- Moves `addEnabledElement` to internal, and has it fire when `cornerstone` emits `ELEMENT_ENABLED`
- Added `removeEnabledElement`, fires when `cornerstone` emits `ELEMENT_DISABLED`
    - Removes tools from store associated with the disabled element
- Removed `no underscores` linter rule; we use this consistently for private methods

NOTE: commented out `dragProbe` in demo until render error is fixed.


dannyrb:
image

dannyrb:
I won't have a fiddle to push until this is merged. Nothing crazy in this PR, but it does tidy some things up.

src/store/removeTool.js
```diff
@@ -0,0 +1,32 @@
+import { state } from './index.js';
+
+/**
+ * Removes all tools from the target element with the provided name
```
JamesAPetts: This brings up the point, do we want to allow tools to have the same name? Seems a little dangerous with the way the current framework is set up. Although, I also can't see a reason why anyone would choose to name two tools the same on purpose.

src/store/removeTool.js
```diff
@@ -0,0 +1,32 @@
...
+
+  if (toolIndex >= 0) {
+    state.tools.splice(toolIndex, 1);
+  }
+};
+
+/**
+ * Removes all tools from all enabled elements with the provided name
+ *
+ * @export
```
JamesAPetts: This is a nice tag, we should generally try to flesh out the jsdocs a little more at some point. A tangent, but I also like @api on anything that's meant to be an endpoint, would allow us to auto-generate api docs. I digress.

dannyrb: Agreed. I want to finish up the normal docs first, but then circle back and identify best practices for `jsdocs` and how they influence our autogenerated API docs.

src/store/removeTool.js
dannyrb: 🤷‍♂️

_addTool_
```
  const toolAlreadyAddedToElement = getToolForElement(element, tool.toolName);

  if (toolAlreadyAddedToElement) {
    console.warn(
      `${tool.toolName} has already been added to the target element`
    );

    return;
  }
```

For the time being, we don't allow tool's with the same name to be added to the same element

JamesAPetts: Oh, didn't realize that's already a thing! :)

src/store/internals/addEnabledElement.js
```diff
@@ -3,24 +3,31 @@ import {
...
-// TODO: Then we could just respond to those events and not worry about tracking/lifecycle
-export default function (enabledElement) {
-  // TEMP -> We are piggy-backing off Cornerstone here, UUID
-  // Generation should go core later, this is more of a POC.
+} from '../../eventDispatchers/index.js';
+import generateGUID from '../generateGUID.js';
+import external from '../../externalModules.js';
+import store from '../index.js';
+
+// TODO: It would be nice if this automatically added "all tools"
```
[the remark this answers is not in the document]
dannyrb: I'm not sure. I think... Any tool added with the more global `addTool` would be added to a list of tools that should be initially added to any new `enabledElements`. Similarly, `removeTool` would remove from that list.

Whereas `addToolForElement`/`removeToolForElement` would not modify the list of "initial tools" for an element.

I've also toyed with... Having a way to restore the state of a disabled element if the `uuid` of an `enabledElement` matches a previously destroyed `uuid`.

PR #597, opened 2018-09-25
Freehand touch

vsaase:
adds touch support to FreehandMouseTool

I appreciate your comments. 

remaining issues:
- lines are not correctly moved in the touchMoveHandle animation
- needs a global variable for the touchMove aboveFinger offset


src/tools/FreehandMouseTool.js
```diff
@@ -501,20 +501,25 @@ export default class FreehandMouseTool extends BaseAnnotationTool {
 
     const config = this.configuration;
 
-    const nearby = this._pointNearHandleAllTools(eventData);
-    const handleNearby = nearby.handleNearby;
-    const toolIndex = nearby.toolIndex;
-
+    //const nearby = this._pointNearHandleAllTools(eventData);
```
JamesAPetts: What is the rationale for replicating _pointNearHandleAllTools() with a loop doing the same thing? It was factored into functions to make this call easier to follow complicated. Happy to discuss this point if you think there is a valid reason to change this.

src/tools/FreehandMouseTool.js
```diff
@@ -501,20 +501,25 @@ export default class FreehandMouseTool extends BaseAnnotationTool {
...
+      x: toolState.data[config.currentTool].handles[config.currentHandle].x,
+      y: toolState.data[config.currentTool].handles[config.currentHandle].y
     };
 
     this._modifying = true;
-    config.currentHandle = handleNearby;
-    config.currentTool = toolIndex;
-
-    this._activateModifty(element);
+    this._activateModify(element);
```
JamesAPetts: Thanks for correcting this!

src/tools/FreehandMouseTool.js
```diff
@@ -1001,6 +1029,11 @@ export default class FreehandMouseTool extends BaseAnnotationTool {
     element.addEventListener(EVENTS.MOUSE_DOWN, this._drawingMouseDownCallback);
     element.addEventListener(EVENTS.MOUSE_MOVE, this._drawingMouseMoveCallback);
 
+    element.addEventListener(EVENTS.TOUCH_START, this._drawingMouseMoveCallback);
```
JamesAPetts: @dannyrb, should TOUCH_START really be mapped to the mouseMoveCallback? This isn't my area of expertise.

JamesAPetts:
@dannyrb, I had a look through the code, but would you be able to chime in on the touch-specifics, and check this works well on an actual touch device?

src/tools/FreehandMouseTool.js
vsaase: _pointNearHandleAllTools, by calling _pointNearHandle, checks if the point is less than a distance of config.spacing (=5 for default) away from the handle. When we call handleSelectedCallback we already know that a handle was selected. In touchStart.js the distance must be less than distanceFromHandle (=28 for default) for the handleSelectedCallback to be called.

src/tools/FreehandMouseTool.js
vsaase: i'm not quite sure why this is needed, but it only works that way at the moment

src/tools/FreehandMouseTool.js
vsaase: maybe config.spacing has to be changed anyway when using touch, as the value of 5 makes it hard to complete a polygon

vsaase:
pencil mode works for me. when increasing config.spacing, it gets easier to complete the drawing, but the polygon becomes more jagged

vsaase:
thanks for offering to invite me to slack, here is my email at gmail.com: vsaase

JamesAPetts:
@vsaase aha, nice that you got pencil mode too. I send you an invite to the cornerstonejs slack.
I replied to your comment, other than that I'm happy, but I'll wait on @dannyrb  to chime in before we merge.

Thanks again for the contribution and nice job 💯 .

dannyrb:
On my list to review this -- Bit of a crazy day. I have the first part of tomorrow blocked off on my calendar to take a peak :+1:

src/eventDispatchers/touchEventHandlers/touchStart.js
```diff
@@ -96,17 +96,23 @@ export default function (evt) {
       coords,
       distanceFromHandle
     );
+    
+    firstToolWithMoveableHandles.handleSelectedCallback(evt, moveableHandle, dataWithMoveableHandle);
```
dannyrb: Do we need a check to make sure this callback exists on the tool?

src/manipulators/touchMoveHandle.js
```diff
@@ -14,7 +14,7 @@ const touchEndEvents = [
   EVENTS.TOUCH_END,
   EVENTS.TOUCH_DRAG_END,
   EVENTS.TOUCH_PINCH,
-  EVENTS.TOUCH_PRESS,
+//  EVENTS.TOUCH_PRESS,
```
dannyrb: If we're removing this, then we probably also want to remove this:

https://github.com/cornerstonejs/cornerstoneTools/blob/20f9c715ca952f0e92a03c08c79c2a3daaef7b17/src/manipulators/touchMoveHandle.js#L128-L133

I'm not 100% sold on this being the fix, but I agree that there's something wonky going on with `TOUCH_PRESS`. Did you notice any other issues with other tools when making this change?

@swederik do we have any tools that care about the "TOUCH_PRESS" event?

src/tools/FreehandTool.js
```diff
@@ -612,17 +615,40 @@ export default class FreehandMouseTool extends BaseAnnotationTool {
...
     data.active = true;
     data.highlight = true;
-    data.handles[currentHandle].x = config.mouseLocation.handles.start.x;
-    data.handles[currentHandle].y = config.mouseLocation.handles.start.y;
-
-    if (currentHandle !== 0) {
-      const lastLineIndex = data.handles[currentHandle - 1].lines.length - 1;
-      const lastLine = data.handles[currentHandle - 1].lines[lastLineIndex];
+    
+    if(eventData.type == EVENTS.TOUCH_DRAG){
```
dannyrb: When you bind `_editMouseDragCallback` to specific events, you can set an parameter to indicate if the event is touch or mouse instead of checking for this here? Similarly, if you feel the method deviates enough, you can separate out the touch/mouse logic into separate methods that are called after the shared logic is executed? (This might make unit testing a bit easier, and the code easier to read)

src/tools/FreehandMouseTool.js
dannyrb: This is all a little odd as `freehand` deviates from most other tools by having it's own "event loop". Because the tool is locked at this point, you're relying on the events emitted by cornerstone, where the dispatchers will continue to return prematurely because of the lock.

It looks like... `_activateDraw` is only called if we're trying to create new data.  I would need to trace these to see what's going on with these events, and why it only works when setup this way. 🤷‍♂️

dannyrb:
`touchStart.js` and `touchMoveHandle.js` are the only items that need scrutinized a little harder before I'd merge this, as they have the potential to mess with other tools. I'll look into the `TOUCH_PRESS` event a little further.

@vsaase please let me know your thoughts. If we can tidy up those two parts, and assuming you have the green light from @JamesAPetts, I think this should be merge-able fairly soon 👍

src/eventDispatchers/touchEventHandlers/touchStart.js
JamesAPetts: Its a virtual member function, with a default implementation for annotation tools, so we are safe to just call it 👍 .

src/tools/FreehandMouseTool.js
JamesAPetts: Maybe its just so that the lines get updated with mouse position? Feel free to make a separate method if binding all of `_drawingMouseMoveCallback` is overkill.

JamesAPetts:
After a little tidying/comment up of `handleSelectedCallback`, and when the `TOUCH_PRESS` stuff is resolved, I'm rather happy. I made a comment about `_drawingMouseMoveCallback`, if it makes sense to make a touch version instead of piggybacking on this method, feel free.

src/tools/FreehandTool.js
vsaase: ok, methods separated

src/manipulators/touchMoveHandle.js
vsaase: unfortunately I do not know much about that, maybe some problems with touchMoveAllHandles? how do I move all handles btw?

src/tools/FreehandMouseTool.js
JamesAPetts: With the recent changes to freehand it be better to use the same drag mechanism as mouse uses.

dannyrb:
@vsaase is this ready for review? I see some outstanding conflicts -- not sure what/if you need from us.

vsaase:
hi, yes, this is ready for review, might be outdated though. atm i cant find time to further work on this I am afraid

JamesAPetts:
Hey, I've picked this up as inspiration and packaged it in #892. Feel free to have a look at my attempt there, thanks a lot for this!

dannyrb:
@all-contributors please add @vsaase for his helpful ideas!

PR #598, opened 2018-09-26
vNext - Module initialisation

JamesAPetts:
Initialize modules upon initialization of cornerstoneTools on a cornerstone enabled element.
The brush module actually only initializes the `colormap` once, as the map is global (for now). But in the general case an initialization routine can run on each new `enabledElement`.

Any module that exposes an `init (enabledElement) `function will have this called when `addEnabledElement` is called.

This should hopefully fix the issues @vsaase was having.


vsaase:
ok, confirmed, this fixes my issue

dannyrb:
Instead of calling this per added element, would this make more sense to call in the `cornerstoneTools.init()`?

JamesAPetts:
Perhaps, Its only used for the brush right now obviously, but could see some modules wanting to initialize per element (e.g. for some overlay module, checking what modality of image is loaded)?

Not sure if its all that useful. Maybe we should just stick it in `cornerstoneTools.init()` as you say?

JamesAPetts:
Actually I might add a little more to the api for registering 3rd party modules...

PR #606, opened 2018-09-27
Feature/smart tool mode changing

dannyrb:
### Previously
```js
csTools.setToolActive('Length') // OR
csTools.setToolActive('Length', { mouseButtonMask: 1 });
```

### Now
```js
// ~~ Still works
csTools.setToolActive('Length'); // OR
csTools.setToolActive('Length', { mouseButtonMask: 1 }); // OR

// ~~ Also Works
csTools.setToolActive('Length', {
    mouseButtonMask: 1,
    isTouchActive: true
}); // OR
csTools.setToolActive('Length', 1); // OR
csTools.setToolActive('Length', 1, true); // OR
```

Addresses: https://github.com/cornerstonejs/cornerstoneTools/projects/1#card-12693686
CC: @diego0020 

> We need a way to _not_ activate a tool for touch, such that activating many tools at once doesn't leave multiple tools for touch fighting for an event by accident.

### DIFF
--- a/src/base/BaseTool.js
+++ b/src/base/BaseTool.js
@@ -49,8 +49,13 @@ export default class BaseTool {
     return this._options;
   }
 
+  /**
+   * Merges provided options with existing options
+   *
+   * @memberof BaseTool
+   */
   set options (options) {
-    this._options = options;
+    this._options = Object.assign({}, this._options, options);
   }
 
   clearOptions () {
--- a/src/base/BaseTool.js
+++ b/src/base/BaseTool.js
@@ -92,9 +92,7 @@ export default class BaseTool {
       if (typeof mixinCollection[mixinName] === 'object') {
         Object.assign(this, mixinCollection[mixinName]);
       } else {
-        console.warn(
-          `${this.name}: mixin ${mixins} does not exist.`
-        );
+        console.warn(`${this.name}: mixin ${mixins} does not exist.`);
       }
     }
   }
--- a/src/base/BaseTool.js
+++ b/src/base/BaseTool.js
@@ -69,9 +69,9 @@ export default class BaseTool {
    */
   clearOptions () {
     this._options = {};
-    this.internalOptions().forEach((option) => {
-      this._options[option] = undefined;
-    });
+    this.internalOptions().forEach(
+      (option) => (this._options[option] = undefined)
+    );
   }
 
   /**
--- a/src/base/BaseTool.js
+++ b/src/base/BaseTool.js
@@ -58,6 +58,11 @@ export default class BaseTool {
     this._options = Object.assign({}, this._options, options);
   }
 
+  /**
+   * Clears the tools options.
+   *
+   * @memberof BaseTool
+   */
   clearOptions () {
     this._options = {};
   }
--- a/src/base/BaseTool.js
+++ b/src/base/BaseTool.js
@@ -24,7 +24,10 @@ export default class BaseTool {
 
     //
     this.data = {};
+    // Options are set when a tool is added, during a "mode" change,
+    // Or via a tool's option's setter
     this._options = {};
+    // Configuration is set at tool initalization
     this._configuration = Object.assign({}, configuration);
 
     // True if tool has a custom cursor, causes the frame to render on every mouse move when the tool is active.
--- a/src/base/BaseTool.js
+++ b/src/base/BaseTool.js
@@ -62,12 +62,45 @@ export default class BaseTool {
   }
 
   /**
-   * Clears the tools options.
+   * Clears the tools options; preserves internal options, but
+   * with `undefined` values.
    *
    * @memberof BaseTool
    */
   clearOptions () {
     this._options = {};
+    this.internalOptions().forEach((option) => {
+      this._options[option] = undefined;
+    });
+  }
+
+  /**
+   * Internal options that "MUST" be in the tool's options if
+   * certain conditions are met. Method is also good for inspecting
+   * options that can be used to change tool behavior.
+   *
+   * @readonly
+   * @memberof BaseTool
+   */
+  get internalOptions () {
+    const internalOptions = [];
+
+    // Should be on _every_ mouse tool
+    if (this.supportedInteractionTypes.contains('mouse')) {
+      internalOptions.push('mouseButtonMask');
+    }
+
+    // Should be on _every_ tool that is touch AND mouse?
+    // Because doubleTap, pinch, etc. won't ever be set to "ACTIVE"
+    // Unless the intent is to set `isTouchActive` true?
+    if (
+      this.supportedInteractionTypes.contains('mouse') &&
+      this.supportedInteractionTypes.contains('touch')
+    ) {
+      internalOptions.push('isTouchActive');
+    }
+
+    return internalOptions;
   }
 
   /**
--- a/src/base/BaseTool.js
+++ b/src/base/BaseTool.js
@@ -57,7 +57,7 @@ export default class BaseTool {
    *
    * @memberof BaseTool
    */
-  set options (options) {
+  addOptions (options) {
     this._options = Object.assign({}, this._options, options);
   }
 
--- a/src/eventDispatchers/shared/customCallbackHandler.js
+++ b/src/eventDispatchers/shared/customCallbackHandler.js
@@ -15,10 +15,11 @@ export default function (handlerType, customFunction, evt) {
 
   if (handlerType === 'touch') {
     tools = tools.filter(
-      (tool) => tool.options.touchEnable || tool.options.touchEnable === undefined
-    );  
+      (tool) =>
+        tool.options.isTouchActive || tool.options.isTouchActive === undefined
+    );
   }
-  
+
   tools = tools.filter((tool) => typeof tool[customFunction] === 'function');
 
   if (tools.length === 0) {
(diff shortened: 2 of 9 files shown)


dannyrb:
This is only the first prong of a two tong approach to address:

> We need a way to _not_ activate a tool for touch, such that activating many tools at once doesn't leave multiple tools for touch fighting for an event by accident.

We'll also try flipping tools to `passive` if a tool is `active` and overlaps with `mouseButtonMask` or `isTouchActive`

dannyrb:
https://github.com/cornerstonejs/cornerstoneTools/projects/1#card-11916782

dannyrb:
- Don't allow tools w/o `touch` to set `isTouchActive`?
- Don't allow tools w/o `mouse` to set `mouseButtonMask`?
- Left + Right click is error prone

dannyrb:
@JamesAPetts this PR tackles to project board issues. Curious to hear your feedback -- Should make for a good Friday discussion. If these work as expected, and are what we want, I'll drum up the appropriate docs.

Complexity like this makes me think these core methods need unit tests asap.

src/store/setToolMode.js
```diff
@@ -8,67 +8,137 @@ import { state } from './../store/index.js';
  * respond to user input, and can create new data
  *
  * @export
- * @param {*} element
- * @param {*} toolName
- * @param {*} options
+ * @param {object} element
```
JamesAPetts: Side note: We need to blitz the whole lib before release and fix all these params, `Object` should be upper case for instance. I've also used `Array` incorrectly throughout, this is how it should be done:
```
{Object}
{string}
{number}
{boolean}
{Object[]}
{string[]}
{number[]}
{boolean[]}
```

Also if something is optional, we put the variable name in brackets after, e.g.:

`@param {(Object|number} [options]`

further reading: [here](http://usejsdoc.org/tags-param.html) and [here](http://usejsdoc.org/tags-type.html).

src/store/setToolMode.js
```diff
@@ -125,17 +216,104 @@ function setToolModeForElement (mode, changeEvent, element, toolName, options) {
...
+ * @param {*} isTouchActive
+ * @private
+ */
+function _setActiveToolWithMatchingMouseButtonMaskToPassive (
+  tool,
+  element,
+  options
+) {
+  const mouseButtonMask =
+    (typeof options === 'number' ? options : options.mouseButtonMask) ||
```
JamesAPetts: Again, maybe we could do something a bit neater with default options to make it so checks like this don't need to happen? Is there a particular reason that options should be just a number?

src/store/setToolMode.js
```diff
@@ -8,67 +8,137 @@ import { state } from './../store/index.js';
...
+  element,
+  toolName,
+  options,
+  isTouchActive
+) {
+  const tool = getToolForElement(element, toolName);
+
+  if (tool) {
+    _setActiveToolWithMatchingMouseButtonMaskToPassive(tool, element, options);
+    _setActiveToolWithIsTouchActiveToFalse(
```
JamesAPetts: Both of these function names were very difficult to parse for me, I had to look at the function to get what was trying to be done. I think its because we are in `setToolActiveForElement`, and now we are talking about the _previously_ active tool straight away (if I understand correctly). I reckon this either needs a lot of comments, or lines 23 to 36 refactored to another function such as `resolveInputConflicts()` or something.

Not sure on the best solution, feel free to bounce back with a better idea.

src/store/setToolMode.js
```diff
@@ -125,17 +216,104 @@ function setToolModeForElement (mode, changeEvent, element, toolName, options) {
...
 
+/**
+ *
+ *
+ * @param {*} element
+ * @param {*} options
+ * @param {*} isTouchActive
+ * @private
+ */
+function _setActiveToolWithMatchingMouseButtonMaskToPassive (
```
JamesAPetts: Other than minor comments, the implementation looks good to me 👍.

src/store/setToolMode.js
```diff
@@ -125,17 +216,104 @@ function setToolModeForElement (mode, changeEvent, element, toolName, options) {
...
+function _setActiveToolWithIsTouchActiveToFalse (
+  tool,
+  element,
+  options,
+  isTouchActive
+) {
+  let value;
+
+  if (typeof options === 'number' && typeof isTouchActive === 'boolean') {
+    value = isTouchActive === true;
```
JamesAPetts: Can't this just be `value = isTouchActive;`? You are already checking if its `boolean`, so its not going to be `undefined`.

src/store/setToolMode.js
```diff
@@ -125,17 +216,104 @@ function setToolModeForElement (mode, changeEvent, element, toolName, options) {
...
+  } else if (
+    typeof options === 'object' &&
+    typeof options.isTouchActive === 'boolean'
+  ) {
+    value = options.isTouchActive === true;
```
JamesAPetts: And again, `value = options.isTouchActive;`?

src/store/setToolMode.js
```diff
@@ -79,27 +149,48 @@ const setToolPassive = setToolMode.bind(
...
- * @param {*} changeEvent
+ * @param {string} mode
+ * @param {string} changeEvent
+ * @param {object} element
+ * @param {string} toolName
+ * @param {(object|number)} options
+ * @param {boolean} isTouchActive
  * @returns
  */
-function setToolModeForElement (mode, changeEvent, element, toolName, options) {
```
JamesAPetts: I feel what you say about needing unit tests 😅 .

src/eventDispatchers/shared/customCallbackHandler.js
```diff
@@ -15,10 +15,11 @@ export default function (handlerType, customFunction, evt) {
 
   if (handlerType === 'touch') {
     tools = tools.filter(
-      (tool) => tool.options.touchEnable || tool.options.touchEnable === undefined
-    );  
+      (tool) =>
```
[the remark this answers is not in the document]
dannyrb: Maybe we shouldn't treat `undefined` as touch enabled? There was a previous bit of code PR'd from someone on @diego0020's team, I believe, that had it working this way. I was mostly updating the config option's name here.

I feel like setting `touch` active should be deliberate? We don't default `mouseButtonMask` to 1, so we probably shouldn't default `touch` to active either.

> We could have some default _configuration in BaseTool, and then Object.assign(this._configuration, confguration), where configuration is the argument passed to the constructor of BaseTool.

I think that makes sense for initialization.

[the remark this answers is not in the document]
dannyrb: `isTouchActive` is weird. It's primarily meant to make activating a tool more clear when it works for both `touch` and `mouse`. Kind of a... YOU ARE activating this for mouse, but "are you sure?" you want it to become the active `touch` tool as well.

If that's what we're using it for, then it doesn't make sense on any `touch` tool that doesn't also support `mouse`. For example, a `doubleTap` or gesture based tool would not be activated ONLY for mouse, so it's safe to assume we're activating it for it's specific `touch` scenario.

It's almost like we need a `touchMask` that indicates "off", single finger, 2-finger-touch, 3-finger-touch, specific gesture, etc. But with the caveat that some `touch` tools can only support a subset of that mask.

Confusing stuff. As of right now, I can't think of an elegant solution -- what we have "works", but I imagine this will be something that gets polished up in a later major version.

dannyrb:
@JamesAPetts:

I think you'll find the changes much cleaner. Interested to hear your thoughts on the added `internalOptions` method on `baseTool`

My biggest hang-up is how to exclude `touch` tools that don't also support `mouse` from the `isTouchActive` filter. We could go back to `isTouchActive === undefined`, and make sure it never is for mouse/touch paradigm tools. But I don't super love that.

src/base/BaseTool.js
```diff
@@ -49,12 +52,55 @@ export default class BaseTool {
...
+
+  /**
+   * Internal options that "MUST" be in the tool's options if
+   * certain conditions are met. Method is also good for inspecting
+   * options that can be used to change tool behavior.
+   *
+   * @readonly
+   * @memberof BaseTool
+   */
+  get internalOptions () {
```
JamesAPetts: `get internalOptions()` is perhaps a little confusing. One would expect it to return an options object with current keys/values, whereas you get an array of mandatory option names.

src/store/setToolMode.js
```diff
@@ -23,14 +23,7 @@ const setToolActiveForElement = function (
...
-    _setActiveToolWithMatchingMouseButtonMaskToPassive(tool, element, options);
-    _setActiveToolWithIsTouchActiveToFalse(
-      tool,
-      element,
-      options,
-      isTouchActive
-    );
-
+    _resolveInputConflicts(element, tool, options, isTouchActive);
     // TODO: Find active tool w/ active two finger?
```
JamesAPetts: Not too hot on the touch stuff, but this could all be baked into _resolveInputConflicts also I think? Also, if the method gets too large we could consider putting it in a separate file.

src/store/setToolMode.js
```diff
@@ -236,21 +229,41 @@ function setToolMode (mode, changeEvent, toolName, options, isTouchActive) {
...
+ * @param {*} element
+ * @param {*} tool
+ * @param {*} options
+ * @param {*} isTouchActive
+ * @private
+ */
+function _resolveInputConflicts (element, tool, options, isTouchActive) {
+  // There are two API overloads for changing a tool's mode. This helps us
+  // Determine where to find our mask values.
+  const isValueApiOverload = typeof options === 'number';
```
JamesAPetts: This is a much cleaner and easier to follow imo 👍 .

The methods `_resolveMouseInputConflicts` and `_resolveTouchInputConflicts` are easy to parse also.

dannyrb:
Merging this, then circling back to change `isTouchActive` behavior

PR #612, opened 2018-10-01
Play with adding a cursor to `magnify`

dannyrb:
# WIP

Addresses: https://github.com/cornerstonejs/cornerstoneTools/projects/1#card-12189929

_Not ready for review_

## Thoughts

- When multiple active tools have an associated cursor, which one do we show?
    - Does this only really matter for `mouse`?
- Which tools are prime candidates for cursors?
    - Do we have existing SVGs for these?
- What can we add to make setting, updating, removing cursors easier?
- Max cursor size is `128x128`, but `32x32` to be safe
    - Are there any current situations that would force us to use canvas cursors?
- Cursor helpers for basic shapes?
    - Circle + radius + color?
    - Box + rotationAngle + length + width + color?
    - Triangle + rotation Angle + ...?


dannyrb:
@JamesAPetts has picked this up in #637 . Closing this particular PR.

PR #614, opened 2018-10-02
Strategies docs. Added a new setDefaultStrategy method to BaseTool.

JamesAPetts:
Added docs for tool strategies.

@galelis, I'm tagging you for review, as you've implemented a few tools utilizing this mechanism. My description is relatively brief, but I would appreciate if you could check if everything is correct.

I also added a `setDefaultStrategy ()` method to `BaseTool`.


docs/latest/anatomy-of-a-tool/strategies.md
```diff
@@ -1,3 +1,16 @@
 ## Strategies {#strategies}
 
-_This section needs content_
+Tools may optionally have multiple strategies of operation. `BaseTool`, from which all tools inherit, takes two arguments related to the tools strategies:
+
+| Argument | Type | Description |
+|----------|------|-------------|
+| strategies | Object |An `Object` comprised of `function`s that take `evt` as an argument and perform an operation.
```
galelis: As this function is being created outside of the tool class, it also has access as a parameter to `tool.configuration` so it can have access to it values.

docs/latest/anatomy-of-a-tool/strategies.md
```diff
@@ -1,3 +1,16 @@
...
+| defaultStrategy | string | The name of the default strategy to apply. The name should be identical to a property name in `strategies`|
+
+Upon instantiation of the tool, the `activeStrategy` is set to the `defaultStrategy`. If `defaultStrategy` is absent the first function of the `strategies` is used. The strategies mechanism is entirely optional.
+
+`BaseTool` has the function `applyActiveStrategy (evt)` which will execute the `activeStrategy`, and return any value returned by the strategy.
```
galelis: we can call out here that we pass the configuration as well.

PR #619, opened 2018-10-02
Add JSDoc comments to all synchronization files

kevinleedrum:
This just adds JSDoc comments to all the synchronization handlers, Synchronizer methods, etc. in `src/synchronization`.

### DIFF
--- a/src/synchronization/Synchronizer.js
+++ b/src/synchronization/Synchronizer.js
@@ -3,16 +3,27 @@ import external from '../externalModules.js';
 import convertToVector3 from '../util/convertToVector3.js';
 import { clearToolOptionsByElement } from '../toolOptions.js';
 
+/**
+ * Return an array filtered to only its unique members
+ *
+ * @private
+ * @param {Array} array - The array to filter
+ * @returns {Array}
+ */
 function unique (array) {
   return array.filter(function (value, index, self) {
     return self.indexOf(value) === index;
   });
 }
 
-// This object is responsible for synchronizing target elements when an event fires on a source
-// Element
-// @param event can contain more than one event, separated by a space
-// TODO - > ES6 class syntax?
+/**
+ * Synchronize target and source elements when an event fires on the source element
+ *
+ * @param {String} event - The event(s) that will trigger synchronization. Separate multiple
+ * events by a space
+ * @param {Function} handler - The function that will make the necessary changes to the target
+ * element in order to synchronize it with the source element
+ */
 function Synchronizer (event, handler) {
   const cornerstone = external.cornerstone;
   const that = this;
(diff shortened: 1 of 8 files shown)


src/synchronization/Synchronizer.js
```diff
@@ -23,14 +35,26 @@ function Synchronizer (event, handler) {
...
+   * Update the event handler to perform synchronization
+   * @param {Function} handler - The event handler function
+   */
   this.setHandler = function (handler) {
     eventHandler = handler;
   };
 
+  /**
+   * Return a reference to the event handler function
+   * @param {Function} handler - The event handler function
```
JamesAPetts: I believe this is just a copy-paste error, could you delete this `@param` line, as this function has no arguments 👍

src/synchronization/Synchronizer.js
```diff
@@ -247,41 +304,76 @@ function Synchronizer (event, handler) {
...
   this.remove = function (element) {
     that.removeTarget(element);
     that.removeSource(element);
   };
 
-  // Returns the source elements
+  /**
+   * Get the array of source elements
+   *
+   * @returns {Array}
```
JamesAPetts: For arrays please use typed definitions, e.g. here replace `{Array}` with `{HtmlElement[]}`.
This is so it renders correctly in our auto-generated api docs.

Similarly for other types, e.g. a string array is donated by `{string[]}`.

I'd be grateful if you could make this change throughout the PR :).

src/synchronization/panZoomSynchronizer.js
```diff
@@ -1,6 +1,13 @@
 import external from '../externalModules.js';
 
-// This function synchronizes the target zoom and pan to match the source
+/**
+ * Synchronize the target zoom and pan to match the source
+ *
+ * @param {Synchronizer} synchronizer - The Synchronizer instance that attaches this
```
JamesAPetts: If you want to use `{Synchronizer}` throughout instead of `{Object}`, we will also require a [`@typedef`](http://usejsdoc.org/tags-typedef.html) for `Synchronizer`.

Feel free to just use `{Object}` for now, so we can merge this and the definition can made further down the line (in a different PR).

src/synchronization/stackImagePositionOffsetSynchronizer.js
```diff
@@ -3,11 +3,20 @@ import { getToolState } from '../stateManagement/toolState.js';
...
+/**
+ * Calculate a position in space that is offset from the source image's position,
+ * and synchronize the target stack to the image that is closest to that position
+ *
+ * @param {Synchronizer} synchronizer - The Synchronizer instance that attaches this
+ * handler to an event
+ * @param {HTMLElement} sourceElement - The source element
+ * @param {HTMLElement} targetElement - The target element
+ * @param {Object} eventData - The data object from the triggering event
+ * @param {Vector3} positionDifference - An object with { x, y, z } values that will be
```
JamesAPetts: Note: Its fine for this PR, but we should make global definitions for `{Vector2}` and `{Vector3}` in `cornerstoneTools`. Could you replace `{Vector3}` for `{Object}` for now, until we have a definition. Just so that everything in the auto-generated api docs references something.

src/synchronization/Synchronizer.js
kevinleedrum: Oops good catch!

src/synchronization/Synchronizer.js
kevinleedrum: I didn't even know that was a thing in JSDoc, so thanks for the heads up!

PR #625, opened 2018-10-04
Plugin architecture round 2 WIP

JamesAPetts:
WIP - You can now register any common functionality you wish, and build custom tools on top of these. When you register an item it stores it in the import library with a uri comprised with the item's type and its name, e.g. `mixin/myMixin`. This allows users to use implement new common functionality from npm/local libraries into custom third-party tools.

Edit: Rewriting docs now.

## Item Types

The types typically registered by third parties are:
- `base`
- `mixins`
- `manipulators`
- `util`

`cornerstoneTools` has these additional types (third parties could expand these, but are unlikely to):
- `drawing` (for drawing api functions)
- `stateManagement`

Additionally, custom modules can be added to the store.

## Registration

Registration is abstracted to a simple top-level `register` function, taking type, name and item as arguments e.g.:

```
cornerstoneTools.register('module', 'helloWorldModule', myModule);
cornerstoneTools.register('mixin', 'evenMoreHelloWorld', myMixin);
```

## Import

Core and registered functionality can then be retrieved by the top-level `import` function, e.g.:

```
const evenMoreHelloWorldMixin = cornerstoneTools.import('mixins/evenMoreHelloWorld');
```

And store modules may be retrieved from the modules object:

```
const modules = cornerstoneTools.import('store/modules');
const helloWorldModule = modules.helloWorldModule;
```
Or in a single line using a destructuring assingment.
```
const {helloWorldModule} = cornerstoneTools.import('store/modules');
```

## Tools

Third party tools may be easily created as follows:
```
const BaseTool = cornerstoneTools.import('base/BaseTool');

class ExampleTool extends BaseTool {
    constructor (name = 'Example') {
      super({
        name,
        supportedInteractionTypes: ['mouse', 'touch'],
        mixins = [
         `enabledOrDisabledBinaryTool` // A core mixin
         `hevenMoreHelloWorldMixin` // A  third-party registered mixin
        ]
      });

      this._myManipulator = cornerstoneTools.import('manipulators/aRegisteredManipulator');
    }
  }
```

Which could then be packaged into a npm library, or just used locally within the application.
The tool can easily be added to a `cornerstoneTool`s instance like any other tool:

```
cTools.addTool(ExampleTool);

```

To ensure third party tools work correctly. All relevant items should be registered to `cornerstoneTools` prior to instantiation. Then the tools can easily be added to the instance.


dannyrb:
> Netlify: Build exceeded maximum allowed runtime

Well.. That stinks. Not sure why this would be taking > 20 minutes to build though.

src/base/BaseTool.js
```diff
@@ -111,16 +111,35 @@ export default class BaseTool {
...
+
+      const split = mixinUri.split('/');
+
+      if (split.length === 2) {
+        // Fill in ommited type term (mixin).
+        mixinUri = `${split[0]}/mixins/${split[1]}`;
+      } else if (split.length === 1) {
+        // Assume core if no namespace given.
+        mixinUri = `core/mixins/${split[0]}`;
+      }
```
dannyrb: Okay, so we're checking for...

`moduleName/mixins/mixinName` OR
`core/mixins/mixinName`

I'm wondering if it would be more valuable to have all mixins in the same namespace, and enforce no name collisions? If you're importing multiple plugins, they could override other mixins if you're using the simple string. I think it might be easier to just have them vendor prefix the mixins name itself:

- All mixins are at `mixins/mixinName`
- Vendors should adopt the following pattern when creating a mixin name
    - Vendor prefix: `ast` (for Asteris); Mixin name: `alertEverything`
    - When registering `.register('astAlertEverything')`

There's less confusion around what namespace to use, all namespaces are now smaller, and mixin namespace collision due to modules becomes less likely? FWIW, vendor prefixing like that is a popular concept in WebComponents.

src/import.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  switch (depth) {
+    case 1:
+      return lib[namespace];
+    case 2:
+      return lib[namespace][directory];
+    case 3:
+      return lib[namespace][directory][item];
+  }
```
dannyrb: I don't think we need to get this crazy. Most namespaces should be as simple as:

- Type of thing: `mixin`
- Name of the thing: `myMagicalMixin`

And ideally, it should always be passed in as a single string `mixin/myMagicalMixin`

That alone should ensure uniqueness, and is a perfectly fine key for an associative array. We can avoid all of the splitting and just return whatever exists at that location, or a warning if the item doesn't exist. Only need to make this as complex as it needs to be.

src/index.js
```diff
@@ -35,144 +32,23 @@ export { default as ZoomMouseWheelTool } from './tools/ZoomMouseWheelTool.js';
...
-} from './synchronization/stackImagePositionSynchronizer.js';
-export {
-  default as stackImagePositionOffsetSynchronizer
-} from './synchronization/stackImagePositionOffsetSynchronizer.js';
-export {
-  default as stackImageIndexSynchronizer
-} from './synchronization/stackImageIndexSynchronizer.js';
-export {
-  default as panZoomSynchronizer
-} from './synchronization/panZoomSynchronizer.js';
```
dannyrb: I'm not 100% sure I would consider these internal. They feel similar to the "top level tools" that we expose for the quick implementations.

JamesAPetts: No problem moving these back up, I just moved everything down that wasn't a 'tool' as a first stab. I think we need to trawl through the list and discuss which belongs on the top level and which doesn't.

src/index.js
```diff
@@ -35,144 +32,23 @@ export { default as ZoomMouseWheelTool } from './tools/ZoomMouseWheelTool.js';
...
-
-export { default as toolStyle } from './stateManagement/toolStyle.js';
-export {
-  addToolState,
-  getToolState,
-  removeToolState,
-  clearToolState,
-  setElementToolStateManager,
-  getElementToolStateManager
-} from './stateManagement/toolState.js';
```
dannyrb: I think these make sense as internal, but we may want to identify _abstractions_ that should exist to replace functionality like this that we're taking away? restoring/clearing state feels like something you might want to do for a still fairly simple implementation, but I'm not 100% sure these methods are the best way to expose that.

src/index.js
```diff
@@ -35,144 +32,23 @@ export { default as ZoomMouseWheelTool } from './tools/ZoomMouseWheelTool.js';
...
-export { default as moveNewHandle } from './manipulators/moveNewHandle.js';
-export {
-  default as moveNewHandleTouch
-} from './manipulators/moveNewHandleTouch.js';
-export {
-  default as touchMoveAllHandles
-} from './manipulators/touchMoveAllHandles.js';
-export { default as touchMoveHandle } from './manipulators/touchMoveHandle.js';
+// Exposed entities for 3rd party usage/plugins.
+export { default as import } from './import.js';
```
dannyrb: Should we also expose `register`? What if I want to register a single mixin? IE. I have my own codebase, and I'm not a plugin author. I just have a list of mixins I'd like to add?

src/lib.js
```diff
@@ -0,0 +1,217 @@
...
+    getters,
+    modules,
+    setToolMode: {
+      setToolPassiveForElement,
+      setToolActiveForElement,
+      setToolDisabledForElement,
+      setToolEnabledForElement
+    }
+  },
+  core: {
```
dannyrb: I go back and forth on this. I'm not sure if we need to qualify what is an is not core.

src/thirdParty/getPluginList.js
```diff
@@ -0,0 +1,5 @@
+import { lib } from '../lib.js';
+
+export default function () {
+  return Object.getOwnPropertyNames(lib);
```
dannyrb: it might be nice if you could also get all properties prefixed with `mixins` or any other "type"

src/import.js
JamesAPetts: Hmm sure, I just thought it'd be nice to be able to grab a whole plugin object. I felt it could be useful when making a tool that wants to import all of the functionality from a different plugin. I guess the boilerplate isn't too heavy if you have to import each item.

I was just thinking from a DX perspective, as a developer making a new plugin, it'd be nice to import a useful looking plugin, log the object to the console and see whats there. Just wondering if it may be a bit annoying to have to go to the actual plugin source and see what items are present there?

src/base/BaseTool.js
```diff
@@ -111,16 +111,35 @@ export default class BaseTool {
...
-   * @param  {Array} mixinArray An array of mixin names (strings).
+   * @param  {Array} mixinArray An array of mixin URIs (strings).
    */
   _applyMixins (mixins) {
+    // TODO:
+    // TODO: deal with: missing namespace + type
+
+    if (typeof cToolsImport !== 'function') { // Makes testing a lot easier.
+      return;
+    }
```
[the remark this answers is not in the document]
JamesAPetts: It allows you to use `BaseTool` without mocking it, otherwise tools try to instantiate `mixin`s that may not be registered to the library yet. I'd be happy to discuss any ideas you'd have for a better approach, or some jest magic that makes this unnecessary.

src/base/BaseTool.js
JamesAPetts: We can easily prefix based on the plugin/vendor name. My main motivation for this approach is in the other comment.

Perhaps we can scour and build an object upon request if we are clever about the prefixing. Will discuss strategies in a while :+1: :).

src/lib.js
```diff
@@ -0,0 +1,217 @@
...
+      drawLines,
+      drawJoinedLines,
+      drawCircle,
+      drawEllipse,
+      drawRect,
+      fillOutsideRect,
+      drawTextBox,
+      drawArrow,
+      fillBox,
+      fillTextLines
```
dannyrb: Ahhh. So nice to have these namespaces somewhere and off the top level

src/thirdParty/registerItems.js
```diff
@@ -0,0 +1,64 @@
...
+  if (!lib[namespace]) {
+    lib[namespace] = {};
+  }
+}
+
+function buildPathIfNonExistant (namespace, type) {
+  if (!lib[namespace][type]) {
+    lib[namespace][type] = {};
+  }
+}
```
dannyrb: If we switch to string paths, I think this could be a lot simpler. I also wonder if we can just have `register`, expose it top level, and allow people to create their own loop if they want it.

One less API to maintain, and reduced complexity from a versioned plugin object?

src/tools/DoubleTapFitToWindowTool.test.js
```diff
@@ -7,6 +7,10 @@ jest.mock('../externalModules.js', () => ({
   }
 }));
 
+jest.mock('../import.js', () => ({
+  default: jest.fn()
+}));
```
dannyrb: There are ways to do this globally as part of a jest setup, and then override that behavior in an individual test suite if you don't want the default

src/thirdParty/registerPlugin.js
```diff
@@ -1,72 +1,35 @@
...
 }
 
-
-function isPluginNameRegistered (name) {
-  return pluginList.some((registeredPluginName) => {
-    return registeredPluginName === name;
-  });
+function isPluginNameRegistered (namespace) {
+  return lib[namespace] !== undefined;
 }
```
dannyrb: I have mixed feelings about this abstraction. I understand why and the convenience, I also think it will be much more brittle than only exposing import/register/registerModule

If you really like this, I can compromise, but I'd like to haggle with you on the schema for this

src/index.js
JamesAPetts: Currently these haven't been removed, just moved to the import interface. I'll leave them 'as is' for now, but perhaps we should abstract them further down the road 👍 .

examples/thirdParty/index.html
```diff
@@ -142,38 +140,13 @@
...
-        tool: HelloWorldMouseTool,
-        name: 'HelloWorldMouseTool'
-      }
-    ]
-  };
-
-  //====== END - Define a simple plugin ======//
+  // Register the thirdParty functionality
+  cornerstoneTools.register('module', 'helloWorldModule', myModule);
+  cornerstoneTools.register('mixin', 'evenMoreHelloWorld', myMixin);
```
dannyrb: So much less code. I love it

examples/thirdParty/index.html
```diff
@@ -222,17 +195,11 @@
...
-  // Add the tool
   const toolName = toolBtn.getAttribute('data-tool');
-  // Add the 3rd party tool:
-  const apiTool = cTools.thirdParty.tools.HelloWorldMouseTool;
-
-  console.log(apiTool);
+  // Add the 3rd party tool (may be locally defined, or from an npm library)
 
-  cTools.addTool(apiTool);
+  cTools.addTool(HelloWorldMouseTool);
```
dannyrb: Then just a simple addTool. Ahhhhh beautiful

src/import.js
```diff
@@ -0,0 +1,44 @@
+import { lib } from './lib.js';
+
+/**
+ * Imports functionality from cornerstoneTools for use in external packages/plugins.
+ *
+ * @param  {string} uri the import path for the entity to import.
+ * @return {Class|Object|Function} The entity requested.
+ */
+export default function (uri) {
+  return lib[uri];
```
dannyrb: 🔥 🔥 🔥

src/init.js
```diff
@@ -52,7 +51,6 @@ export default function (configuration) {
     removeToolForElement,
     setToolOptions,
     setToolOptionsForElement,
-    thirdParty,
```
dannyrb: I'm happy to see this gone 👍

src/thirdParty/registerMixin.js
```diff
@@ -1,25 +1,33 @@
...
+ * Register an mixin to cornerstoneTools.
  *
- * @param  {Object} newMixin The mixin to register.
  * @param {string} name The name of the mixin.
+ * @param {Object} mixin The mixin itself.
+ * @param {boolean} [overwrite=false] Whether an mixin should be overwritten,
+ *                                    should it have the same name.
  */
-export default function (newMixin, name) {
+export default function (name, mixin, overwrite = false) {
```
dannyrb: If the logic for `registerX` is similar, you can make a single function, and then spin off a version per type using `.bind()`?

- Less code
- Changes update each type
- Less to unit test

dannyrb: RE: `setToolMode`

dannyrb:
Much, much cleaner @JamesAPetts. I'm loving it 👍

JamesAPetts:
Hey @dannyrb , going to merge this for now, the `registerX` issue we can work on nearer the time, I just want to keep the vNext pushing a little bit so we don't get more of the 3-PR's-then-everything-breaks issues. If you have anything else you'd like to change, you know where to DM me!

PR #635, opened 2018-10-11
WIP: Bidirectional Tool 

galelis:
WIP: Tool is still in development, this is just to be able to test on netlify


JamesAPetts:
Just chiming into say this is looking really nice!

src/tools/bidirectionalTool/renderToolData.js
```diff
@@ -7,8 +7,8 @@ import updatePerpendicularLineHandles from './updatePerpendicularLineHandles.js'
 import toolStyle from '../../stateManagement/toolStyle.js';
 import toolColors from '../../stateManagement/toolColors.js';
 import { getToolState } from '../../stateManagement/toolState.js';
-import { getNewContext, draw, setShadow, drawLine } from '../../util/drawing.js';
-import drawLinkedTextBox from '../../util/drawLinkedTextBox.js';
+import { getNewContext, draw, setShadow, drawLine } from '../../drawing/index.js';
+import drawLinkedTextBox from '../../drawing/drawLinkedTextBox.js';
```
JamesAPetts: Sorry for all the shifts, nearing the end of pre-release! This should be the last major move before v3 :). Keep up the good work man!

JamesAPetts:
Hey, just some feedback, there is a small bug introduced in a recent commit. Clicking on a handle with the bidirectional tool active always starts a new annotation (i.e. there is no passive functionality). All in all looking good though, keep up the good work!

galelis:
@JamesAPetts hmmm Good catch mate! 
I was always testing it with only bidirectional. I was able to reproduce what you just described, I will fix that. 
Thanks for being proactive in this PR :) its helping a lot

JamesAPetts:
Hey @galelis , I fixed the issue we were having. We are going to merge this now, so we can get on with some testing and @swederik erik can use it in CCC.

Feel free to open another PR if there are outstanding issues! :)

PR #637, opened 2018-10-15
Feature/svg cursors

JamesAPetts:
Following up on @dannyrb's #612 PR (thanks for getting the ball rolling!):

Implements the following properties to `BaseTool`:

## svgCursor

If `svgCursor` is set, the mouse cursor will be changed to the set value when the tool is `active` and `mouseButtonMask === 1`. I.e. the cursor for the primary mouse click is shown. Touch tools are unaffected.

SVG cursors are stored in `assets/svgCursors.js`, you can alter their `mousePoint` property to change the mouse position to either the top left (e.g. magnify) or center (e.g. pan) of the icon.

## hasCanvasCursor

If `hasCanvasCursor` is true, the canvas will re-render on `MOUSE_MOVE` whilst the tool is active. This is intended for tools that need to render a cursor on the canvas, if the cursors could potentially be greater than 128x128 pixels in size. Canvas cursors should be used sparingly, and SVG cursors are preferable wherever possible.

### hideDefaultCursor

If `hideDefaultCursor` is true and `svgCursor` is unset, the default cursor will be hidden when hovering over the cornerstone enabled element. You may want to do this if using canvas cursors, but as canvas cursors can be complex and dynamic, this is optional. See `FreehandMouseSculpterTool` to see an example of this being switched on/off during tool usage.

As an example I've implemented svg cursors for `MagnifyTool`, `RotateTool` and `PanTool`; as well as canvas cursors for `BrushTool` and `FreehandMouseSculpterTool`. Tools without custom cursors use the default mouse cursor.


dannyrb:
As this builds on mine, I'll close #612. Tagged myself to review this. I'll try to give it a look before Monday and draft some good follow-up questions.

JamesAPetts:
@dannyrb , sounds good, note this is probably quite outdated by now. Happy to discuss/scrap/rewrite anything, this was very much a playground at the time 👍 .

dannyrb:
@JamesAPetts what's left for us to discuss on this one? It's been a while, haha.

JamesAPetts:
Hey @dannyrb , not much. Iirc we didn't implement it purely because we don't have actual cursors for all tools, and it doesn't make sense to release it until its functional for all core tools. So more of an art issue.

PR #641, opened 2018-10-17
Prevent circular import dependency

kofifus:
import toolStyle and textStyle from where they are defined instead of index.js


dannyrb:
Thanks for this :+1:

Issue #643, opened 2018-10-17
`mouseWheelInput.js`: Two finger scrolling on Firefox scrolls the whole page


talkdirty:
### Prerequisites

* [x] Are you running the latest version?
* [x] Are you reporting to the correct repository?
* [x] Did you perform a cursory search?

### Description

Hi!

We noticed a peculiar thing specific to Firefox related to `mouseWheelInput.js`. Two finger scrolling (on a Laptop with a touchpad) will scroll the page and images simultaneously.
We found the potential culprit of this, which is the legacy `DOMMouseScroll` event used by `mouseWheelInput.js`. More recent versions of Firefox don't need the Firefox specific `DOMMouseScroll` anymore, and accept the generic `wheel` event. We tried replacing `DOMMouseScroll` with `wheel` (i.e., setting `const mouseWheelEvents = ['wheel']`) in `mouseWheelInput.js`, and it works perfectly with this.

I'd be happy to open a PR for this if anyone can confirm the issue. [Here is the documentation](https://developer.mozilla.org/en-US/docs/Web/Events/wheel) of wheel on MDN, along with a browser compatibility table.

Thanks a lot for this awesome project!
Best Regards

### Steps to Reproduce

1. Latest Firefox 64.0b1; Laptop with a Touchpad and two finger scrolling
2. Cornerstone viewer in a scrollable context, using `mouseWheelInput.js` for scrolling through images
3. Use two-finger scrolling on the cornerstone viewer

**Expected behavior:** Page should not scroll

**Actual behavior:** Page and images viewed in cornerstone scroll simultaneously

### Codepen With Reproduction of Issue:

- [Based on scroll example](https://codepen.io/talkdirty/pen/qJxQap)
Make the output window small enough so it scrolls, then try two finger scrolling on firefox. The whole page will scroll.

dannyrb:
@talkdirty thanks for the reproduction and detailed issue!

Your proposed fix makes sense to me. Feel free to create a PR addressing the issue and tag me as the reviewer (or `@mention` me), and I will be happy to merge the fix  🎉

talkdirty:
Thanks for the quick response! See the referenced PR. @swederik might probably weigh into this, as he appears to be the original author of the functionality. Note that I could only test this on recent Edge, Firefox, and Chrome. I don't know how the behavior is on previous versions and IE. Also, I think the file could now be refactored a bit (i.e., it should be retested if the hack is still relevant)

[...]

PR #644, opened 2018-10-17
Use wheel event only in mouseWheelInput.js

talkdirty:
Previously, both mousewheel and DOMMouseScroll Events were used, the
latter of which is firefox specific. DOMMouseScroll shows problems when
two finger scrolling on Laptops. Newer browser versions all use the
standardized wheel event now, which should be the right thing to use
here.

See [this issue](https://github.com/cornerstonejs/cornerstoneTools/issues/643)

### DIFF
--- a/src/inputSources/mouseWheelInput.js
+++ b/src/inputSources/mouseWheelInput.js
@@ -23,7 +23,7 @@ function mouseWheel (e) {
   if (e.type === 'DOMMouseScroll' && e.axis === 1) {
     return;
   }
-
+  
   e.preventDefault();
 
   let x;
@@ -70,7 +70,7 @@ function mouseWheel (e) {
   triggerEvent(element, EVENTS.MOUSE_WHEEL, mouseWheelData);
 }
 
-const mouseWheelEvents = ['mousewheel', 'DOMMouseScroll'];
+const mouseWheelEvents = ['wheel'];
 
 function enable (element) {
   // Prevent handlers from being attached multiple times


talkdirty:
@dannyrb

dannyrb:
@talkdirty, I found a good resource for which versions of which browser vendors use which scroll events. It was included with a snippet of code to help normalize the event handling across browsers. I'll revisit this PR now that I have that information. I'll likely update our listener to normalize the event and well :+1:

talkdirty:
That's great, thank you! Out of interest, could you link the resource you
found? Also let me know if you require any assistance for this.
Georg

dannyrb:
https://github.com/facebookarchive/fixed-data-table/blob/master/src/vendor_upstream/dom/normalizeWheel.js

dannyrb:
@talkdirty, alright. I've merged the fix for this to `vNext`. We have the option of backporting this to `v2`, but we would lose some browser support because I dropped listening for `'mousewheel', 'DOMMouseScroll'` and instead only listen to `wheel`. I'm okay with doing that in `vNext`, because it's the next major release, and I'm okay with the breaking change there.

Whether we should do this for `v2` isn't as cut and dry, even though I really, really hope no one is using `cornerstone` on `<= IE 9` or older versions of evergreen browsers. I would be willing to backport it if @swederik gave the 👍

I believe #643 can be addressed with a non-breaking change, but I don't have the bandwidth to look into it.


*tl;dr:*

- I will look at and merge any non-breaking solutions
- If @swederik gives a 👍, I would spend the time to merge the `vNext` solution to this branch
- I recommend waiting for `vNext` RC or bravely porting existing code to it now if this issue is plaguing you

I'll close this PR for now; feel free to continue discussion here, or on the related issue

Issue #645, opened 2018-10-19
calculations using rowPixelSpacing and colPixelSpacing when they are null


kofifus:
since commit https://github.com/cornerstonejs/cornerstoneWADOImageLoader/commit/da4c03933b4b6f634c59614a6aa1a479980f2968#diff-0e1db73251040e841d0e0183c5ebdf55

rowPixelSpacing and colPixelSpacing default to null

however many tools (ie angletool) use calculations including these without checking for null values

ie in https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/imageTools/angleTool.js#L100-L103

we have:

const dx1 = (Math.ceil(data.handles.start.x) - Math.ceil(data.handles.end.x)) * eventData.image.columnPixelSpacing;
const dy1 = (Math.ceil(data.handles.start.y) - Math.ceil(data.handles.end.y)) * eventData.image.rowPixelSpacing;
const dx2 = (Math.ceil(data.handles.start2.x) - Math.ceil(data.handles.end2.x)) * eventData.image.columnPixelSpacing;
const dy2 = (Math.ceil(data.handles.start2.y) - Math.ceil(data.handles.end2.y)) * eventData.image.rowPixelSpacing;

similar problems exist in pan.js, pointProjector.js etc

it seems some tools solve this by defaulting these values to 1.

This should be fixed.

Also I thought  angle measurements you do not need a pixel spacing, it can be done purely on geometry.

PR #647, opened 2018-10-19
bugfix defaulting rowPixelSpacing and colPixelSpacing to 1

kofifus:
fixes 
https://github.com/cornerstonejs/cornerstoneTools/issues/645

make sure rowPixelSpacing and colPixelSpacing default to 1 if null before being used in divisions etc ..


dannyrb:
I think this is fine, because the modified tools aren't reporting a measurement in a specific unit. I'm going to merge this for now.

@JamesAPetts, we'll likely want to port these changes to `vNext`

PR #651, opened 2018-10-22
Expose the requestPoolManager.

JamesAPetts:
Fixes a bug in which `requestPoolManager` was being imported to index, but not exported.


JamesAPetts:
Nuking this, starting a new branch removing the cTools instance stuff for now. Support for multiple instances of `cornerstoneTools` is being postponed until a future major version.

PR #652, opened 2018-10-22
Removed the idea of a 'cornerstoneTools' instance for v3.

JamesAPetts:
The instance made integration in applications a little clunky at this stage. Fully encapsulated instanced could be useful in the future, but the half-half scenario we've had in vNext has been axed.


JamesAPetts:
@galelis , could you check that all the functionality you think OHIF would need is exposed at `src/index.js`?

PR #658, opened 2018-10-24
feature/Smarter global add tools

dannyrb:
## Work in progress

Ideally, for our simpler setups, this should reduce or remove the need for a tool manager entirely.

### Goal

When you enable a new element, it's added tools and state should reflect that of globally added and state changed tools.

- This should be configurable. Some users may want a "blank enabledElement" every time one is created.

### Two options...

1. Track all global state changes, resolve conflicts for those changes, and use those as the "defaults" for when we enable a new element
2. Add the ability to "duplicate" the enabled/active tools from one enabledElement to the other
    - Least complex

The tricky parts here are:
- If you call `setToolActiveForElement` you can throw things out of whack. It makes it difficult to blanket copy options/configuration/active interaction types.
    - Maybe if a `forElement` call is made, we disable this restoration behavior?
- Conflict resolver logic takes place in the `ForElement` portion of the internal code. This makes it difficult to reuse and reliably restore the individual interaction type's state per tool.

### DIFF
--- a/src/base/BaseTool.js
+++ b/src/base/BaseTool.js
@@ -99,7 +99,7 @@ export default class BaseTool {
     return internalOptions;
   }
 
-  setDefaultStrategy() {
+  setDefaultStrategy () {
     this.activeStrategy = this.defaultStrategy;
   }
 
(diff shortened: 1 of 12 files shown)


JamesAPetts:
Let me know when this in a state where you'd like it reviewed 😄 👍 .

dannyrb:
This still needs a better implementation for repeating history, but it should do the trick for the very short term.

PR #667, opened 2018-10-26
fix(tool-naming): Pass configuration.name into tool constructor

evren217:


dannyrb:
Hi @evren217,

I'm wondering if we should update our tools to merge configuration instead of taking just `name`? I'm butting up some use cases where it would be nice to specify the strategy, number of touchPointers, etc. At instantiation instead of after the tool has been added for each enabledElement

galelis:
I agree with @dannyrb about getting configurations merged and also name added into the name added into the correctly property as its not a configuration. 

We got this issue when we were trying to set a name into the tools for the viewer. 

Do you think we can get this merged and I can create an issue in the board so we can work in the config merge later?

dannyrb:
Sounds good to me, @galelis

PR #673, opened 2018-10-29
Fix: Calling `init` multiple times should not leave lingering listeners

dannyrb:
This PR patches the `init` method

- Make sure we remove any lingering listeners before adding new ones
- Update `windowResizeHandler`'s `disable` method to remove the event listener (previously added)


dannyrb:
CC: @swederik

PR #681, opened 2018-10-31
Feature: Expand supported schema for tooldata's handles

dannyrb:
- Before: Tool handles had to be an object of named handles
- After: Tool handles can be an object of named handles, an array of handles, or an object of named handles and named arrays that contain handles


dannyrb:
@JamesAPetts this is a bit reckless -- If it breaks anything for you, ping me and I'll add tests and patch it up.

Issue #693, opened 2018-11-07
Throttle Probe Tool Calculations


hsahin:
### Description

Adding more than ~5 probes makes the canvas unresponsive.
This does not happen when adding more than 20 freehand object (containing more than 100 points).

## Steps to Reproduce

1. Add 5 probe points.
2. Panning and zooming becomes slow.

**Expected behavior:**

Have a smooth experience.

**Actual behavior:**
Slow and sluggish canvas tool.

### Codepen With Reproduction of Issue:

cornerstone.enable(element);
cornerstoneTools.mouseInput.enable(element);

cornerstone.loadImage(imageId).then(function(image) {
    cornerstone.displayImage(element, image);
    cornerstoneTools.probe.activate(element, 1);
    cornerstoneTools.probe.enable(element);
}

dannyrb:
This is likely do to how the tool determines when it should recalculate values. Newer methods throttle the calculation, and/or only update the value if the previous value becomes invalid.

More than happy to review a PR for this fix. The vNext branch would likely need it as well

hsahin:
@dannyrb Could you review #698?

dannyrb:
@hsahin , I believe this is a common problem for any tool that looks at pixel values. I've updated the issue and added labels to address the core issue here. While I will look at a fix for the `master` branch, I would prefer to see the work done in `vNext` to reduce the effort required to port changes.

dannyrb:
@evren217 was this addressed in your recent changes? I think we're throttling all annotation tool recalculations now.

evren217:
That is correct. We can close this issue.

PR #698, opened 2018-11-08
Feature to disable probe textbox by setting the configuration. 

hsahin:
This PR hotfixes issue #693. It does not solve the underlying problem which is the call to getRGBPixels in the draw function of the probe tool(After investigating a while with firefox perfomance tool).

The current behavior makes the canvas too unresponsive if the amount of probes is larger than ~5 (See #693.) 

The new behavior is that the text box is not drawn and that makes the page more responsive if the textbox is not desired.

Users do not have to make changes with the current implementation. If the want to remove the textbox it is possible by setting the configuration object in their application. 
e.g.

const config = { disableTextBox: true};

cornerstoneTools.probe.setConfiguration(config);

This disables the rendering of the textbox.

* **Other information**:
It would be a nice addition to add this configuration to all tools which use a textbox.

### DIFF
--- a/src/imageTools/probe.js
+++ b/src/imageTools/probe.js
@@ -61,6 +61,7 @@ function onImageRendered (e) {
   const context = getNewContext(eventData.canvasContext.canvas);
 
   const fontHeight = textStyle.getFontSize();
+  const config = probe.getConfiguration();
 
   for (let i = 0; i < toolData.data.length; i++) {
     const data = toolData.data[i];
@@ -75,6 +76,8 @@ function onImageRendered (e) {
 
       // Draw the handles
       drawHandles(context, eventData, data.handles, color);
+
+      if (config.disableTextBox) return;
 
       const x = Math.round(data.handles.end.x);
       const y = Math.round(data.handles.end.y);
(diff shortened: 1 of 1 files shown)


dannyrb:
@hsahin, if there is no text box, what value is this tool providing?

dannyrb:
@hsahin that makes sense. Thank you for providing an example.
I will merge this into v2, but I can't promise it will be immediately available in the upcoming v3 release. I'd like to revisit this behavior to see if there is a better way to implement it in the context of all tools and the global tool settings.

src/imageTools/probe.js
```diff
@@ -76,6 +77,8 @@ function onImageRendered (e) {
       // Draw the handles
       drawHandles(context, eventData, data.handles, color);
 
+      if (config.disableTextBox) return;
+
```
dannyrb: In v2, I don't remember if it is a given that `config` will be defined. You may want to updated this to `config && config.disableTextBox`

dannyrb:
Thanks, @hsahin! I hope to see you around ^_^
Don't be afraid to start discussions around enhancements or to report bugs. For a sneak preview of what's slated next for cornerstone tools, be sure to check out the `vNext` branch

hsahin:
@dannyrb No, thank you! I think vNext will be amazing!

PR #728, opened 2018-11-26
Chore: remove karma, fix security warnings raised by npm audit

dannyrb:
We had 18+ `npm audit` security warnings. 3 critical, most low, in our `devDependencies`. This PR resolves those issues, and removes a large number of unused dependencies we've accumulated over time. It also updates the three tests we had in `karma/mocha/chai` to `jest`


dannyrb:
👋 Not explicitly requesting a review for this one @swederik and @JamesAPetts, but it may be worth double checking I didn't delete any dependencies that are critical for your workflows.

PR #715, opened 2018-11-19
fix(src/magnify.js): fix magnify use cornerstone.enable bug whith use…

xingbofeng:
… webgl renderer

* **Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Bug fix

* **What is the current behavior?** (You can also link to an open issue here)
when use webgl renderer, previous , magnify can render image with wrong wwwc, now is correct.

* **What is the new behavior (if this is a feature change)?**
when use webgl renderer, previous , magnify can render image with wrong wwwc, now is correct.

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
NO

* **Other information**:

PR #730, opened 2018-11-26
fix(src/tools/MagnifyTool.js): fix magnify use cornerstone.enable bug…

xingbofeng:
fix this [bug](https://github.com/cornerstonejs/cornerstoneTools/pull/715)


dannyrb:
Thanks, @xingbofeng!

Issue #703, opened 2018-11-12
Store the angle inside the tool for outside access


algates90:
### Prerequisites

* [X ] Are you running the latest version?
* [X] Are you reporting to the correct repository?
* [ ] Did you perform a cursory search?

For more information, see the `CONTRIBUTING` guide.

### Description

cornerstoneTools.getToolState(element,'length') -- returns the calculated length value 

![image](https://user-images.githubusercontent.com/30894700/48352905-52945f00-e68e-11e8-976e-502217e851bc.png)

But the same doesnt work for 'angle' -- Does not return the calculated value.

It would be great, if we could add this. Its just one line change needed, that I have done locally now. But would be good if its added to a release.

![image](https://user-images.githubusercontent.com/30894700/48353002-971ffa80-e68e-11e8-8423-51348caf6b2f.png)

Solution : Inside the imageRendered function for angle.

var rAngle = (0, _roundToDecimal2.default)(_angle, 2);
// Store the angle inside the tool for outside access
data.angle = rAngle;

dannyrb:
This issue was originally for v2. In v3, we store the angle in toolState via `cachedStats`

PR #744, opened 2018-12-04
Store the angle value into angle toolState

rodrigolabs:

The value `rAngle` was added to angle toolState.

Linked issue: https://github.com/cornerstonejs/cornerstoneTools/issues/703

> cornerstoneTools.getToolState(element,'length') -- returns the calculated length value
>
> image
> 
> But the same doesnt work for 'angle' -- Does not return the calculated value.
> 
> It would be great, if we could add this. Its just one line change needed, that I have done locally now. But would be good if its added to a release.
> 
> image

`cornerstoneTools.getToolState(element, 'angle')` will return the `rAngle` value.

screen shot 2018-12-03 at 9 42 10 pm

No.

* **Other information**:

It's already implemented on `vNext` branch:

https://github.com/cornerstonejs/cornerstoneTools/blob/90dcd97a98e95669df315c83a6966acc26ba9586/src/tools/annotation/AngleTool.js#L192

---

@algates90 FYI

### DIFF
--- a/src/imageTools/angleTool.js
+++ b/src/imageTools/angleTool.js
@@ -109,6 +109,9 @@ function onImageRendered (e) {
       angle *= (180 / Math.PI);
 
       const rAngle = roundToDecimal(angle, 2);
+
+      data.rAngle = rAngle;
+
       const str = '00B0'; // Degrees symbol
       const text = rAngle.toString() + String.fromCharCode(parseInt(str, 16));
 


JamesAPetts:
Hey @rodrigolabs,

I like the idea of storing the angle in the tool data, but just a thought: wouldn't it be better to calculate and store the angle whenever the annotation is created/modified? Then in `onImageRendered` we could just grab the value and render it to the canvas, rather than recalculating it every frame?

I wouldn't want to block this PR, as it is replicating what is in vNext, but just as more of a permanent solution?

Cheers,
James

rodrigolabs:
Hi @JamesAPetts,

I was beginning to make this change when I realized that if we calculate this value only after an angle is created, we'll lose the reference on this text:

giphy

What's your opinion on this case?

JamesAPetts:
Hmm, good point, on my phone atm, so not in a good place to check through the source, correct me if I'm wrong, but surely we could update it on drag also? I.e. whenever the node changes position?

dannyrb:

FYI, if you check out the latest `RectangleRoi`, you can see how we're caching the tool data from calculation there. We also throttle those same calculations. I'd be curious to hear your thoughts.

PR #747, opened 2018-12-05
Fix BaseTool merge configurations

### LINKED ISSUE #738: Default tool configuration is overwritten by custom configuration passed to constructor
I think we should merge default and custom configuration objects.
See: https://github.com/cornerstonejs/cornerstoneTools/blob/vNext/src/tools/ReferenceLinesTool.js#L31

galelis:
# Changes
- Move merge into `baseTool.js`
- Fix `BaseBrushTool.js` after changes on `BaseTool.js`
- Add some unit tests for BaseTools and deepmerge

- Update all tools to start using this approach

(Closes #738)

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "gitbook-plugin-sitemap": "1.2.0",
     "jest": "23.4.0",
     "jest-junit": "5.1.0",
-    "lodash": "4.17.11",
+    "lodash.merge": "^4.6.1",
     "nodemon": "1.18.6",
     "semantic-release": "15.13.4",
     "shx": "^0.3.2",
--- a/package.json
+++ b/package.json
@@ -64,7 +64,6 @@
     "gitbook-plugin-sitemap": "1.2.0",
     "jest": "23.4.0",
     "jest-junit": "5.1.0",
-    "lodash.merge": "^4.6.1",
     "nodemon": "1.18.6",
     "semantic-release": "15.13.4",
     "shx": "^0.3.2",
@@ -75,8 +74,9 @@
   },
   "dependencies": {
     "@babel/runtime": "7.1.2",
-    "cornerstone-math": "0.1.7",
-    "debug": "4.1.1"
+    "debug": "4.1.1",
+    "lodash.merge": "^4.6.1",
+    "cornerstone-math": "^0.1.7"
   },
   "files": [
     "dist/cornerstoneTools.js",
--- a/package.json
+++ b/package.json
@@ -75,7 +75,6 @@
   "dependencies": {
     "@babel/runtime": "7.1.2",
     "debug": "4.1.1",
-    "lodash.merge": "^4.6.1",
     "cornerstone-math": "^0.1.7"
   },
   "files": [
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "gitbook-plugin-sitemap": "1.2.0",
     "jest": "23.4.0",
     "jest-junit": "5.1.0",
+    "lodash": "4.17.11",
     "nodemon": "1.18.6",
     "semantic-release": "15.13.4",
     "shx": "^0.3.2",
@@ -74,8 +75,8 @@
   },
   "dependencies": {
     "@babel/runtime": "7.1.2",
-    "debug": "4.1.1",
-    "cornerstone-math": "^0.1.7"
+    "cornerstone-math": "0.1.7",
+    "debug": "4.1.1"
   },
   "files": [
     "dist/cornerstoneTools.js",
--- a/src/tools/annotation/ArrowAnnotateTool.js
+++ b/src/tools/annotation/ArrowAnnotateTool.js
@@ -28,8 +28,8 @@ import { arrowAnnotateCursor } from '../cursors/index.js';
  * @extends Tools.Base.BaseAnnotationTool
  */
 export default class ArrowAnnotateTool extends BaseAnnotationTool {
-  constructor(configuration = {}) {
-    const defaultConfig = {
+  constructor(props = {}) {
+    const defaultProps = {
       name: 'ArrowAnnotate',
       supportedInteractionTypes: ['Mouse', 'Touch'],
       configuration: {
@@ -41,11 +41,8 @@ export default class ArrowAnnotateTool extends BaseAnnotationTool {
       },
       svgCursor: arrowAnnotateCursor,
     };
-    const initialConfiguration = Object.assign(defaultConfig, configuration);
 
-    super(initialConfiguration);
-
-    this.initialConfiguration = initialConfiguration;
+    super(props, defaultProps);
     this.preventNewMeasurement = false;
   }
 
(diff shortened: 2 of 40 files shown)


src/tools/base/BaseTool.js
```diff
@@ -1,4 +1,5 @@
 import mixins from './../../mixins/index.js';
+import lodash from 'lodash';
```
dannyrb: Is there any way we can depend on _only_ `lodash`'s `merge` method instead of all of `lodash`?

dannyrb: Bonus points if we can pull in just this utility method's source to reduce the number of external dependencies we have.

src/tools/annotation/ArrowAnnotateTool.js
```diff
@@ -27,8 +27,8 @@ import { textBoxWidth } from './../../drawing/drawTextBox.js';
  * @extends Tools.Base.BaseAnnotationTool
  */
 export default class ArrowAnnotateTool extends BaseAnnotationTool {
-  constructor(configuration = {}) {
-    const defaultConfig = {
+  constructor(props = {}) {
+    const defaultProps = {
```
dannyrb: This naming is better, but we may want to use something a bit more verbose to reduce confusion. I'll noodle on it.

src/tools/base/BaseBrushTool.js
```diff
@@ -14,22 +14,8 @@ const { state, setters } = store.modules.brush;
...
-    super({
-      name,
-      strategies,
-      defaultStrategy,
-      configuration,
-      supportedInteractionTypes,
-      mixins,
-    });
+  constructor(...args) {
+    super(...args);
```
dannyrb: You added these lines here, but not to `BaseAnnotationTool`?

src/tools/base/BaseTool.js
```diff
@@ -17,20 +18,28 @@ import mixins from './../../mixins/index.js';
...
+    props = lodash.merge({}, defaultProps, props);
+
+    const {
+      name,
+      strategies,
+      defaultStrategy,
+      configuration,
+      supportedInteractionTypes,
+      mixins,
+    } = props;
```
dannyrb: I'm having a hard time remembering. Does this set the object's properties because the variables were instantiated here? Or do we need explicitly use `this.`? For clarity's sake, we may want to use `this`

src/tools/base/BaseTool.js
galelis: Good point indeed! But I thought about that too. 

We have lodash to perform a merge function for webpack config. 
We could easily change from the whole lodash to only cloneDeep, merge and mergeWith. 
But some of our npm packages also uses the whole lodash module, so dividing it to lodash.clonedeep/merge/mergewith will actually increase the size of our package.. at least the map increased in my tests.

src/tools/annotation/ArrowAnnotateTool.js
galelis: I will be happy to find out a better name :)

src/tools/base/BaseBrushTool.js
galelis: So, I can add it there too.. but as we were not doing anything else in the constructor off `BaseAnnotationTool`, it may be useless.

src/tools/base/BaseTool.js
galelis: It does not. and also we do it one by one because we do some checks, as: 
- if we don't have any `defaultStrategy` we set the first. 
- if we don't have `strategies` we set as an empty array
...

src/tools/base/BaseTool.test.js
```diff
@@ -16,31 +16,118 @@ jest.mock('./../../import.js', () => ({
 }));
 
 describe('BaseTool.js', () => {
-  describe('default values', () => {
-    it('has a default mode of "disabled"', () => {
+  describe('constructor()', () => {
+    it('should be instantiate with default mode of "disabled"', () => {
```
JamesAPetts: `instantiate` -> `instantiated`

package.json
```diff
@@ -56,13 +56,14 @@
     "gitbook-plugin-sitemap": "1.2.0",
     "jest": "23.4.0",
     "jest-junit": "5.1.0",
-    "lodash": "4.17.10",
+    "lodash.merge": "^4.6.1",
```
dannyrb: We've marked this as a `devDependency`, but because we `import` it into `BaseTool.js`, it should be under `dependencies`

dannyrb:
You're my hero, @galelis! How do you feel about `deepMerge`?
If you resolve the one conflict, I would be more than happy to pull the `merge` lever.

galelis:
> You're my hero, @galelis! How do you feel about `deepMerge`?
> If you resolve the one conflict, I would be more than happy to pull the `merge` lever.

I am happy with it. It is doing the merge as I was expecting. 
I tested on my OHIF, worked well with bidirectional and arrowAnnotate

I will fix the conflict and change all other tools to use it.

src/tools/base/BaseTool.js
galelis: Fixed this issue.

src/tools/base/BaseBrushTool.js
galelis: I came back with the constructor.

src/tools/base/BaseTool.test.js
galelis: I fixed the typo

package.json
galelis: I have removed both dependencies.

src/tools/base/BaseTool.js
```diff
@@ -1,4 +1,5 @@
 import mixins from './../../mixins/index.js';
+import merge from 'lodash.merge';
```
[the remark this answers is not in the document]
galelis: I agree with you Danny, thanks for you insights. 
I have moved the merge function into a base js file into our code, avoiding the use of a npm module only for that.

galelis:
Hey @dannyrb,
I have fixed the conflicts, and change all remaining tools.

dannyrb:
@galelis, I realize this is waiting on me. I have 👀 on it. Sorry for the delay.

package.json
```diff
@@ -75,8 +74,8 @@
   },
   "dependencies": {
     "@babel/runtime": "7.1.2",
-    "cornerstone-math": "0.1.7",
-    "debug": "4.1.1"
+    "debug": "4.1.1",
+    "cornerstone-math": "^0.1.7"
```
dannyrb: Can we switch this back to an exact version?

dannyrb:
Does this account for the new CircleTool?

dannyrb:
@galelis, this looks how I remember it. AFAIK, this doesn't break anything (if we've accounted for the circle tool). We can merge, and address any breaks in a patch. This should get us to the expected behavior ^_^

Thanks for all of your work on this.

galelis:
@dannyrb Just updated the CircleRoi with latest changes to get props merged as the others :) 

@evren217 and Danny, seems that its working, have you guys checked? 



NOTEs: Crosshair and DoubleTapToFit is not working (I think) but its also not working on other netlify instances, I can chack and fix in another PR)

dannyrb:
DoubleTapToFit works for me on my phone. I'm going to merge this. I'll try to be extra responsive this next week to any issues that may have been caused by this with patch updates -- hopefully we're in the clear ^_^

PR #756, opened 2018-12-19
Freehand tool improvements

drozdzynski:
Freehand tool feature

**What is the new behavior (if this is a feature change)?**
Now you can set configuration for freehand tool:
- do not force close roi while drawing (`alwaysClose: false`)
- set pencil as default mode (`defaultPencil: true`)

**Does this PR introduce a breaking change?** 
No breaking changes

**Features for cornerstoneTools v2 because brush tool not working correctly with this version.**

### DIFF
--- a/src/imageTools/freehand.js
+++ b/src/imageTools/freehand.js
@@ -49,8 +49,8 @@ let configuration = {
   movingTextBox: false,
   currentHandle: 0,
   currentTool: -1,
-  defaultPencil: true,
-  alwaysClose: false
+  defaultPencil: false,
+  alwaysClose: true
 };
 
 /**
--- a/src/imageTools/freehand.js
+++ b/src/imageTools/freehand.js
@@ -48,7 +48,9 @@ let configuration = {
   modifying: false,
   movingTextBox: false,
   currentHandle: 0,
-  currentTool: -1
+  currentTool: -1,
+  defaultPencil: true,
+  alwaysClose: false
 };
 
 /**
@@ -178,7 +180,9 @@ function mouseDownActivateCallback (e) {
     return;
   }
 
-  if (eventData.event.shiftKey) {
+  if (config.defaultPencil && !eventData.event.shiftKey) {
+    config.activePencilMode = true;
+  } else if (!config.defaultPencil && eventData.event.shiftKey) {
     config.activePencilMode = true;
   }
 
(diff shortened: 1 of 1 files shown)


JamesAPetts:
Hey, thanks for your contribution!

I'm happy to merge this approach to v2 if that's what you require, however we'd need a slight bit more work on your `alwaysClose: false` config option. In this case I believe you should not calculate the stats (i.e. don't store it in toolData), and thus also not show the stats window. This is because in the case of an unclosed polygon the stats are undefined, and its super important that you don't add erroneous statistics that may be used in a treatment pipeline.

For reference, If you want to check out `3.0.0-rc1`, we have freehand mode by default in this fashion:

- Drag with the mouse -> freehand.
- Click, move, click -> Polygonal.

I.e. totally intuitive (as you seem to be going for), and no shift-clicking required. You can also switch between them freely whilst you are drawing a region.

`vNext` is going to be the maintained branch going forwards, (we are currently at rc1, and a full upgrade is coming very soon!). So if you can choose to contribute a similar open polygon mode there instead if you wish.

> Features for cornerstoneTools v2 because brush tool not working correctly with this version.
What exactly do you mean by this? The brush tool has some issues in v2, but how does this relate to the current PR?

JamesAPetts:
> when create two instances of cornerstone brush draws in wrong place and starts to throw a lot of exceptions

Ahh yes, this was one of the big motivators for v3. In v3 you can have the brush working on as many viewports as you wish. Have you tried the new release candidate?

> eraser tool doesn't work correctly with brush

The brush has its own eraser built-in in v3, by ctrl+clicking.

> and it was easier for me to modify freehand instead of brush.

What did you want to achieve with the brush? The brush in v3 is fully featured, with multiple segmentation colors, overlapping segmentations possible, fully transformable with canvas rotation/flips, etc.

> I only disable drawing of line between first and last handle. Like you see in screenshoot tool stats are not undefined.

Ok, but could you explain the use case for this? You still want the stats, but you don't close the polygon. If the user draws a straight line (a valid open polygon), there will be some stats there that are rather meaningless? Perhaps I'm missing the use-case, if you could elaborate.

Thanks,

Just to be clear v3 will the maintained version moving onwards from ~January.

drozdzynski:
We had cornerstoneTool `v0.14` and with this old version I can easily create new tool based on `freehand`, we decide to upgrade cornerstone to stable `v2` and now I cant easily create new tool. We use corenerstone in production that's why we need this modification inside cornerstoneTools v2. `freehand` without closing is inside our client business logic and this is the reason of this pull request. We plan to migrate to cornerstoneTools `v3` just after it will be released in stable channel.

dannyrb:
@drozdzynski, that sounds like a frustrating situation to be in. Unfortunately, we don't have the bandwidth to assess how new features and changes in behavior (for version 2) will impact `cornerstone-tools` users. I recommend creating a fork of `cornerstone-tools` v2, and adding the changes you need. When v3 is released and you've had a chance to assess whether it meets your current needs, we would be more than happy to discuss whether enhancements like this would be a good candidate for adoption.

dannyrb:

PR #757, opened 2018-12-20
vNext - Reorganize lib modules for better es6 support

drozdzynski:
Improvement of import modules outside cornerstoneTools in es6 schema instead of function.

**What is the current behavior?** (You can also link to an open issue here)
```javascript
import cornerstoneTools from "cornerstoneTools";
const drawTextBox = cornerstoneTools.import("drawing/drawTextBox");
```

**What is the new behavior (if this is a feature change)?**
```javascript
import cornerstoneTools from 'cornerstoneTools'
import * as lib from 'cornerstoneTools/src/lib'
import { drawing } from 'cornerstoneTools/src/lib'
import { drawTextBox } from 'cornerstoneTools/src/lib/drawing'

cornerstoneTools.lib.drawing.drawTextBox(...args)
lib.drawing.drawTextBox(...args)
drawing.drawTextBox(...args)
drawTextBox(...args)
```

**Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
All `cornerstoneTools.import(module)` should be replaced with es6 import schema.

* **Other information**:
`src` directory has been removed from `.npmingnore`. It is now included inside package on release.


drozdzynski:
Hello,

Quill based imports are bad if you want to use only specific part of code.
If I use import based on Quill, my final JS file will be bigger than if I use `import thing from '...'`.

> Does this only work for users not leveraging bundling?

In browser you can still get lib's by `cornerstoneTools.lib.drawing.drawTextBox` like I wrote before.

>  I see you removed `src` from `.npmignore`. Can you expand a bit on your use case and any limitations you foresee us encountering?

`src` directory must be included in NPM release package to be able to import it inside project which uses for example webpack, parcel etc.

dannyrb:
@drozdzynski, your feedback makes sense. I think we'll see all tools eventually available in "packs" or their own NPM modules. At that point, most of the code in this library would be core to making sure users can effectively create, add, and manage tools. The benefits of treeshaking become minimal, especially when put into the context of loading many large DICOM images for manipulation.

Maybe we can find some middle ground where we add a new output target that works better for your use case, but still allows this one? Ideally, I do want us to be able to offer the original source for treeshaking, but also packaged versions for others' use cases.

drozdzynski:
Thanks for response, I wrote in first message an example with import from root package:
```
import cornerstoneTools from 'cornerstoneTools'
cornerstoneTools.lib.drawing.drawTextBox(...args)
```
So you will be able to import whole package at once.

Issue #762, opened 2019-01-04
Cobbs angle broken


kofifus:
This must have happened in some recent code change cause I know it was working before.

- go to https://rawgit.com/chafey/cornerstoneTools/master/examples/allImageTools/index.html
- click Angle
- create an angle
- now try to grab the center circle connecting the two lines and drag it

That used to create a cobbs angle but now that circle (+label)  just goes out but the lines don't change ..

Would appreciate if this can be looked at !

Thanks!

GabrielFDuarte:
Just solved it here changing the function onImageRendered(e) present in cornerstoneTools.js for the same function present in older versions, like cornerstoneTools.js v0.7.8. You just need to remove eventData at parameter and create const eventData = e.detail; at the top inside function.

GabrielFDuarte:
I'll give you the changes. Still learning how to use GitHub ...
I changed the onImageRendered (e, eventData) function at line 982 of the cornerstoneTools.js file to the old version, v0.7.8 version of the same file.

```
// /////// BEGIN IMAGE RENDERING ///////
function onImageRendered(e) {
    
    const eventData = e.detail;

        // if we have no toolData for this element, return immediately as there is nothing to do
        var toolData = cornerstoneTools.getToolState(e.currentTarget, toolType);
        if (toolData === undefined) {
            return;
        }

        // we have tool data for this element - iterate over each one and draw it
        var context = eventData.canvasContext.canvas.getContext('2d');
        context.setTransform(1, 0, 0, 1, 0, 0);

        //activation color 
        var color;
        var lineWidth = cornerstoneTools.toolStyle.getToolWidth();
        var font = cornerstoneTools.textStyle.getFont();
        var config = cornerstoneTools.angle.getConfiguration();

        for (var i = 0; i < toolData.data.length; i++) {
            context.save();

            // configurable shadow
            if (config && config.shadow) {
                context.shadowColor = config.shadowColor || '#000000';
                context.shadowOffsetX = config.shadowOffsetX || 1;
                context.shadowOffsetY = config.shadowOffsetY || 1;
            }

            var data = toolData.data[i];

            //differentiate the color of activation tool
            if (data.active) {
                color = cornerstoneTools.toolColors.getActiveColor();
            } else {
                color = cornerstoneTools.toolColors.getToolColor();
            }

            // draw the line
            context.beginPath();
            context.strokeStyle = color;
            context.lineWidth = lineWidth;

[...]

PR #766, opened 2019-01-13
Fix broken cobbs angle

kofifus:
solves issue https://github.com/cornerstonejs/cornerstoneTools/issues/762

The recent version of `angleTool.js` changed to use `drawJoinedLines` which is wrong for cobbs angle (where the lines  are not joined).
This PR uses `drawLines` instead which gives the desired result.


dannyrb:
closes #762 

Thanks for this fix, @kofifus
(and for your patience as we get through the holiday season and beginning of the year).

PR #770, opened 2019-01-23
Store initial measurementData in tool state

kofifus:
Sore a _copy_ of the initial measurementData in the tool state that can later be used to compare with the current state.

A scenario where this is useful is to solve the current issue where users can create 'dud' annotations (ie length or angle) by just doing a click. Below is a function that can be associated with a mouseup event on the element that uses this PR to  detect and remove such annotations:

```
function cleanEmptyTools(element) {
	const handlesEqual = (h1, h2) => {
		if (h1.start && (!h2.start || h1.start.x!==h2.start.x || h1.start.y!==h2.start.y)) return false;
		if (h1.end && (!h2.end || h1.end.x!==h2.end.x || h1.end.y!==h2.end.y)) return false;
		if (h1.start2 && (!h2.start2 || h1.start2.x!==h2.start2.x || h1.start2.y!==h2.start2.y)) return false;
		if (h1.end2 && (!h2.end2 || h1.end2.x!==h2.end2.x || h1.end2.y!==h2.end2.y)) return false;
		return true;
	}
	
	const toolsState = T.globalImageIdSpecificToolStateManager.toolState;
	let enabledElement = null;
	try { enabledElement = cornerstone.getEnabledElement(element); } catch (e) { return; }
	if (!enabledElement || !enabledElement.image) return;
	const imageId = enabledElement.image.imageId;
	if (!imageId || !toolsState.hasOwnProperty(imageId)) return;
	const imageIdToolState = toolsState[enabledElement.image.imageId];
	
	let imageNeedsUpdate=false;
	Object.keys(imageIdToolState).forEach(toolType => {
		let toolData = imageIdToolState[toolType].data;
		toolData.forEach((data, index) => {
			if (!data.initial) return;

			var remove=false;
			if ((toolType==='length' || toolType==='angle') && handlesEqual(data.handles, data.initial.handles)) remove=true;
			
			if (remove) {
				toolData.splice(index, 1);
				imageNeedsUpdate=true;
			}
		});
	});

	if (imageNeedsUpdate) cornerstone.updateImage(element); 
}
```

(above code only shows handling 'dud' length and angle but can easily be extended for all annotations)

### DIFF
--- a/src/stateManagement/toolState.js
+++ b/src/stateManagement/toolState.js
@@ -20,6 +20,7 @@ function getElementToolStateManager (element) {
 function addToolState (element, toolType, measurementData) {
   const toolStateManager = getElementToolStateManager(element);
 
+  if (!measurementData.initial) measurementData.initial = JSON.parse(JSON.stringify(measurementData));
   toolStateManager.add(element, toolType, measurementData);
 
   const eventType = EVENTS.MEASUREMENT_ADDED;
--- a/src/stateManagement/toolState.js
+++ b/src/stateManagement/toolState.js
@@ -20,7 +20,9 @@ function getElementToolStateManager (element) {
 function addToolState (element, toolType, measurementData) {
   const toolStateManager = getElementToolStateManager(element);
 
-  if (!measurementData.initial) measurementData.initial = JSON.parse(JSON.stringify(measurementData));
+  if (measurementData.initial === undefined) {
+    measurementData.initial = JSON.parse(JSON.stringify(measurementData));
+  }
   toolStateManager.add(element, toolType, measurementData);
 
   const eventType = EVENTS.MEASUREMENT_ADDED;


dannyrb:
If I understand correctly, you want this API change to make it easier to `clear` tools like `length` and `angle` when the interaction causes one or more handles to sit directly on top of one another? Am I understanding correctly? Are there other problems this API change helps solve?

I think I'd prefer to see:

- A change that will offset/move handles if they are "too close" to one another
- OR `angle` / `length` tool specific fixes for this issue

If knowing the `initial` values helps solve other problems, then I would be more open to considering the change.

kofifus:
The PR is general - the idea is to make the _initial_ measurement data of tools available to external code. This can be useful whenever there is a need to calculate something based on the difference between the initial position and the current (or final) one.

My use case is not one of the ones you describe, though the PR maybe useful in those situations as well.

My use case is to prevent 'empty' annotations, if you got for example to https://rawgit.com/chafey/cornerstoneTools/master/examples/allImageTools/index.html and press on 'angle' and then just click on the image you will get an empty annotation. The same will happen with length etc.

I needed to prevent such 'empty' annotations in our implelmentation, that is to remove them on mouseup, I wanted to avoid modifying the internal implementation of angleTool or lengthTool etc. With this PR it is possible to write external code that can determine if a tool's measurement data is different than its initial data and if no remove it - as I demonstrated in the `cleanEmptyTools` function above that can be attached a mouseup event handler.

As said above this is just one use case and storing the initial measurement data can be valuable in other scenarios.

Thanks!

Zaid-Safadi:
@kofifus, if you have an event when the mouse is up (drawingEnded/Completed) and the data has been calculated and provided in the event then I believe you can validate the values to make a decision and delete empty measurements, right?

If this is the case then please follow this issue #787 

However, if you need to store the initial data, there is no reason to do this in hte library, you can listen to the "added" event and attach the data from your application.

kofifus:
> if you have an event when the mouse is up (drawingEnded/Completed) and the data has been calculated and provided in the event then I believe you can validate the values to make a decision and delete empty measurements, right?

Even if I had such an event I will not have the original measurementData - that's exactly the point.
Sure tools do store information which can be helpful in doing such decisions - ie handles.start for length etc, however it is IMO cleaner to have access to the initial measurement data as part of the tool data as this may be useful in general and allows much better reasoning about what changed from start to finish.

`However, if you need to store the initial data, there is no reason to do this in hte library, you can listen to the "added" event and attach the data from your application.`

yes I can but I will have to do this for each tool I use individually which is ugly

kofifus:
correction - on second though (and test) I can indeed store the initial measurement data from my application using `MEASUREMENT_ADDED`

I still feel this would be generally useful but I'll close it for now.

PR #771, opened 2019-01-26
Update ellipse statistics to calculate min/max

dannyrb:


dannyrb:
Note: `package-lock.json` was causing issues with CircleCI. It looks like we may have some outdated dev dependencies somewhere in our tree.

CC: @kevinleedrum - your min/max changes

PR #778, opened 2019-01-29
Elliptical touch-up

dannyrb:
#### Problem

- ROI calculations were not populated on first placement
- Calculations were only invalidated on `upHandler` for `moveAll` or `moveHandle`
- Misc. performance and code smell issues

#### Solution

- Update `moveAllHandles`, `moveHandle` and `moveNewHandle` to invalidate annotation data when appropriate
- Break `EllipticalRoi` code into smaller, composable methods
- Add calculation throttling (`110 ms`) to improve performance
- Slipped in the ability to toggle on additional stats about the ellipse
- Slipped in the ability to toggle `HU` as a unit for `CT` images

#### Potential Future Enhancements

- Set throttling w/ configuration
- Set unit as `mm` or `cm`
- Configurable `manipulator` to not invalidate when only a textbox has been moved
- Update documentation
- More unit tests <3

#### Side-by-side mean & stdDev

image

#### `showMinMax: true`

image

### DIFF
--- a/src/tools/annotation/EllipticalRoiTool.js
+++ b/src/tools/annotation/EllipticalRoiTool.js
@@ -38,6 +38,10 @@ export default class EllipticalRoiTool extends BaseAnnotationTool {
     const defaultConfig = {
       name: 'EllipticalRoi',
       supportedInteractionTypes: ['Mouse', 'Touch'],
+      configuration: {
+        // showMinMax: false,
+        // showHounsfieldUnits: true,
+      },
     };
     const initialConfiguration = Object.assign(defaultConfig, configuration);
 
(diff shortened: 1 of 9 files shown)


src/tools/annotation/EllipticalRoiTool.js
```diff
@@ -311,112 +237,257 @@ export default class EllipticalRoiTool extends BaseAnnotationTool {
...
+        meanStdDevSUV.mean.toFixed(2)
+      )}`;
+      const stdDevSuvString = `${SUVtext}${numbersWithCommas(
+        meanStdDevSUV.stdDev.toFixed(2)
+      )}`;
+
+      const targetStringLength = Math.floor(
+        context.measureText(`${stdDevString}     `).width
+      );
+      while (context.measureText(meanString).width < targetStringLength) {
```
dannyrb: It's a little silly, but I'm pretty happy about this.

PR #779, opened 2019-01-30
Feature/rectangle roi touch up

dannyrb:
## Reactangle ROI Enhancements

Sister to: https://github.com/cornerstonejs/cornerstoneTools/pull/778

image

### DIFF
--- a/src/tools/annotation/RectangleRoiTool.js
+++ b/src/tools/annotation/RectangleRoiTool.js
@@ -333,8 +333,8 @@ function _calculateStats(image, element, handles, modality, pixelSpacing) {
     mean: roiMeanStdDev.mean || 0,
     variance: roiMeanStdDev.variance || 0,
     stdDev: roiMeanStdDev.stdDev || 0,
-    // min: ellipseMeanStdDev.min || 0,
-    // max: ellipseMeanStdDev.max || 0,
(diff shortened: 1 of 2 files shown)


src/tools/annotation/RectangleRoiTool.js
```diff
@@ -351,12 +351,16 @@ function _calculateRectangleStats(sp, rectangle) {
...
+  let max = sp ? sp[0] : null;
 
   for (let y = rectangle.top; y < rectangle.top + rectangle.height; y++) {
     for (let x = rectangle.left; x < rectangle.left + rectangle.width; x++) {
       sum += sp[index];
       sumSquared += sp[index] * sp[index];
-      count++;
+      min = Math.min(min, sp[index]);
+      max = Math.max(max, sp[index]);
+      count++; // TODO: Wouldn't this just be sp.length?
```
kevinleedrum: Considering the fact that `sum += sp[index]` isn't checking to see if `index` is out of bounds, and `count` and `index` should always be the same, I think you're right.  I think `count` will always end up being the same as `sp.length` after the `for` loops.  I also don't think `count` is even used once it's returned from both `_calculateRectangleStats` and  `_calculateStats`.

PR #788, opened 2019-02-04
Tool to display DICOM overlays

kofifus:
When enabled this tool will disable the DICOM overlays stored in `image.overlays`

The overlays data is created by `cornerstoneWADOImageLoader` - see https://github.com/cornerstonejs/cornerstoneWADOImageLoader/pull/239

Implements DICOM standard C.9 - http://dicom.nema.org/dicom/2013/output/chtml/part03/sect_C.9.html

### DIFF
--- a/src/imageTools/overlays.js
+++ b/src/imageTools/overlays.js
@@ -1,0 +1,77 @@
+import external from '../externalModules.js';
+
+// const toolType = 'overlays';
+
+function onImageRendered (e) {
+  const eventData = e.detail;
+
+  if (!eventData || !eventData.enabledElement || !eventData.image || !eventData.image.overlays || !eventData.image.overlays.length === 0) {
+    return;
+  }
+
+  const enabledElement = eventData.enabledElement;
+  const context = enabledElement.canvas.getContext('2d');
+  const overlays = eventData.image.overlays;
+
+  const imageWidth = Math.abs(enabledElement.viewport.displayedArea.brhc.x - enabledElement.viewport.displayedArea.tlhc.x) * enabledElement.viewport.displayedArea.columnPixelSpacing;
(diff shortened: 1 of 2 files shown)


dannyrb:
@kofifus, I don't believe this is a strong candidate for version 2. Ideally, I think it's better suited as a maintained 3rd party plugin in Version 3.

You can find an example of a Version 3 Plugin here: https://github.com/QSolutionsLLC/cornerstone-tool-image-statistics

If you ignore all of the tooling, the meat of the plugin lives in this one file: https://github.com/QSolutionsLLC/cornerstone-tool-image-statistics/blob/master/src/index.js


The advantage of tool plugins is that they allow us to be more agile in developing core features for `cornerstone-tools` without needing to update/maintain _every single tool_ as we do. It also empowers our community to create more advanced, niche tools that may not benefit our wider audience. We're working on a way to find 3rd party plugins and list them on our website or on a "marketplace".

We can continue the discussion here if you would like. I'm going to close this issue for now. This is a plugin I would be interested in helping to build as a 3rd party plugin -- please don't hesitate to follow up with me on this.

kofifus:
OK up to you, to me this is a misunderstanding of what overlays are - they are part of the DICOM standard describing what constitutes an image, essentially they are part of the pixel data. Any PACS I know shows them out of the box not as a 'feature' or 'tool' (3rd party or not), and every radiologist and practice will expect them to be there - in fact I think there may be legal issues for practices if you show images without overlays as they may contain vital information.

kofifus:
It will take me awhile to move to V3, we did quite substantial modifications to cornerstone so merging may be tricky so I don't expect to submit this as a 3rd party plugin soon - especially since I don't feel it is!

cornerstonejs/cornerstoneWADOImageLoader#239 can still be merged and used as the basis of future work ..

dannyrb:
@kofifus, our super official process for determining where this should live, and in what form:

image

So we at least have consensus on that it should be maintained in a cornerstone repository. Ideally as a tool. I'll take a peak at your code sometime this week and what transitioning to the new tool setup would entail. Most overlay tools take < 30 minutes to transition

kofifus:
thx

PR #793, opened 2019-02-05
Resolve merge conflicts

dannyrb:


dannyrb:
@JamesAPetts this takes only the changelog updates from master, but syncs history

dannyrb:
If we "lose" any functionality in the transition to v3, it likely happened in one of this commits

Issue #775, opened 2019-01-27
Freehand handles not closed after loading state data


Zaid-Safadi:
### Prerequisites

* [ ] Are you running the latest version?
V2.4.0
* [x] Are you reporting to the correct repository?
* [x] Did you perform a cursory search?

### Description
After loading the state of the freehand data. click on the first handle which closes the shape and drag, the handle and closing line will be separated and shape is not closed anymore.

### Steps to Reproduce

1. Save the freehand state
2. load the state to draw the line
3. Click and move the first handle point
4. The only one handle is moved leaving the shape open.

**Expected behavior:** 
The point should be moved and both first and last handles moving together to keep the shape closed.

**Actual behavior:** 
Only one handle moves.

Zaid-Safadi:
When drawing the handles, the last handle line is an object that points to the first handle. The code intentionally doesn't update both handles on drag since it is the same object. However, once the data is saved/loaded, the object reference is lost.

To fix this, we should always update the last handle line to sync with the first handle position:

dragHandle  method:
  ```
if (currentHandle) {
    const lastLineIndex = data.handles[currentHandle - 1].lines.length - 1;
    const lastLine = data.handles[currentHandle - 1].lines[lastLineIndex];

```
can be changed such that:
```
  if(currentHandle === 0){
    handleIndex = data.handles.length-1;
  } else {
    handleIndex = currentHandle - 1;
  }

  if (currentHandle >= 0) {
    const lastLineIndex = data.handles[handleIndex].lines.length - 1;
    const lastLine = data.handles[handleIndex].lines[lastLineIndex];


```

If fix is OK, I'll go ahead and submit a PR.

dannyrb:
@Zaid-Safadi, if you make a PR for v2, and vouch that this fix seems reasonable, and works after you've tested it a bit, I can merge it. 

@JamesAPetts, it may be worth checking to see if this is a `vNext` issue as well.

I know we've been saying it for a while now, but we do intend to switch to vNext _very soon_. Our only hold-up right now is the _getting started_ documentation is outdated.

Zaid-Safadi:
Fixed in v2 and v3

PR #804, opened 2019-02-06
Fix saving and loading the freehand moust tool. Issue # 775 for v3

Zaid-Safadi:
Updated freehand example

This PR fixes the issue described here:
Freehand handles not closed after loading state data #775

Also, another issue has been fixed that is caused by mixing the freehand "handles" as an Array type and an Object type at the same time causing the serialization of the handles array to lose the object data fields "textBox", "start"... once serialized.

The "handles" has been changed to be an object which contains a separate field called "dataHandles" of type Array.

Finally, although the freehand now can be saved/loaded successfully, the save using JSON.serialize will fail if the sculpture tool is applied on the freehand because of circular object dependency. However, this is best handled as a separate issue.

Saving/loading the freehand allow the last handle to be moved seperately from the start handle as described here #775 

Also, once loaded, the text box doesn't show up and you can't draw any more freehand because of debugger errors. 

Freehand can be saved/loaded successfully and text box rendered.

No

* **Other information**:

### DIFF
--- a/src/tools/annotation/FreehandMouseTool.js
+++ b/src/tools/annotation/FreehandMouseTool.js
@@ -87,7 +87,7 @@ export default class FreehandMouseTool extends BaseAnnotationTool {
       invalidated: true,
       color: undefined,
       handles: {
-        dataHandles: [],
+        points: [],
       },
     };
 
@@ -112,7 +112,7 @@ export default class FreehandMouseTool extends BaseAnnotationTool {
    * @returns {Boolean}
    */
   pointNearTool(element, data, coords) {
-    const validParameters = data && data.handles && data.handles.dataHandles;
+    const validParameters = data && data.handles && data.handles.points;
 
     if (!validParameters) {
       throw new Error(
@@ -140,9 +140,9 @@ export default class FreehandMouseTool extends BaseAnnotationTool {
   distanceFromPoint(element, data, coords) {
     let distance = Infinity;
 
-    for (let i = 0; i < data.handles.dataHandles.length; i++) {
+    for (let i = 0; i < data.handles.points.length; i++) {
       const distanceI = external.cornerstoneMath.point.distance(
-        data.handles.dataHandles[i],
+        data.handles.points[i],
         coords
       );
 
@@ -168,14 +168,18 @@ export default class FreehandMouseTool extends BaseAnnotationTool {
   distanceFromPointCanvas(element, data, coords) {
     let distance = Infinity;
 
+    if (!data) {
+      return -1;
+    }
+
     const canvasCoords = external.cornerstone.pixelToCanvas(element, coords);
 
-    const dataHandles = data.handles.dataHandles;
+    const points = data.handles.points;
 
-    for (let i = 0; i < dataHandles.length; i++) {
+    for (let i = 0; i < points.length; i++) {
       const handleCanvas = external.cornerstone.pixelToCanvas(
         element,
-        dataHandles[i]
+        points[i]
       );
 
       const distanceI = external.cornerstoneMath.point.distance(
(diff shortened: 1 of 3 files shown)


src/tools/annotation/FreehandMouseTool.js
```diff
@@ -110,7 +112,7 @@ export default class FreehandMouseTool extends BaseAnnotationTool {
    * @returns {Boolean}
    */
   pointNearTool(element, data, coords) {
-    const validParameters = data && data.handles;
+    const validParameters = data && data.handles && data.handles.dataHandles;
```
JamesAPetts: Hmm, could we rename the `dataHandles` array to just `points`. This would avoid tongue twisters like `data.handles.dataHandles`.

src/tools/annotation/FreehandMouseTool.js
```diff
@@ -767,10 +790,10 @@ export default class FreehandMouseTool extends BaseAnnotationTool {
     const config = this.configuration;
     const currentTool = config.currentTool;
 
-    const dataHandles = toolState.data[currentTool].handles;
+    const dataHandles = toolState.data[currentTool].handles.dataHandles;
 
     // Don't allow the line being modified to intersect other lines
-    if (dataHandles.invalidHandlePlacement) {
+    if (toolState.data[currentTool].handles.invalidHandlePlacement) {
```
JamesAPetts: We should use the cached value from line 793 here.

src/tools/annotation/FreehandMouseTool.js
Zaid-Safadi: agree, this is actually what I had in mind initially but I saw "dataHandles" used in the code to reference this structure so I re-used it.

src/tools/annotation/FreehandMouseTool.js
Zaid-Safadi: the value in 793 is for the "dataHandles/points" array. This is accessing the handles object. we can still cache it though and use it for line 814

src/tools/annotation/FreehandMouseTool.js
JamesAPetts: That's fair, I originally used `dataHandles` as a shorthand for the reference to `data.handles`, but if we have another property under `data.handles` as we do here, we should give it a new name 👍 .

src/tools/annotation/FreehandMouseTool.js
JamesAPetts: Aha, sorry I missed that, I guess this highlights the `handles.dataHandles` confusion haha. But yeah, caching it for line 814 is a good shout 👍 .

dannyrb:
@Zaid-Safadi, I see you have a mix of issues/pull-requests for versions 2 and 3 in progress. Sometime this week, can you let me know what you need from me to move these forward? Feel free to `@mention` me on Slack if it's easier.

JamesAPetts:
Likewise, I'm happy with all the PR once my very minor comments are addressed :). Feel free to ping me again/message me on slack if you have any questions.

src/tools/annotation/FreehandMouseTool.js
Zaid-Safadi: Renamed to "points"

src/tools/annotation/FreehandMouseTool.js
Zaid-Safadi: done

Zaid-Safadi:
@dannyrb @JamesAPetts, thanks both for reviewing the changes. It was a busy week but I was planning to take care of these PRs this weekend :)

I merged the v2 PRs and resolved any comments for this v3. I'll keep the issues that are not resolved in v3 open such that we can also apply a similar fix.

dannyrb:
Thanks for your hard work, @Zaid-Safadi!

Do you want me to cut path releases for v3 and v2?

For v2, publishing should be as simple as pulling that branch, bumping the version, and running `npm install` then `npm publish`

For v3, we've switched to `semantic-release` and versions are automatically updated and published when commits are formatted correctly. You can "fake" it this time by updating the docs, staging those changes, and running `npm run commit`. Follow the prompts that apply to the v3 change you made. PR that commit against master. On merge, a changelog and release will be cut.

If you would rather I cut the releases, I can on Monday.

Zaid-Safadi:
Hey @dannyrb, I have one more commit that I'll be adding soon to v2 for the "completed" event before publishing a new version. 

As for v3, I'll try to actually submit a PR for one of the issues fixed in v2 and trigger the release. If I didn't get to this today, please feel free to push it on Monday.

Issue #833, opened 2019-02-12
Since 3.0 multipart tools hog event inputs. Can't zoom/pan whilst in a draw loop.


JamesAPetts:
With the introduction of `cornerstoneTools 3.0`'s new spangled event dispatcher, some tools are able to block event dispatching and deal with mouse/touch input internally to the tool. This is useful for e.g. `FreehandMouseTool`, where an indeterminate number of clicks/drags are are required to complete the annotation.

This currently comes with the downside of not being able to use tools bound to other input during the tools usage. It'd be really useful to be able to zoom and/or pan _during_ a multi-part tools draw loop.

After some discussion, @dannyrb and I came to the conclusion that only non-annotation tools (i.e. when `!(tool instanceof BaseAnnotationTool)` should be usable during a multi-part draw loop. If you have a use case that requires annotation tools to be used during a multi-part draw loop, please leave a comment and discuss.

TODO:
- Rename `state.isToolLocked` to `state.multiPartToolActive`.
- Change the dispatchers to still dispatch events to non-annotation tools whilst `state.multiPartToolActive === true`.
- Create a `multiPartTool` mixin with helpers for easy mouse/touch mask management for event loops. The idea is that this will be useful but optional, as multi-part tools by nature are rather non-standard.

JamesAPetts:
Addressed in part by #836 .

Creating a multi-part tool will be a future addition.

PR #836, opened 2019-02-13
Feature/event dispatch locking changes

JamesAPetts:
feat(Event Dispatchers): Can now use non-annotation tools during a multi-part tool's loop.
    
Added a new flag, isMultiPartTool to tools, and a global state trigger you may set called isMultiPartToolActive. If this is true, only non-annotation tools may be used whilst you are in a multi-part tool's event loop.
    
re #833

### DIFF
--- a/src/eventDispatchers/mouseEventHandlers/mouseDown.js
+++ b/src/eventDispatchers/mouseEventHandlers/mouseDown.js
@@ -6,6 +6,7 @@ import getToolsWithMoveableHandles from '../../store/getToolsWithMoveableHandles
 import { findHandleDataNearImagePoint } from '../../util/findAndMoveHelpers.js';
 import getInteractiveToolsForElement from './../../store/getInteractiveToolsForElement.js';
 import getToolsWithDataForElement from './../../store/getToolsWithDataForElement.js';
+import filterToolsUseableWithMultiPartTools from './../../store/filterToolsUsableWithMultiPartTools.js';
 
 /**
  * MouseDown is called before MouseDownActivate. If MouseDown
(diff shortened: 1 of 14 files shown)


JamesAPetts:
TODO:

- Moving an individual node of the freehand tool locks things up.
- Should we disable scrolling during multi-tool use? Surprisingly scrolling whilst using the freehand tool here doesn't break anything, but you can't continue to draw until you scroll back to the slice. Is there any need for the stack scroll tools to be usable with a multi-tool?

pieper:
> * Should we disable scrolling during multi-tool use? Surprisingly scrolling whilst using the freehand tool here doesn't break anything, but you can't continue to draw until you scroll back to the slice. Is there any need for the stack scroll tools to be usable with a multi-tool?

For a similar tool in Slicer it was helpful to allow scrolling in the middle of drawing.  It lets you better understand the 3D nature of the outline.  What I did was display the polyline-so-far as a dashed line to indicate it's not active.  The line goes back to solid when you are back on the slice where you started.

JamesAPetts:
@pieper Hmm, that makes sense. With the current way `cornerstoneTools` works, rendering the WIP ROI in a different slice might be challenging. A tool's data is drawn by checking if there is annotation data for that imageId, so you'd lose the reference when you scroll.

I guess we could save a reference to the WIP toolData on the tool itself when start drawing, then in the `renderToolData` method we could check if the imageId isn't the same as the WIP annotation. You could then overlay this onto a different image. 

Then you delete the reference on the tool when the annotation is complete.
It feels a bit hacky, but a way to do something like this in the current framework.

pieper:
The way the current Slicer Segment Editor operates is to let you scroll while keeping your WIP drawing and then applying it to the slice where you complete it.  This works pretty well too and might be easier and more logical to implement.  Basically you would just move the tool state to the new image.

JamesAPetts:
Following this discussion I won't block scroll from being used. Any expansion to the freehand tool UI to improve the UX whilst scrolling would be introduced in a subsequent PR.

PR #838, opened 2019-02-15
Chore/add examples

dannyrb:
## WIP


dannyrb:
This isn't going to work flawlessly on the first go. Merging is mostly to kickstart the CI and to see how it handles the updated config.

This...

- Adds examples in Jekyll in `/examples`
- Spins up a docker instance w/ Ruby
- Builds the example's website/output
- Runs the `npm run docs:publish` command

If this works as expected, we still need to:

- Update the baseURL and env variables for `/examples`
- Copy the `examples` output to what we'll push to GitHub pages
- Link each example to it's output/source code in GitHub

Issue #787, opened 2019-02-03
Add measurement completed event


Zaid-Safadi:
### Prerequisites

* [x] Are you running the latest version?
* [x] Are you reporting to the correct repository?
* [x] Did you perform a cursory search?

### Description
Currently the library support the following events for measurements:
- MEASUREMENT_ADDED: 'cornerstonetoolsmeasurementadded',

-   MEASUREMENT_MODIFIED: 'cornerstonetoolsmeasurementmodified',

-   MEASUREMENT_REMOVED: 'cornerstonemeasurementremoved',

This is a proposal to support a new event that is fired once drawing the measurement is completed.
-   MEASUREMENT_COMPLETED: cornerstonetoolsmeasurementcompleted

### Steps to Reproduce
1. Draw a measurement tool

**Expected behavior:** [What you expected to happen]
An event to be fired once the user complete drawing the measurement 

**Actual behavior:** [What actually happened]
No event exists

dannyrb:
Hmm.. This is interesting and makes sense for most annotation tools that aren't "valid" until multiple steps are completed. Some interesting things to consider:

- The main benefit of this over using _added_ and _modified_ is that you don't need to "validate" annotations are in a "complete" storable/re-storable state.
    - IE. An unclosed `freehand`, or an `angle` with only two handles
- Is a measurement truly "added" if it is not complete? Should we already be using this event for annotations that have been successfully "added/completed"?
- Are there any other `lifecycle events` for an Annotation/measurement that we should be considering? Are the previous events already clearly defined and consistently used?
    - Consider: multi-step tools, tools that may have multiple different, but "completed" states, and temporary annotations (dragProbe)

I'm mostly thinking out loud. I like this event, and we'll probably move forward on adding it, but while I have a captive audience, worth shaking out other considerations ^_^

Zaid-Safadi:
Hey @dannyrb , here are some thoughts around this...

I can think of multiple use-cases where I need to do something once the user completes drawing the measurement. These are some examples that I actually encountered:
- validation 
- displaying a dialog
- updating the measurement
- updating another UI to add the measurement.

For example, the arrow annotation displays an input dialog after completion to capture a label. This same behavior might be needed with all other measurements where you want to allow the user to add a label to an ROI (same as in the OHIF lesion tracker target/non-target labeling)

> "Is a measurement truly "added" if it is not complete?" 

[...]

PR #845, opened 2019-02-16
Add measurement completed event #787

Zaid-Safadi:
Updated the example AllImageTools

Introduce a new measurement event "cornerstonemeasurementcompleted" that is fired for the following tools:
1. Angle Tool
2. Arrow Annotate
3. Elliptical ROI
4. Freehand
5. length
6. Probe
7. Rectangle ROI
8. Simple Angle
9. Angle Tool

No such event exists. No way to tell when a user completes drawing the measurement/annotation.

1. An event "cornerstonemeasurementcompleted" is fired once the user completes drawing/moving the annotation.
2. When this event is fired, the tool data will contain any calculation that the tool will produce when drawing the image. In the past, calculating the data only happens on rendering, now it also happens before firing this event.

No
completedeventsdata2

* **Other information**:

### DIFF
--- a/src/imageTools/angleTool.js
+++ b/src/imageTools/angleTool.js
@@ -1,3 +1,4 @@
+import EVENTS from '../events.js';
 import external from '../externalModules.js';
 import mouseButtonTool from './mouseButtonTool.js';
 import touchTool from './touchTool.js';
@@ -8,6 +9,8 @@ import drawHandles from '../manipulators/drawHandles.js';
 import { getToolState } from '../stateManagement/toolState.js';
 import lineSegDistance from '../util/lineSegDistance.js';
 import { getNewContext, draw, setShadow, drawLines } from '../util/drawing.js';
+import triggerEvent from '../util/triggerEvent.js';
+import getColRowPixelSpacing from '../util/getColRowPixelSpacing.js';
 
 const toolType = 'angle';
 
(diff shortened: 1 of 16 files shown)


src/imageTools/angleTool.js
```diff
@@ -8,6 +9,8 @@ import drawHandles from '../manipulators/drawHandles.js';
 import { getToolState } from '../stateManagement/toolState.js';
 import lineSegDistance from '../util/lineSegDistance.js';
 import { getNewContext, draw, setShadow, drawLines } from '../util/drawing.js';
+import triggerEvent from '../util/triggerEvent.js';
```
dannyrb: I see that we're importing `EVENTS` and `triggerEvent` to all of these tools so we can create a `fireCompleted`. Can we instead create a helper method that we import?

`triggerMeasurementCompletedEvent(element, toolType, data)`

Zaid-Safadi: yes, good catch!

src/util/getColRowPixelSpacing.js
```diff
@@ -0,0 +1,19 @@
...
+    rowPixelSpacing = imagePlane.rowPixelSpacing || imagePlane.rowImagePixelSpacing;
+    colPixelSpacing = imagePlane.columnPixelSpacing || imagePlane.colImagePixelSpacing;
+  } else {
+    rowPixelSpacing = image.rowPixelSpacing;
+    colPixelSpacing = image.columnPixelSpacing;
+  }
+
+  return { rowPixelSpacing,
+    colPixelSpacing };
+}
```
dannyrb: Nice :+1:

dannyrb:
I'm hesitant to merge this until we have a v3 PR that adds the same functionality. Mostly because this would be adding a feature that most would be surprised to lose if they decided to upgrade to v3.

Do you have a timeline on when you need this? Could it wait until I have something comparable in place for the `master` branch (end of next week)? If not, I can work with you on cutting a release earlier, but if there's no rush then postponing would be ideal.

On the code itself, if you can address the my first comment, and things are working as expected with a quick `Netlify` test, I'm okay with this being merged.

Thanks for your hard work on this, @Zaid-Safadi! I appreciate you taking the time -- Always a pleasure ^_^

Zaid-Safadi:
I understand your point @dannyrb , let me see if I can speed things up a little and help out with master. 

I'll create a branch and start with some of the tools I am familiar with, hopefully this will help. I guess if the wait time between both versions is less than a week then we shouldn't be worried much about people upgrading and losing features.

Lets shoot for early next week (Monday/Tuesday?) to merge if possible?

src/imageTools/angleTool.js
Zaid-Safadi: this is now updated

dannyrb:
@all-contributors please add @Zaid-Safadi for this awesome code pr

sericaia:
Is there any way we can have this on v3 @dannyrb ? Thanks

whulbert4:
Just making a note here in case anyone else runs into the same issue. The changes here to the Arrow Annotate tool caused a bug when the anyHandlesOutsideImage comes back true in addNewMeasurement. The onHandleDoneMove never gets called to add the mouse callback events back.

My solution was to revert back to the 2.4.0 release vs. the 2.5.0 release that had this PR in it.

Not sure if there will be any more 2.X releases (though I did see a 2.6.0 one recently), but happy to try and put a PR together to fix the bug if so.

PR #852, opened 2019-02-16
WIP: DON'T MERGE - Add measurement completed event

Zaid-Safadi:
I created the PR early to discuss the updates needed to add the measurement completed event to master.

Since v3 is a complete re-design, I'd like to know if there is a more standard approach to let the tools know when a handle is done moving. 

I am also thinking wherever this method "doneMovingCallback" is called we should notify that the handle is also changed.

@dannyrb and @JamesAPetts any thoughts on this?

### DIFF
--- a/src/events.js
+++ b/src/events.js
@@ -181,6 +181,11 @@ const EVENTS = {
   /**
    *  @type {String}
    */
+  MEASUREMENT_COMPLETED: 'cornerstonetoolsmeasurementcompleted',
+
+  /**
+   *  @type {String}
+   */
   TOOL_DEACTIVATED: 'cornerstonetoolstooldeactivated',
 
   /**
--- a/src/manipulators/moveHandle.js
+++ b/src/manipulators/moveHandle.js
@@ -5,6 +5,7 @@ import { removeToolState } from '../stateManagement/toolState.js';
 import triggerEvent from '../util/triggerEvent.js';
 import { clipToBox } from '../util/clip.js';
 import { state } from './../store/index.js';
+import { publishHandleDoneMovingEvent } from '../util/moveHandleDoneEventBroker.js';
 
 const runAnimation = {
   value: false,
@@ -212,6 +213,12 @@ function _upOrEndHandler(
   if (typeof options.doneMovingCallback === 'function') {
     options.doneMovingCallback();
   }
+
+  // eslint-disable-next-line prettier/prettier
+  const args = { element,
+data: annotation };
+
+  publishHandleDoneMovingEvent(toolName, args);
 
   external.cornerstone.updateImage(element);
 }
--- a/src/manipulators/moveNewHandle.js
+++ b/src/manipulators/moveNewHandle.js
@@ -5,6 +5,7 @@ import { removeToolState } from '../stateManagement/toolState.js';
 import triggerEvent from '../util/triggerEvent.js';
 import { clipToBox } from '../util/clip.js';
 import { state } from './../store/index.js';
+import { publishHandleDoneMovingEvent } from '../util/moveHandleDoneEventBroker.js';
 
 const _moveEvents = {
   mouse: [EVENTS.MOUSE_MOVE, EVENTS.MOUSE_DRAG],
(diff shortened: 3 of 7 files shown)


dannyrb:
I think the applied change will be very similar to version 2's. Most of the work we did in v3 is to change which active/passive tool is able to handle an event, not necessarily _how_ they handle it.

IE. When we click, instead of each tool's event handlers competing to handle the event first, it instead flows through the `eventDispatcher`

- *User Clicks*
- `eventListeners/mouseEventListeners.js` normalized click event
- `eventDispatchers/mouseToolEventDispatcher.js` finds the best "interactive" tool to dispatch the event to
    - clicked `pointNearHandle`? --> `tool.handleSelectedCallback`
    - clicked `pointNearTool`? --> `tool.toolSelectedCallback`
    - No tool satisfy either? `mouseDownActivate` --> `tool.addNewMeasurement`

The `baseAnnotationTool` has a default `handleSelectedCallback` method, and a default `toolSelectedCallback` method. The first uses `moveHandleNearImagePoint` and the latter uses `moveAnnotation` -- These should be fine as is.

Where we'll likely need to call your new event is in the `addNewMeasurement` method that annotation tools default to, or in the custom `addNewMeasurement` method on tools that provide one:

Custom:

- AngleTool
- ArrowAnnotateTool
- BidirectionalTool
- CobbAngleTool
- FreehandMouseTool

(See a pattern with the tools that override the default behavior?)

The default `addNewMeasurement` methods are:

- `/src/eventDispatchers/mouseEventHandlers/addNewMeasurement.js`
- `/src/eventDispatchers/touchEventHandlers/addNewMeasurement.js`

I hope that helps. Let me know if you have questions. @JamesAPetts, feel free to correct anything I've goofed.

Zaid-Safadi:
@dannyrb I've reviewed the design and I think we are on the same page.

I am firing the event in two places: `moveHandle `and `moveNewHandle`, these are what gets called by the `addNewMeasurement` function:
https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/eventDispatchers/mouseEventHandlers/addNewMeasurement.js#L22-L34
```
  const handleMover =
    Object.keys(measurementData.handles).length === 1
      ? moveHandle
      : moveNewHandle;

  handleMover(
    eventData,
    tool.name,
    measurementData,
    measurementData.handles.end,
    tool.options,
    'mouse'
  );
```

Technically, in v2 we are now firing these events by the callback method in this `handleMover` like. Now, we are passing an `options` object that contains a callback for when moving is **"Done"**. The issue is, this callback/options -as I think it is- could be set by the client and we shouldn't be modifying it with our own that fires the event.

As for subscribing to the event, I think I am following a similar design to the dispatcher in the `moveHandleDoneEventBroker.js` (Pub/Sub pattern) which doesn't require tools to compete and listen to all the events and rather be called.  However, I think a better location for this would be under the `eventDispatchers` folder and probably renamed to something like `moveHandleDoneEventDispatcher` (or a better name altogether!)


Hope this makes sense and happy to hear your thoughts.

mikehazell:
Hey @Zaid-Safadi, @dannyrb  - Looks like this work may have stalled for the v3 codebase.

Does anyone mind if I have a go at it? I would follow the v2 implementation #845.

dannyrb:
@mikehazell, I think that would be welcomed ^_^

My memory's a bit hazy, but I'm pretty sure we want to raise this event when a new annotation's toolData is considered "complete", or more technically, when it's toolData first represents a working/placed annotation.

IE. A `cobbAngle` annotation with two lines, or a closed `freehandRoi` tool

The biggest "up for debate" question is the best place to fire the event. _Most_ tools are complete when their `addNewMeasurement` method is called. A few tools (multiple-step tools) either have a set number of steps required to be complete, or need to meet a special condition.

@Zaid-Safadi completed this change in v2 and likely has more concrete thoughts on it.

Zaid-Safadi:
Hi @mikehazell , you are v. welcome to help out in any way possible to get this closer to the finish line. Apologies I couldn't focus on this as my schedule got pretty occupied.

I think if you build on this PR and follow v.2 implementation #845 you will cover most of the cases. For the multi-step tools (e.g. freehand, cobb angle...) you can fire the event from the respective class as **only a freehand knows when a freehand is complete.**

@dannyrb and I would be happy to help and advice if you have any questions.

mikehazell:
Sounds good. I'll get cracking on this asap.

dannyrb:
Completed here: https://github.com/cornerstonejs/cornerstoneTools/pull/913

PR #857, opened 2019-02-21
Fix the issue which prevents changing bidirectional by handles

evren217:
- Set state.isToolLocked flag to prevent applying the active tool (if another tool is selected) while moving bidirectional handles


dannyrb:
@galelis, if this has your approval, I'm fine with a merge. I haven't touched the bi-directional tool, so you're the better reviewer here.

@evren217, thanks for using the new `semantic-release` syntax and for this PR! 🚀

galelis:
LGTM! 

@dannyrb why is it merge blocked ? unit tests?

dannyrb:
@galelis, cornerstone-tools has a "team" of users authorized to merge to `master` (as it can automate a new release). Still feeling out the best way to empower users, but also be more deliberate about who can do what.

I've added you to the "Release Publishers" team

galelis:
ooh didn't know about that. 
I was actually asking because of the ci tasks that was failing, thought one of those was blocking the merge.

Issue #774, opened 2019-01-27
freehand tool doesn't fire MEASUREMENT_MODIFIED event


Zaid-Safadi:
### Prerequisites

* [ ] Are you running the latest version?
Using v2.4.0
* [x] Are you reporting to the correct repository?
* [x] Did you perform a cursory search?

For more information, see the `CONTRIBUTING` guide.

### Description

The freehand tool doesn't fire the cornerstonetoolsmeasurementmodified event as the other measurement tools

Not sure if this issue is fixed in vNext.

### Steps to Reproduce

1. Register to the event element.addEventListener("cornerstonetoolsmeasurementmodified", onModified);
2. draw a freehand 
3. Event is not being called.

**Expected behavior:** 
The event "cornerstonetoolsmeasurementmodified" be fired and the handler is called.

**Actual behavior:**
Event is not fired.

Zaid-Safadi:
This is fixed for v2 in branch v2.4.x.
I will leave it open to apply the fix for v3

JamesAPetts:
Hey @Zaid-Safadi , if I get time tomorrow I can do this for v3/master :).

Zaid-Safadi:
That would be really nice @JamesAPetts ! please let me know know if there is anything I can help with. Thank you

dannyrb:
Fixed in 3.10

PR #865, opened 2019-03-04
add firing measurement modified event

sisobus:

issue #774 
This is fixed with #786 in version 2.4.x. But not yet in version 3.
I fix this issue in version3 because i need this feature. by reference to #786.

The freehand tool doesn't fire the cornerstonetoolsmeasurementmodified event as the other measurement tools

The event "cornerstonetoolsmeasurementmodified" be fired and the handler is called.


src/tools/annotation/FreehandMouseTool.js
```diff
@@ -980,6 +992,8 @@ export default class FreehandMouseTool extends BaseAnnotationTool {
     }
 
     external.cornerstone.updateImage(element);
+
+    this.fireModifiedEvent(element, data);
```
JamesAPetts: Thanks for adding these event fires, looks good to me.

src/tools/annotation/FreehandMouseTool.js
```diff
@@ -810,6 +813,13 @@ export default class FreehandMouseTool extends BaseAnnotationTool {
     external.cornerstone.updateImage(eventData.element);
   }
 
+  /**
+   *
+   */
+  _measurementModifyCallback(evt) {
```
JamesAPetts: Adding a callback and adding an internal `MEASUREMENT_MODIFIED` event listener throughout this code would only make sense if the callback does something. If you intend to add some functionality to this callback. I'm happy to wait and re-review.

However, if its just being left here as placeholder I see no benefit in adding the callback and associated `addEventListener`/`removeEventListener` commands at this time.

src/tools/annotation/FreehandMouseTool.js
sisobus: Thanks.

src/tools/annotation/FreehandMouseTool.js
sisobus: I think so.

So I will delete this callback and EventListeners. thanks for review.

sisobus:
I think it is done. Could you review this?

dannyrb:
@JamesAPetts, I think I've resolved the CI issue. Please feel free to merge this if it has your green light.

PR #866, opened 2019-03-04
Fix bug mouseup buttons

sisobus:

When I edited freehand tool, it didn't call endDrawing because eventData.buttons always return zero in _editMouseUpCallback function.

MouseEvent.buttons in Web API return remaining masks of button clicking now. For example, when I clicked left button and right button, If I take my hand off left button, MouseEvent.buttons return 2.

I solved this issue using MouseEvent.which in mouseup callback.

Freehand edit never done.

Freehand edit done successfully. and eventData.buttons return correct value as intended.


JamesAPetts:
Hey, could we deal with #865 first? Since this builds on top of it and seems to have a lot of the same code.

In general this is very interesting:

"MouseEvent.buttons in Web API return remaining masks of button clicking now. For example, when I clicked left button and right button, If I take my hand off left button, MouseEvent.buttons return 2."

@dannyrb perhaps we can build a more general solution to this?

sisobus:
This bug fixed. I think the code below is removed in latest code. So, I think this issue could be closed.


```javascript
// src/tools/annotation/FreehandMouseTool.js
if (!this.options.mouseButtonMask.includes(eventData.buttons)) {
   return;
}
```

PR #873, opened 2019-03-05
Fix all eslint warnings

mikehazell:
> ^ How can I find these guidelines?

The goal of this update is to clear out the eslint warning noise and to make lint-ing a more useful / useable tool and enable making the eslint config more strict again.

The vast majority of warnings are `warning  Missing JSDoc @returns for function` which is easily fixed by adding `@returns {void}` as most of these method return nothing.

So far I've gotten the number of warnings down from `224` to `133`.

I wanted to get this PR up before getting all the way to make sure this work is useful to you.

Before:
image

* **What is the new behaviour (if this is a feature change)?**

Current Progress:
image

No breaking changes here 🤞.

The caveat to this is that when eslint is telling me that code is not being used I am deleting that code.
There are a few places where there may be side effects that are hard for me to know about. I'll call these out here as they come up.

TODO: add list of potentially dangerous deletions.

* **Other information**:

### DIFF
--- a/src/store/modules/brushModule.js
+++ b/src/store/modules/brushModule.js
@@ -19,6 +19,7 @@ const setters = {
    * Sets the brush radius, account for global min/max radius
    *
    * @param {number} radius
+   * @returns {void}
    */
   radius: radius => {
     state.radius = Math.min(Math.max(radius, state.minRadius), state.maxRadius);
@@ -29,6 +30,7 @@ const setters = {
    * Sets the brush color map to something other than the default
    *
    * @param  {Array} colors An array of 4D [red, green, blue, alpha] arrays.
+   * @returns {void}
    */
   brushColorMap: colors => {
     const colormap = external.cornerstone.colors.getColormap(state.colorMapId);
@@ -126,39 +128,18 @@ const getters = {
 /**
  * EnabledElementCallback - Element specific initilisation.
  * @public
- * @param  {Object} enabledElement  The element on which the module is
+ * @param  {Object} enabledElement - The element on which the module is
  *                                  being initialised.
+ * @returns {void}
  */
 function enabledElementCallback(enabledElement) {
   setters.elementVisible(enabledElement);
 }
 
 /**
- * RemoveEnabledElementCallback - Element specific memory cleanup.
- * @public
- * @param  {Object} enabledElement  The element being removed.
- */
-function removeEnabledElementCallback(enabledElement) {
-  if (!external.cornerstone) {
-    return;
-  }
-
-  const cornerstoneEnabledElement = external.cornerstone.getEnabledElement(
-    enabledElement
-  );
-
-  const enabledElementUID = cornerstoneEnabledElement.uuid;
-  const colormap = external.cornerstone.colors.getColormap(state.colorMapId);
-  const numberOfColors = colormap.getNumberOfColors();
-
-  // Remove enabledElement specific data.
-  delete state.visibleSegmentations[enabledElementUID];
-  delete state.imageBitmapCache[enabledElementUID];
-}
-
-/**
  * OnRegisterCallback - Initialise the module when a new element is added.
  * @public
+ * @returns {void}
  */
 function onRegisterCallback() {
   _initDefaultColorMap();
--- a/src/store/modules/brushModule.js
+++ b/src/store/modules/brushModule.js
@@ -137,6 +137,31 @@ function enabledElementCallback(enabledElement) {
 }
 
 /**
+ * RemoveEnabledElementCallback - Element specific memory cleanup.
+ * @public
+ * @param  {Object} enabledElement  The element being removed.
+ * @returns {void}
+ */
+// TODO -> Test this before adding it to the module.
+function removeEnabledElementCallback(enabledElement) {
+  if (!external.cornerstone) {
+    return;
+  }
+
+  const cornerstoneEnabledElement = external.cornerstone.getEnabledElement(
+    enabledElement
+  );
+
+  const enabledElementUID = cornerstoneEnabledElement.uuid;
+  const colormap = external.cornerstone.colors.getColormap(state.colorMapId);
+  const numberOfColors = colormap.getNumberOfColors();
+
+  // Remove enabledElement specific data.
+  delete state.visibleSegmentations[enabledElementUID];
+  delete state.imageBitmapCache[enabledElementUID];
+}
+
+/**
  * OnRegisterCallback - Initialise the module when a new element is added.
  * @public
  * @returns {void}
--- a/src/store/setToolMode.js
+++ b/src/store/setToolMode.js
@@ -556,6 +556,7 @@ function _determineStringBindings(toolName, options, interactionTypes) {
   const globalTool = store.state.globalTools[toolName];
 
   if (globalTool) {
+    // eslint-disable-next-line new-cap
     const tool = new globalTool.tool(globalTool.configuration);
 
     tool.supportedInteractionTypes.forEach(interactionType => {
--- a/src/tools/FreehandSculpterMouseTool.js
+++ b/src/tools/FreehandSculpterMouseTool.js
@@ -94,12 +94,11 @@ export default class FreehandSculpterMouseTool extends BaseTool {
    * Event handler for MOUSE_DOWN.
    *
    * @param {Object} evt - The event.
+   * @returns {boolean}
    */
   preMouseDownCallback(evt) {
     const eventData = evt.detail;
     const config = this.configuration;
-
-    const imageNeedsUpdate = false;
 
     if (config.currentTool === null) {
       this._selectFreehandTool(eventData);
@@ -117,6 +116,7 @@ export default class FreehandSculpterMouseTool extends BaseTool {
    *
    * @event
    * @param {Object} evt - The event.
+   * @returns {void}
    */
   activeMouseDragCallback(evt) {
     const config = this.configuration;
@@ -146,6 +146,7 @@ export default class FreehandSculpterMouseTool extends BaseTool {
    * Event handler for MOUSE_UP during the active loop.
    *
    * @param {Object} evt - The event.
+   * @returns {void}
    */
   activeMouseUpCallback(evt) {
     const eventData = evt.detail;
@@ -174,6 +175,7 @@ export default class FreehandSculpterMouseTool extends BaseTool {
    *
    * @private
    * @param  {type} evt description
+   * @returns {void}
    */
   _renderHoverCursor(evt) {
     const eventData = evt.detail;
@@ -242,6 +244,7 @@ export default class FreehandSculpterMouseTool extends BaseTool {
    *
    * @public
    * @param {Object} evt - The event.
+   * @returns {void}
    */
   newImageCallback(evt) {
     this._deselectAllTools(evt);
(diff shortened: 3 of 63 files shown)


src/store/modules/brushModule.js
```diff
@@ -126,39 +128,18 @@ const getters = {
...
-
-  const enabledElementUID = cornerstoneEnabledElement.uuid;
-  const colormap = external.cornerstone.colors.getColormap(state.colorMapId);
-  const numberOfColors = colormap.getNumberOfColors();
-
-  // Remove enabledElement specific data.
-  delete state.visibleSegmentations[enabledElementUID];
-  delete state.imageBitmapCache[enabledElementUID];
-}
-
```
dannyrb: @JamesAPetts, do you use this anywhere?

src/tools/base/BaseBrushTool.js
```diff
@@ -50,7 +50,9 @@ class BaseBrushTool extends BaseTool {
    *
    * @abstract
    * @param {Object} evt - The event.
+   * @returns {void}
    */
+  // eslint-disable-next-line no-unused-vars
   renderBrush(evt) {
```
dannyrb: Thank you for using `// eslint-disable-next-line no-unused-vars` here.
It's really helpful to know, in these situations, what is passed to the override-able class methods

dannyrb: 🤔 I wonder if there is an eslint rule that covers, or creates an exception, for cases like this.

src/tools/base/BaseBrushTool.js
```diff
@@ -441,16 +461,6 @@ class BaseBrushTool extends BaseTool {
...
-    const enabledElement = external.cornerstone.getEnabledElement(element);
-
-    const image = enabledElement.image;
-
-    const dim = {
-      xy: image.columns * image.rows,
-      z: image.rows,
-      xyz: image.columns * image.rows * imageIds.length,
-    };
-
```
dannyrb: @JamesAPetts did you have plans for this?

src/util/brush/getCircle.js
```diff
@@ -8,7 +8,7 @@
  * @param  {number} columns    The number of columns.
  * @param  {number} [xCoord = 0] The x-location of the center of the circle.
  * @param  {number} [yCoord = 0] The y-location of the center of the circle.
- * @returns {number[][]}        Array of pixels contained within the circle.
+ * @returns {Array.number[]}        Array of pixels contained within the circle.
```
dannyrb: 🤔 This is probably correct, but it's not what I would have guessed

src/util/debounce.js
```diff
@@ -188,7 +189,7 @@ function debounce(func, wait, options) {
     const isInvoking = shouldInvoke(time);
 
     lastArgs = args;
-    lastThis = this;
+    lastThis = this; // eslint-disable-line consistent-this
```
dannyrb: Weird, but we stole this from a _underscore_ or something similar.

dannyrb:
@mikehazell, changes like this are very, very welcome, and incredibly helpful. Let's leave this one as is, especially if the other changes are less of a given. THANK YOU for taking the time to clean things up and make this PR.

If @JamesAPetts signs off, we can merge this ( and do anything else in a different PR )

mikehazell:
Can I squeeze in one more commit. It should fix the currently broken build of the API documentation.

mikehazell:
Ok, I couldn't help myself (OCD FTW) so I've gone through the rest as well. Hope that was ok.

Now down to 4 warnings for "defined/assigned but never used".

These look like WIP or future api so I didn't want to delete them.

image

src/tools/FreehandSculpterMouseTool.js
```diff
@@ -520,10 +532,12 @@ export default class FreehandSculpterMouseTool extends BaseTool {
...
    *
-   * @param  {type} i                     The index to check.
-   * @param  {type} points                The points.
-   * @param  {type} indiciesToInsertAfter The working list of indicies to
-   *                                      insert new points after.
+   * @param {number} i - The index to check.
+   * @param {Object} points - The points.
+   * @param {Array} indiciesToInsertAfter - The working list of indicies to insert new points after.
+   * @param {number} maxSpacing
+   * @modifies indiciesToInsertAfter
```
JamesAPetts: `@modifies` isn't part of the `JSDoc` tag list yet [docs](http://usejsdoc.org/). Its not being released until 3.6.0. I'm not saying we shouldn't preemptively tag stuff now, but it feels a bit odd to have a tag that doesn't do anything. What do you think @dannyrb @swederik ?

src/store/modules/brushModule.js
JamesAPetts: We are currently not using it, but we should. Instead of deleting this we should add it to the export so its called by `store/internals/removeEnabledElement.js`. Then test it of course.

If OP is willing to undo the deletion of these lines, I'll hook it up and test it in a coming patch.

src/store/setToolMode.js
```diff
@@ -556,6 +556,7 @@ function _determineStringBindings(toolName, options, interactionTypes) {
   const globalTool = store.state.globalTools[toolName];
 
   if (globalTool) {
+    // eslint-disable-next-line new-cap
```
JamesAPetts: My opinion of this may differ to @dannyrb 's, but here is my argument: It is a class constructor, even if its a reference, so `GlobalTool` should be capitalized, instead of adding a linting exception. There are likely a few other places in the codebase where this is true.

src/tools/base/BaseBrushTool.js
JamesAPetts: Might be cleaner to add `// eslint no-unused-vars: 0` before the block of `abstract` methods, then after the block turn the rule back on with `// eslint no-unused-vars: 1`. This will remove the bloat at the top of each abstract method, and also forces us to keep the abstract methods nicely organized together at the top of the class.

src/tools/base/BaseBrushTool.js
JamesAPetts: I plan to tidy up this functionality, but I'm happy for these lines to be deleted as they aren't used. This is something I wrote at the Project Week hackathon and cheekily merged to master.

src/util/brush/getCircle.js
JamesAPetts: The `.` syntax is soon to be (or already is?) deprecated in  `jsdoc3`as far as I know, its `google/closure-compiler` syntax. You can see arrays defined as `string[]` [here](http://usejsdoc.org/tags-param.html), and here is a note that `number[][]` is standards compliant: https://github.com/jsdoc3/jsdoc/issues/705

JamesAPetts: Here is another more recent example of Hegemonic ignoring close-compiler syntax (by ignoring `dandv`'s suggestion and saying there is no way to document individual array elements in jsdoc3 https://github.com/jsdoc3/jsdoc/issues/1073

I'm happy to discuss any of this and don't mean to strong-arm. The docs for params are pretty bare, but `type[]` and `type[][]` do seem to be the standard.

src/store/modules/brushModule.js
JamesAPetts: Or to keep the linter happy, comment out the lines and add `// TODO -> Test this before adding it to the module.`

src/util/brush/getCircle.js
mikehazell: I don't have a great deal of experience with JSDoc so happy for whatever you say is the right thing to do here. The linter was complaining that `number[][]` was a "JSDoc syntax error". Hence the change.

JamesAPetts: Aha, that's interesting. The two aren't necessarily in sync with each other. If eslint prefers your syntax, and the API docs still autogenerate fine... maybe we should use that instead. Like I said the docs aren't super clear on this. I'm not opinionated either way, lets wait for @dannyrb  to chime in.

src/store/modules/brushModule.js
mikehazell: I've reverted this change.

src/store/setToolMode.js
mikehazell: Capitalising `GlobalTool` won't fix this. It's the `.tool` property on `globalTool` which is the constructor.

Could rewrite as follows to avoid the linter error.
```js
const globalTool = store.state.globalTools[toolName];
if (globalTool) {
   const { tool: Tool, configuration } = globalTool;
   const tool = new Tool(configuration);

  // ... 
```

The other option would be to change every usage of `globalTools.tool`  to `globalTools.Tool`.

src/tools/base/BaseBrushTool.js
mikehazell: Maybe it's cleaner to document this with [`@abstract`](http://usejsdoc.org/tags-abstract.html) and an `@example` which lists the arguments which will be passed, rather than having code that never runs.

src/tools/FreehandSculpterMouseTool.js
mikehazell: Ah yeah, I might have gotten that from the closure compiler docs. I'll take it out.

mikehazell: So I have removed that line, but it turns out this is in a bunch of places in this file but also `FreehandMouseTool.js` and `BaseBrushTool.js`.

dannyrb: I'm not opposed to non-standard tags. `Docma`, what we use to generate API docs, supports non-standard tags, or we can simply strip them out when generating.

If it adds helpful information, and we use the tag consistently, then it feels like a win to me.

src/tools/base/BaseBrushTool.js
dannyrb: @mikehazell, that sounds like a great solution. Does the linter yell at us for defining parameters not included in the signature? If not, then I like this approach.

src/util/brush/getCircle.js
dannyrb: @mikehazell, I'm okay with keeping your changes here "as is" for now. We'll want to create a follow-up research item on this one. 

If `Docma` is happy with the new syntax, but the linter is not, we may need a custom lint rule/modification to acoomodate. If `Docma` is also unhappy, then I see no immediate need to change, but I'd want to create a related issue in `Docma`'s repo.

src/tools/base/BaseBrushTool.js
JamesAPetts: Yeah if the linter is happy with that I also love that solution.

src/tools/FreehandSculpterMouseTool.js
JamesAPetts: @dannyrb lets just review if we can make Docma happy with this sometime in the future and then we'll add it back in if it is?

src/store/setToolMode.js
dannyrb: I'm okay with your current fix or the workaround. While `Tool` should probably be capitalized, let's file that away mentally as a breaking change for a different day.

JamesAPetts: Ahh good catch, I missed that. I'm happy with how you've done it for now. Its an improvement over what we had before. I'd like to do a sweeping review of constructors in the future though.

src/tools/FreehandSculpterMouseTool.js
dannyrb: I was responding more to @mikehazell's note about its usage in other places. No need to strip all of those out. If this has already been removed from this particular block, that's fine.

https://onury.io/docma/api#Docma~BuildConfiguration
`jsdoc.allowUnknownTags`

JamesAPetts:
Thanks! Lets merge this and leave all the discussion points to future work.

dannyrb:
@all-contributors please add @mikehazell for cleaning up our docs and code! 🎉

PR #887, opened 2019-03-15
SVG Cursors

JamesAPetts:
A new stab at the SVG Cursor project.

## Cursors

SVG Cursors are off by default. They are turned on via init by passing the configuration option: `showSVGCursors = true`.

SVG cursors can now be created by
```js
this.svgCursor = new MouseCursor(svgString);
```
Where the `svgString` is just a string containing the cursor in SVG format, the mouse cursor will be placed in a standardized position.

For simple tools this is done in `BaseTool`, by passing an `svgCursor` string to the constructor.

For complex multi-part tools you are free to make as many `MouseCursor`s as you wish and assign these at different parts of the drawing loop within the tool's annotation cycle.

### svgCursor

If `svgCursor` is set, the mouse cursor will be changed to the set value when the tool is `active` and `mouseButtonMask === 1`. I.e. the cursor for the primary mouse click is shown. Touch tools are unaffected.

SVG cursors are stored in `assets/svgCursors.js`, you can alter their `mousePoint` property to change the mouse position to either the top left (e.g. magnify) or center (e.g. pan) of the icon.

### hasCanvasCursor

If `hasCanvasCursor` is true, the canvas will re-render on `MOUSE_MOVE` whilst the tool is active. This is intended for tools that need to render a cursor on the canvas, if the cursors could potentially be greater than 128x128 pixels in size. Canvas cursors should be used sparingly, and SVG cursors are preferable wherever possible.

### hideDefaultCursor

If `hideDefaultCursor` is true and `svgCursor` is unset, the default cursor will be hidden when hovering over the cornerstone enabled element. You may want to do this if using canvas cursors, but as canvas cursors can be complex and dynamic, this is optional. See `FreehandMouseSculpterTool` to see an example of this being switched on/off during tool usage.

## Tools with cursors

I'm making simple icons for now, we plan to put them in a bounding box and have a crosshair on the top left, like my amazing sketch here shows:

untitled(1)

- [x] AngleTool
- [x] ArrowAnnotateTool
- [x] BidirectionalTool
- [x] CobbAngleTool
- [x] EllipticalRoiTool
- [x] FreehandMouseTool
- [x] LengthTool
- [x] ProbeTool
- [x] DragProbeTool
- [x] RectangleRoiTool
- [x] TextMarkerTool
- [x] BrushTool
- [x] CrosshairsTool
- [x] EraserTool
- [x] FreehandSculpterMouseTool
- [x] MagnifyTool
- [x] PanTool
- [x] RotateTool
- [x] StackScrollTool
- [x] WwwcRegionTool
- [x] WwwcTool
- [x] ZoomTool


dannyrb:
Hey, @JamesAPetts! I think this is a good start and a smart approach.

### Global Enabled / Disabled

We may want to default to _not_ showing cursors, but allowing them to be enabled when calling `init` here:

- [Where we set configuration](https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/init.js#L24-L27)
- [Where the defautls live](https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/store/index.js#L17-L21)

In a major version change, we can update them to be on by default.

### Cursors

- We may need to make sure cursors with ambiguous click locations have an `x` in the top left to indicate the click location. For consistency, we may want all cursors to have the `x`?
- Some cursors + tool combos may require some extra love. For example, rotate would benefit from marking the point we're rotating around (actually... This may not be a cursor issue).

JamesAPetts:
Hey @dannyrb, yeah I agree with everything you say. I'm first going to try to get _a_ cursor for every tool, then we can add it as off by default, then we can work a bit more on standardization of cursors.

This SVG stuff isn't too bad, had some free time so trying to learn it a little today, at least to the point where I can rip of Font Awesome icons and compose them to make something usable.

JamesAPetts:
> rotate would benefit from marking the point we're rotating around (actually... This may not be a cursor issue).

I think that's a `canvasCursor` issue. It'd indeed be nice to put a crosshair in the center of the image whilst rotating.

src/eventDispatchers/mouseEventHandlers/mouseMove.js
```diff
@@ -34,9 +34,9 @@ export default function(evt) {
 
   let imageNeedsUpdate = false;
 
-  // If any tools are active, check if they have a cursor, and if so update image.
+  // If any tools are active, check if they have a canvas cursor, and if so update image.
   if (activeTools.length > 0) {
-    imageNeedsUpdate = activeTools.some(tool => tool.hasCursor);
+    imageNeedsUpdate = activeTools.some(tool => tool.hasCanvasCursor);
```
dannyrb: This may not be true if we have svg cursors enabled, and the tool has an svg cursor that takes precedence over its canvas cursor?

src/store/setToolCursor.js
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Creates an SVG Cursor for the target element
+ *
+ * @param {MouseCursor} svgCursor - The cursor.
+ */
+function setToolCursor(element, svgCursor) {
+  if (!globalConfiguration.state.showSVGCursors) {
+    return;
+  }
+  console.log('attempting to set cursor...');
```
dannyrb: We really should look into a logging library where we can set log levels in `init`

src/eventDispatchers/mouseEventHandlers/mouseMove.js
JamesAPetts: Perhaps, currently if something has a canvas cursor, it just means we call a refresh on mouse move. A tool could have both, but if the canvas cursor is present, the screen needs to be refreshed?

A canvas cursor can be anything dynamic/animated (i.e. brush/freehandSculpter cursors), or any cursor that needs to be > 128x128 for any reason. In some of these cases you might want an SVG cursor as well.

src/store/setToolCursor.js
JamesAPetts: Yeah that'd be a good idea.

src/tools/CrosshairsTool.js
```diff
@@ -189,3 +192,24 @@ export default class CrosshairsTool extends BaseTool {
     });
   }
 }
+
+const crosshairsCursor = {
+  svgGroupString: `<path fill="ACTIVE_COLOR" d="M1325 1024h-109q-26 0-45-19t-19-45v-128q0-26
```
dannyrb: Instead of keeping the SVG path data in their respective tools, can we put them somewhere co-located that we can import from? It might make it easier for tools to share cursors; especially 3rd party ones (or to add to the list of usable cursors).

JamesAPetts: Yeah this makes sense. We should probably also instantiate them in that file, and pass an actual `MouseCursor` instance to the constructor of `BaseTool`, rather than passing a string and then building the `MouseCursor` inside.

Then to override the cursor you pass a new `MouseCursor` when instantiating it. Same for third-party tools.

mikehazell:
I love this ❤️
- Awesome that cursors can be overridden by the end user.
- Would be nice to have the "original" svgs available for use in a toolbar, or maybe a method like the `MouseCursor. _generateSVGString` for getting an SVG out without the crosshairs.

dannyrb:
Good note, @mikehazell!

@JamesAPetts, I've started on the docs + example for this. Can you help me spell out what each should cover?:

_Examples:_

- [ ] How to globally enable/disable cursors
- [ ] How to add/set a tool's cursor
    - [ ] SVG Cursor
    - [ ] Canvas Cursor
    - [ ] Default Cursor
    - [ ] No Cursor

_Docs:_ ?? - I'll have a better idea after I take a stab at the examples

JamesAPetts:
Hey guys,

@mikehazell, thanks for the love and feedback, I think a method to grab the original SVGs out makes total sense. I'm going to work on exposing this stuff a little better this morning.

@dannyrb , when you want to go over the examples/docs ping me on slack and we can go through it together.

JamesAPetts:
Hey all,

@mikehazell I've added a getter to the `MouseCursor` class, such that `mouseCursorInstance.iconSVG` returns the icon itself as an SVG.
@dannyrb The cursors are now stored in `tools/cursors`, and added to the lib are the entries:
- `tools/cursors`, which returns an object containing all the cursors, should you wish to get their SVG's or use them in other tools.
- `tools/cursors/MouseCursor`, which returns the `MouseCursor` class.

src/tools/cursors/index.js
```diff
@@ -0,0 +1,381 @@
...
+  crosshairsCursor,
+  eraserCursor,
+  magnifyCursor,
+  panCursor,
+  rotateCursor,
+  stackScrollCursor,
+  wwwcRegionCursor,
+  wwwcCursor,
+  zoomCursor,
+};
```
dannyrb: I don't think the exported constants are hoisted above the default export. Because this is what we're tying to `lib`'s export, these come back as `undefined` for me:

image

dannyrb:
image

src/tools/base/BaseTool.js
```diff
@@ -43,14 +44,19 @@ class BaseTool {
...
     // Or via a tool's option's setter
     this._options = {};
     // Configuration is set at tool initalization
     this._configuration = Object.assign({}, configuration);
 
-    // True if tool has a custom cursor, causes the frame to render on every mouse move when the tool is active.
-    this.hasCursor = false;
+    // True if tool has a canvas cursor, causes the frame to render on every mouse move when the tool is active.
+    this.hasCanvasCursor = false;
+    this.hideDefaultCursor = false;
```
dannyrb: Would we want these values to be set via configuration on init? Not setting them based on incoming config means we can't set them as part of the `globalToolSync`.

src/tools/base/BaseTool.js
```diff
@@ -26,6 +26,7 @@ class BaseTool {
     configuration,
     supportedInteractionTypes,
     mixins,
+    svgCursor,
```
dannyrb: I think this will be the first `class` we pass in as part of configuration. Shouldn't be a big deal, but it might add a hurdle for anyone wanting to serialize config. I double checked, but we should be good with the current `globalSync` logic here:

https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/store/addTool.js#L62-L82

src/tools/cursors/MouseCursor.js
```diff
@@ -0,0 +1,66 @@
...
+    const activeColor = toolColors.getActiveColor();
+
+    return `
+    <svg
+    data-icon="cursor" role="img" xmlns="http://www.w3.org/2000/svg"
+    width="48" height="48" viewBox="0 0 48 48"
+  >
+    <g>
+      <path stroke="${activeColor}" d="M8 16L8 0" id="dUU9BLALA"></path>
+      <path stroke="${activeColor}" d="M16 8L0 8" id="b9wR4oWz"></path>
```
dannyrb: Not necessary for the PR, but it might be nice to override the default pointer.

src/tools/cursors/MouseCursor.js
```diff
@@ -0,0 +1,66 @@
+import toolColors from '../../stateManagement/toolColors.js';
+
+const mousePosition = '8 8';
```
dannyrb: Can we split these into x/y?

src/tools/cursors/MouseCursor.js
```diff
@@ -0,0 +1,66 @@
...
+      width="32" height="32" viewBox="0 0 ${viewBox.x} ${viewBox.y}"
+    >
+      ${this.svgGroupString.replace(
+        /ACTIVE_COLOR/g,
+        `${toolColors.getActiveColor()}`
+      )}
+    </svg>`;
+
+    return new Blob([svgString], { type: 'image/svg+xml' });
+  }
```
dannyrb: I found myself wanting to be able to grab the `string` SVG for the icon and the icon + pointer. Ideally, w/ default values applied, but with the ability to pass in height, width, and color.

src/tools/cursors/MouseCursor.js
```diff
@@ -0,0 +1,66 @@
...
+};
+
+export default class MouseCursor {
+  constructor(svgGroupString, options) {
+    this.svgGroupString = svgGroupString;
+    this.options = options;
+
+    const svgString = this._generateSVGString();
+
+    this.blob = new Blob([svgString], { type: 'image/svg+xml' });
```
dannyrb: This prevents recalc, but it would be nice if we could force an update to set the blob to a version with an updated size/color.

dannyrb:
Feel free to push back on any of my notes. Playing with the examples page was helpful for thinking up better API usage questions. I still have on my list to try out `canvasCursor`

src/tools/cursors/MouseCursor.js
JamesAPetts: Yeah I guess we can and then we can construct it as needed when we generate the cursor.

src/tools/base/BaseTool.js
JamesAPetts: I originally passed a string, but it made more sense from the a plugin dev point of view to pass the class instance.

Also the `svgCursor` is just an object. Classes are just syntactic sugar in ES6.

src/tools/base/BaseTool.js
JamesAPetts: Hmm, any tool so far that has a canvas cursor requires it operationally. At some level its just a flag to signal that the screen should re-render on mouse move. `hasCanvasCursor` existed prior to `svgCursor`s, so perhaps its should just be named 'updateOnMouseMove`?

I think the tool should hard set this based on its requirements. I'm happy to discuss, but being able to switch the brush cursor or the sculpter cursor off just seems like a bad idea to me.

src/tools/cursors/index.js
JamesAPetts: Aha, good catch. Thanks for fixing it.

src/tools/cursors/MouseCursor.js
JamesAPetts: Yeah I've thought about this also. It might be nice to have a global setter for the pointer, and have all svg update as needed.

src/tools/cursors/MouseCursor.js
JamesAPetts: We can just add more getters for this.

src/tools/cursors/MouseCursor.js
JamesAPetts: Yeah, I guess the svg generation is pretty light...

I wonder how bad it is to just implement something like this:

get blob() {
  return this._generateSVGString();
}

(or put the contents of _generateSVGString into the blob getter.)

It'd rebuild the svg element every time you switch your primary tool, but that's likely not too expensive, and it'd at least be dynamic to state changes

JamesAPetts: Did this in latest commit.

src/tools/cursors/MouseCursor.js
JamesAPetts: And new getters there shall be.

src/tools/cursors/MouseCursor.js
JamesAPetts: Done.

src/tools/cursors/MouseCursor.js
JamesAPetts: Done.

src/store/setToolMode.js
```diff
@@ -56,12 +63,39 @@ const setToolActiveForElement = function(
...
 };
 
+function _setToolCursorIfPrimary(element, options, tool) {
+  let mouseButtonMask;
+
+  if (typeof options === 'number') {
+    mouseButtonMask = options;
+  } else {
+    mouseButtonMask = options.mouseButtonMask;
+  }
```
dannyrb: I think @mikehazell's introduced a method that normalizes the options higher up the chain, so we can check this in a consistent way.

IE. `options.mouseButtonMask` will always be an array after being normalized.

src/eventDispatchers/mouseEventHandlers/mouseMove.js
```diff
@@ -34,9 +34,9 @@ export default function(evt) {
 
   let imageNeedsUpdate = false;
 
-  // If any tools are active, check if they have a cursor, and if so update image.
+  // If any tools are active, check if they have a canvas cursor, and if so update image.
   if (activeTools.length > 0) {
-    imageNeedsUpdate = activeTools.some(tool => tool.hasCursor);
+    imageNeedsUpdate = activeTools.some(tool => tool.updateOnMouseMove);
```
dannyrb: I wonder if there's a way to do this without executing _all_ of the render logic. Some of the ROI tools are a bit expensive, even when throttled. It would be nice if we could magically only render an overlay layer with the cursor.

Maybe a second canvas with a transparent background and no pointer events? Sounds like an enhancement for another day.

src/tools/cursors/MouseCursor.js
```diff
@@ -0,0 +1,77 @@
+import toolColors from '../../stateManagement/toolColors.js';
+import { modules } from '../../store/index.js';
+
+const cursorModule = modules.cursor;
+
+export default class MouseCursor {
+  constructor(iconGroupString, options) {
+    this.iconGroupString = iconGroupString;
```
dannyrb: Technically, I think this can be any path data.

src/tools/cursors/MouseCursor.js
```diff
@@ -0,0 +1,77 @@
...
+    const svgString = this._generateIconSVGString();
+
+    return new Blob([svgString], { type: 'image/svg+xml' });
+  }
+
+  get iconWithPointerSVG() {
+    const svgString = this._generateIconWithPointerSVGString();
+
+    return new Blob([svgString], { type: 'image/svg+xml' });
+  }
```
dannyrb: If we only update this property's value when the `svgString` or `options` change, it might be a bit more performant.

src/tools/cursors/MouseCursor.js
```diff
@@ -0,0 +1,77 @@
...
+
+  get iconSVGString() {
+    return this._generateIconSVGString();
+  }
+
+  get iconWithPointerString() {
+    return this._generateIconWithPointerSVGString();
```
dannyrb: Can we update the string methods to take optional `options` object (default's to internal values)? That way we can grab an icon without updating the properties used by an active cursor.

IE. I want 3 or 4 display representation w/ different values, but the SVG Cursor on the canvas should continue to look the same.

src/tools/cursors/index.js
```diff
@@ -0,0 +1,384 @@
+import MouseCursor from './MouseCursor.js';
+
+export const angleCursor = new MouseCursor(
```
dannyrb: A tiny legend of values like `ACTIVE_COLOR` at the top of this file, and I'm loving this.

dannyrb:
@JamesAPetts, this keeps getting better and better ^_^

I think I've hit on all of the edge cases I can think of at this point. If we're in agreement, I can knock out the unit test, example, and doc updates that should accompany this. LMK

src/store/setToolMode.js
mikehazell: @dannyrb That change hasn't been merged (#903) , also the normalisation happens inside `setToolModeForElement` which is called after this method.

src/store/setToolMode.js
```diff
@@ -56,12 +63,39 @@ const setToolActiveForElement = function(
...
 
   // Resume normal behavior
   setToolModeForElement('active', null, element, toolName, options);
 };
 
+function _setToolCursorIfPrimary(element, options, tool) {
+  let mouseButtonMask;
+
+  if (typeof options === 'number') {
+    mouseButtonMask = options;
```
mikehazell:
```suggestion
    mouseButtonMask = [options];
```
You're expecting `mouseButtonMask` to be an array on line 88.

src/store/setToolMode.js
mikehazell: And in this case, there's no guarantee that the `options.mouseButtonMask` being passed in is an array either.

src/tools/cursors/index.js
```diff
@@ -0,0 +1,384 @@
...
+    />`,
+  {
+    viewBox: {
+      x: 640,
+      y: 512,
+    },
+  }
+);
+
+export default {
```
mikehazell: Is this `export default` necessary? Every cursor is being exported individually and if a consumer of this wants every cursor they can do this:

```js
import * as cursors from './cursors';
```

JamesAPetts:
@dannyrb @mikehazell Thanks a lot for the feedback! I should be able to get on this either tomorrow or early next week.

src/eventDispatchers/mouseEventHandlers/mouseMove.js
JamesAPetts: Yeah, we should eventually get around to this.

The brush used to the cornerstone layers api for both the brush cursor and the mask, but it was super clunky to use on an application level/ not just in a POC. If we made a corerstoneTools specific layer I'm sure we could do this in a clean way that doesn't require any additional setup when enabledElements are pulled up/torn down.

This might be a nice thing to do at project week.

src/tools/cursors/MouseCursor.js
JamesAPetts: ¯\_(ツ)_/¯.

I more mean the "icon itself that goes to the bottom-right of the pointer. If you can think of a more descriptive name, happy to change it.

src/tools/cursors/MouseCursor.js
JamesAPetts: Hmm yeah, maybe I'll put some caching in.

JamesAPetts: Hmm, how do we check if the tool color has changed? 🤔

src/tools/cursors/MouseCursor.js
JamesAPetts: I like this, sure.

src/store/setToolMode.js
JamesAPetts: Good catch.

src/tools/cursors/index.js
JamesAPetts: We use it in the lib to export all of the default cursors, but yeah sure, its maybe a bit superfluous. I think its only going to be used in one place, so I'll make your suggested change in the lib.

src/tools/cursors/MouseCursor.js
dannyrb: :l 
...
Shoot.

We could add an event? Or, if we're only setting the SVG cursor when the tool is activated / made primary, then we can use whatever color at that point in time and call it good enough?

JamesAPetts: > Or, if we're only setting the SVG cursor when the tool is activated / made primary, then we can use whatever color at that point in time and call it good enough?

Maybe I misunderstand, but isn't that what we are already doing? I personally thinking adding another event is a little overkill, how often is the active color changed on run time? I've generally seen it as more of a set-and-forget. Even if the active color is changed during run time, it will at least refresh on the next tool activation?

src/tools/cursors/MouseCursor.js
JamesAPetts: Done.

src/tools/cursors/index.js
JamesAPetts: Changed :).

src/store/setToolMode.js
JamesAPetts: Thanks for casting options too an array if its just a number Mike. Can `options.mouseButtonMask` be anything other than an array within the library?

src/tools/cursors/index.js
JamesAPetts: Done.

src/tools/cursors/MouseCursor.js
dannyrb: Ahh! It [looks like we are](https://github.com/cornerstonejs/cornerstoneTools/pull/887/files/cbe4317c1a774780e0a1ea0f9a72b9ee74749d33#diff-0c9bc112b60ea20dbbd1f7953c4c2c29R90). I had this weird thought that we were setting the cursor on every render (oops).

You can ignore my comments then. I was trying to optimize for a situation/issue that we don't have.

PR #888, opened 2019-03-17
chore(api-docs): output docma api docs as part of release build

dannyrb:


dannyrb:
CircleCI issues:

image

Not a good sign that it's overloaded on a Saturday evening :P

Issue #877, opened 2019-03-05
How to trigger a tool recalculation after 'addToolState' for RectangleRoiTool?


pzehle:
### Prerequisites

* [x] Are you running the latest version?
* [x] Are you reporting to the correct repository?
* [x] Did you perform a cursory search?

For more information, see the `CONTRIBUTING` guide.

### Description

I am working currently on a tool that has to calculate a volume, this means, I draw an area using the `RectangleRoiTool`, and then I have to replicate that area on X number of images. I am doing this by using `addToolState` on each image, using the `toolState` from the area I created. The problem is, when the script scrolls to the next image and adds the tool state, it doesn't recalculate the Mean and StdDev of the current image, but it is saving the data from the created area. Is there an option to recalculate the state once it is added? Specifically for the `RectangleRoiTool`.

Thanks!

### Steps to Reproduce

1. Draw a rectangle using `RectangleRoiTool`.
2. Save the tool state on a variable.
3. Add the saved tool state on any other image. The calculations are still the ones from the variable, not the ones that should be from this new image.

**Expected behavior:** Once the tool state is added, it should recalculate the information from this new image.

**Actual behavior:** After setting the tool state, the information from the saved state is static, no recalculation is being triggered.

dannyrb:
@pzehle, I don't believe this feature exists. I would be open to a PR that updates the `BaseAnnotationTool` to accomplish this. Something like...

`forceRecalc` that has a default implementation of setting the tool's measurement as "dirty" and triggering a re-render. This is definitely up for discussion. @JamesAPetts may have other thoughts.

Small note. If you're calculating volume, you may not want to sum area on X slides. You might be better off calculating the volume of your cube, then using that cube's dimensions to determine how to annotate each slice.

JamesAPetts:
This is an interesting problem and one I've thought of for a while. Shared data between annotations are what modules are built for, and this is a perfect use case for that.

I would just give the rects a UID, and register them to a ledger in a module. In the rects `addToolState` function I'd add a call to the module to invalidate all annotations registered in the ledger.

JamesAPetts:
You could also save the volume as a module variable, and have a getter in your render function for the toolState that fetches this, so you only have to keep track of one copy of it.

pzehle:
[...]

PR #889, opened 2019-03-18
Stack tool state api

JamesAPetts:
An API to help with development of tools that require access to `toolState` of a specific tool across an entire stack, e.g. for 3D tools. Inspired by @pzehle's issue #877.

Also provides an easy api method to merge new `toolState` to the global `toolState` cleanly, useful for IO (Perhaps this should live elsewhere?).

What:
4 methods and unit tests for each:

```js
/**
 * getToolStateForStack - Returns the toolState for every imageId in the
 * element's stack. Optionally you can filter to only return a specific tool's
 * state.
 *
 * @param  {HTMLElement} element The element to fetch the stack from.
 * @param  {string} [toolName]  A tool to filter on.
 * @returns {object[]}     The requested toolState.
 */
getToolStateForStack(
  element,
  [tool]
)

/**
 * invalidateToolStateForStack - Invalidates any toolState in the stack that
 * has an `invalidated` method.
 *
 * @param  {HTMLElement} element - The element to fetch the stack from.
 * @param  {string} [toolName] - A tool to filter on.
 * @returns {Object} A reference to the altered global toolState.
 */
invalidateToolStateForStack(
  element,
  [tool]
)

/**
 * mergeToolStateToGlobalToolState - Cleanly merges toolState to the
 * global tool state.
 *
 * @param  {object} newToolState - The new toolState to merge with the
 *                              global tool state.
 * @returns {Object} A reference to the altered global toolState.
 */
mergeToolStateToGlobalToolState(
  toolState
)

/**
 * clearToolStateForStack - clears toolstate for all imageIds in the stack.
 * If a tool is specified, then only clear that tool.
 *
 * @param  {HTMLElement} element The element to fetch the stack from.
 * @param  {string} [toolName]  A tool to filter on.
 * @returns {Object} A reference to the cleaned global toolState.
 */
clearToolStateForStack(
  element,
  [tool]
)
```

The api can be accessed by third party tools via:

```js
const {
  getToolStateForStack,
  invalidateToolStateForStack,
  mergeToolStateToGlobalToolState,
  clearToolStateForStack
} = cornerstoneTools.import('utils/stackToolStateUtils');
```


dannyrb:
I've started to play with this. Do you have a tool you'd recommend it be used for, or a simple POC that would illustrate it's benefits and interaction?

JamesAPetts:
@dannyrb , the problem is that this stuff is generally useful for larger tools which will likely only ever exists at plugins, making a demonstration within the lib a bit difficult.

The simplest POC I can think of would be something like a volume calculation that invalidates the textBox of all annotations for that tool on that element. I'll do this in peppermint when I get around to it, but its not high on our priority list here at the moment.

dannyrb:
@JamesAPetts, we'll close this for now, and revisit when we start playing with features/tools that will leverage it.

I threw a label on it to help it stand-out for when we go looking for it.

Issue #898, opened 2019-03-25
calling setToolModeForElement too many times can cause an exception


mikehazell:
### Prerequisites

* [X] Are you running the latest version?
* [X] Are you reporting to the correct repository?
* [X] Did you perform a cursory search?

### Description

**Bug**

Calling any of the `setToolMode` methods too many time with options of `{ mouseButtonMask: 1 }` will eventually cause the app to crash with a "RangeError: Invalid array length" error.

This is happening this line: https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/store/setToolMode.js#L213

```js
  // Keep the same if not an array (undefined)
  // Reset if empty array
  // Merge if array contains any bindings
  if (
    Array.isArray(options.mouseButtonMask) &&
    options.mouseButtonMask.length !== 0 &&
    Array.isArray(tool.options.mouseButtonMask)
  ) {
    options.mouseButtonMask = options.mouseButtonMask.concat(
      tool.options.mouseButtonMask
    );
  }
```

The code here is concatenating the `mouseButtonMask` array but not removing duplicates.

I'll create a PR shortly with tests showing the issue and a fix.

dannyrb:
Merged 👍

PR #899, opened 2019-03-25
Pr/set tool mode mouse button mask

mikehazell:

Bug Fix

Fixes a bug where the app can have poor performance or even crash when `setToolModeForElement` is called too many times. #898

Within `setToolModeForElement` the `options.mouseButtonMask` array is now de-duplicated.

NO

* **Other information**:
~I've used `new Set()` in here. This could cause a browser compatibility problem, depending on what csTools is supporting. I've left this as the solution as it's the cleanest option but if that is a no-go I can change it to something more es2015 safe.~
On second thought, there's no reason this needs to use `Set`s. Have updated to not do that.

### DIFF
--- a/src/store/setToolMode.test.js
+++ b/src/store/setToolMode.test.js
@@ -1,0 +1,85 @@
+import getToolForElement from './getToolForElement.js';
+import { setToolEnabledForElement } from './setToolMode.js';
+
+jest.mock('./getToolForElement.js');
+
+describe('setToolModeForElement', () => {
+  it('sets the tool.mode to enabled', () => {
+    const mergeOptions = jest.fn();
+    const testTool = {
+      mode: undefined,
+      options: { mouseButtonMask: [] },
+      supportedInteractionTypes: ['Mouse'],
+      mergeOptions,
+    };
+
+    getToolForElement.mockImplementationOnce(() => testTool);
+
+    const toolName = 'MyTool';
+    const element = {};
+
+    setToolEnabledForElement(element, toolName);
+
+    expect(testTool.mode).toBe('enabled');
+  });
+
+  it('can set the buttonMask option', () => {
+    const mergeOptions = jest.fn();
+    const testTool = {
+      mode: undefined,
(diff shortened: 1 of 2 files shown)


mikehazell:
This is the test output before the fix is applied:

image

src/store/setToolMode.test.js
```diff
@@ -0,0 +1,89 @@
...
+  it('sets the tool.mode to enabled', () => {
+    const mergeOptions = jest.fn();
+    const testTool = {
+      mode: undefined,
+      options: { mouseButtonMask: [] },
+      supportedInteractionTypes: ['Mouse'],
+      mergeOptions,
+    };
+
+    getToolForElement.mockImplementationOnce(() => testTool);
```
dannyrb: My JS Unit testing "best practice" knowledge is hazy. It might be better to setup the app state for this scenario instead of mocking implementation behavior. For example:

> We expect X to happen when the application's state is Y

Will hopefully be less brittle than:

> We expect X to happen when implementation detail returns Z

I'm very rusty here, but it may be worth discussing.

dannyrb:
@mikehazell, I'm going to merge this as-is. Thanks for taking the time to write some excellent tests and to fix this behavior. Please do take a look at my note though. I'm actively trying to improve my own testing abilities and would appreciate your thoughts.

src/store/setToolMode.test.js
mikehazell: Maybe the problem here is that the functionality that I want to test is hidden inside something bigger. So because I can't test the unit I have to test something bigger...

mikehazell: I'm going to push another commit to this branch which might shine some light on this and also expose another potential bug.

mikehazell: Also, the more I look at this, the more I wonder why merging these arrays is the right thing to do here.

mikehazell: See PR with more tests here #903

PR #903, opened 2019-03-27
WIP - Testing setToolMode.js

mikehazell:

Tests + Bug Fixes (TBD)

Related to comments on PR: #899

The tests here expose what I believe is unintentional behaviour (or at least an inconsistent API)

`setToolModeForElement` uses a private internal method `_getNormalizedOptions` to, as the name suggests, normalise the given options.

For context:
- the method `_getNormalizedOptions` only actually modifies the `mouseButtonMask` property inside options. Everything else is left as is.

These tests expose some inconsistencies around normalisation:
- it does not always return an object with the `mouseButtonMask` property
- if given an array it just returns that array, in all other cases it returns an object

image

Once implemented these tests should all pass.

Maybe ... it changes the API slightly

* **Other information**:


mikehazell:
Update - I have refactored `_getNormalizedOptions` to pass all the tests.

mikehazell:
Hmm, this implementation doesn't reflect the docs.

See the "Configuration" section of https://tools.cornerstonejs.org/anatomy-of-a-tool/#modes

Specifically the following line is not true:

```js
//  The previous `mouseButtonMask` configuration property's value is updated to `2`
csTools.setToolModeActive("ToolName", { mouseButtonMask: 2 });
```

Currently (in cornerstoneTools ^3) that config will be merged with the existing `mouseButtonMask` values.

@dannyrb - The way the docs are written makes more sense to me as a user than the current implementation does. How should I proceed?

dannyrb:
It's been a while, so let me see if I can recall the intent here (this is why descriptive tests are important). The docs highlight the most common and likely use cases:

- Setting a single `mouseButtonMask`
- Updating the `mouseButtonMask`
- Updating the tool's options without updating the `mouseButtonMask`

We probably shouldn't have, but you can accomplish this in a couple of different ways. Hence the [`_getNormalizedOptions` method](https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/store/setToolMode.js#L634-L640):

```js
// All valid
csTools.setToolModeActive("ToolName", 1);
csTools.setToolModeActive("ToolName", { mouseButtonMask: 1 });
csTools.setToolModeActive("ToolName", { mouseButtonMask: [1] });
```

The bit we've been playing with in these PRs is outside of the common use cases listed above. It's only executed when ([this condition](https://github.com/cornerstonejs/cornerstoneTools/blob/091f3abce122b18db89893ec725de431b023dfbc/src/store/setToolMode.js#L208-L212)):

- The incoming `mouseButtonMask` is a non-empty `array`
- The tool's current `mouseButtonMask` is an `array`

This allows us to set multiple bindings for the tool, and in an additive way. For example:

```js
// Bound to left click
csTools.setToolModeActive("ToolName", { mouseButtonMask: [1] });

// Bound to left click AND right click
// Allows us to not track currently set masks when "adding" or "activating" a new button/mask
csTools.setToolModeActive("ToolName", { mouseButtonMask: [2] });
```

This all gets a bit muddier when we look at the additional logic in `setToolActiveForElement`. You'll notice that we try to [resolve "input conflicts"](https://github.com/cornerstonejs/cornerstoneTools/blob/091f3abce122b18db89893ec725de431b023dfbc/src/store/setToolMode.js#L330-L361) here, so that there cannot be multiple active tools with the same binding. For example:

_Example 1_
```js
// "FirstTool" is active, and will handle events for left click
csTools.setToolModeActive("FirstTool", { mouseButtonMask: 1 });

// "FirstTool" becomes passive
// "SecondTool" is active, and will handle events for left click
csTools.setToolModeActive("SecondTool", { mouseButtonMask: 1 });
```

_Example 2_
```js
// "FirstTool" is active, and will handle events for left click and right click
csTools.setToolModeActive("FirstTool", { mouseButtonMask: [1, 2] });

// "FirstTool" is active, and will handle events for right click
// "SecondTool" is active, and will handle events for left click
csTools.setToolModeActive("SecondTool", { mouseButtonMask: 1 });
```

I think that's a good rough summary of our original intent. I can't guarantee that's what we delivered or that it's the best API interface. We need tests around this, as its a bit hard to follow and pretty core to our event handling/dispatching logic.

Hopefully that brain dump helps. Please feel free to ask questions and critique ^_^

mikehazell:
Hey @dannyrb, is this a common use case:

> to set multiple bindings for the tool, and **in an additive way**.

The **in an additive way** part really complicates the API implementation and makes it harder for the API user to "guess" what is going to happen when setting the tool mode. Especially as the "shorthand" might do something different to the longhand version.

As this was not the documented functionality, maybe we can remove it in favour of a simpler API.

It would still be possible enable additional buttons in user-land but you'd have to go the long way around.

```js
const { options } = csTools.getToolForElement(element, "MyTool"); 
options.mouseButtonMask.push(2);
csTools.setToolModeActiveForElement(element, "MyTool", options);

// Or if you want to do this for all elements
cornerstone.getEnabledElements().forEach(({ element }) => {
    const { options } = csTools.getToolForElement(element, "MyTool"); 
    options.mouseButtonMask.push(2);
    csTools.setToolModeActiveForElement(element, "MyTool", options);
});
```

dannyrb:
> As this was not the documented functionality, maybe we can remove it in favor of a simpler API.

I think I'm okay with that. I don't think many/any consumers are using the feature, and it adds unnecessary confusion/complexity. I would just want the description added as a part of `npm run commit` to be a bit more comprehensive so the change is well documented in the generated release notes.

dannyrb:
Sorry for slacking on this, @mikehazell. Super appreciate the work here. No one takes the time to improve stuff like this, but these tests are incredibly important and high value for all users 🙏

dannyrb:
@all-contributors, please add @mikehazell for these tests and fixes

Issue #901, opened 2019-03-26
console.log messages are too noisey


mikehazell:
### Prerequisites

* [x] Are you running the latest version?
* [x] Are you reporting to the correct repository?
* [x] Did you perform a cursory search?

### Description

There are many places in the codebase which are calling console.log for example on mouse interaction or on state change of a tool. This makes the console very noisy and hard to use as a tool when developing on-top of cornerstoneTools.

The following is a pretty common sight in my console.

![image](https://user-images.githubusercontent.com/814227/54964443-9d20d300-4fc0-11e9-8fe6-498e867b1a24.png)

---

### Suggestion

One solution to this would be to add a tool like [debug](https://github.com/visionmedia/debug#readme) 
 (no affiliation) for logging.

Debug is really useful for library makers to keep useful logging around to aid in debugging issues while having it disabled by default.

It also lets you "scope" logs so they can be filtered in/out and be identified by color.

Debug is enabled/disabled using a field in `localStorage` so the setting will persist through a refresh.
 
```js
// Show all debug logs - note this may include other libraries which also use debug
localStorage.setItem('debug', '*');

// Filter by scope
// - all cornerstoneTools logs
localStorage.setItem('debug', 'cornerstoneTools:*');

// - just the cornerstoneTools:annotation scope
localStorage.setItem('debug', 'cornerstoneTools:annotation');
```

Example output in a chrome console:
![Debug output example](https://user-images.githubusercontent.com/71256/29092033-b65f9f2e-7c39-11e7-8e32-f6f0d8e865c1.png)

I'd be happy to make a PR to get this started.

dannyrb:
Ah! @JamesAPetts and I were just discussing better logging practices. The current state of things is a bit embarrassing. Please feel free to continue down this path ^_^

JamesAPetts:
I approve of any work in this direction. The current console logs are relics from the rapid development of 3.0 haha. It'd be nice to turn it off entirely for production use.

PR #904, opened 2019-03-29
Replace console.log calls with debug

mikehazell:

Replaces all console logs with logging calls via [debug](https://github.com/visionmedia/debug)

The console is full of unformatted logs which are noisy and hard to filter out. #901 

A beautifully silent console unless:

1. An error condition occurs
2. The user enables log output

```js
localStorage.setItem('debug', 'cornerstoneTools:*');
``` 

* **Other information**:


src/eventDispatchers/touchEventHandlers/addNewMeasurement.js
```diff
@@ -7,13 +7,18 @@ import {
   addToolState,
   removeToolState,
 } from '../../stateManagement/toolState.js';
+import debug from 'debug';
+
+const log = debug('cornerstoneTools')
+  .extend('eventDispatchers')
+  .extend('touchEventHandlers');
```
mikehazell: The pattern I'm trying to follow with log scoping is to roughly follow the file path.
There are some cases where the scope might be shared across files within a directory.

.eslintrc.js
```diff
@@ -99,7 +99,7 @@ module.exports = {
     'no-caller': 'warn',
     'no-catch-shadow': 'warn',
     'no-confusing-arrow': 'warn',
-    'no-console': 'off',
+    'no-console': 'error',
```
dannyrb: 🙌

src/thirdParty/registerItem.js
```diff
@@ -15,19 +20,22 @@ import { lib } from '../lib.js';
...
+  }
 
-    if (overwrite) {
-      console.warn(`Overwriting ${type}/${name}`);
-    } else {
-      return;
-    }
+  if (alreadyRegistered) {
+    log('Overwriting %s', itemKey);
   }
```
mikehazell: As I'm going through files, if I find a pattern that feels odd to me I'm doing minor refactors. This is my OCD coming out. If you'd rather I didn't do this please let me know. No hard feelings 😉

mikehazell:
@dannyrb @JamesAPetts  - Checking this in early to see if you can see which direction I'm going  and give you an opportunity to steer if you want to.

This is a big library so I won't get to every file but I am trying to add consistent logging when I see a pattern.

.eslintrc.js
mikehazell: Haha, just makes them easy to find 😁

src/thirdParty/registerItem.js
JamesAPetts: It's all good, I wrote the whole plugin system in about 2 days during the 3.0 hype, any refactoring is very welcome!

src/thirdParty/registerSome.js
```diff
@@ -13,9 +13,7 @@ import register from './register.js';
  * @returns {void}
  */
 export default function(items, overwrite = false) {
-  for (let i = 0; i < items.length; i++) {
-    const { type, name, item } = items[i];
-
+  items.forEach(({ type, name, item }) => {
```
JamesAPetts: This is some pretty creative use of destructuring assignment, I like it.

JamesAPetts:
I love this, very happy with the direction this is going!

package.json
```diff
@@ -75,7 +75,8 @@
   },
   "dependencies": {
     "@babel/runtime": "7.1.2",
-    "cornerstone-math": "^0.1.7"
+    "cornerstone-math": "^0.1.7",
+    "debug": "^4.1.1"
```
dannyrb: Can you lock both of these to a specific version? (kill the `^`)

src/eventDispatchers/mouseEventHandlers/addNewMeasurement.js
```diff
@@ -1,9 +1,15 @@
 import external from '../../externalModules.js';
 import { addToolState } from '../../stateManagement/toolState.js';
 import { moveHandle, moveNewHandle } from '../../manipulators/index.js';
+import debug from 'debug';
+
+const log = debug('cornerstoneTools')
```
dannyrb: I'm not super familiar with this library, but it looks like we're create a debugger in multiple locations. Should we be creating a single `logger` instance in the store, and then extending it, as appropriate, in each module?

src/thirdParty/register.js
```diff
@@ -48,20 +52,22 @@ export default function(type, name, item, overwrite = false) {
...
   if (!type) {
-    console.warn('The type must be given in order to register.');
+    log('The type must be given in order to register.');
 
     return false;
   }
 
   if (!name) {
-    console.warn(`The ${type} must have a name in order to register.`);
+    log(`The %s must have a name in order to register.`, type);
```
dannyrb: Is there a way to indicate that this is an error/warning state?

dannyrb:
I like it (:
My questions so far are:

- How would we enable/disable logging entirely?
    - `init` setting?
    - WebPack to remove all `functions` when building for prod?
        - There is a plugin that does this for `console.log()`
- If we wanted to replace logging with a different library/implementation, but keep this same interface, how hard would it be? Do we need to implement something `adapter pattern`-y
- Is there a way for us to enable/disable logging for portions of the library? Or show different "log levels"?
    - `warnings`, `errors`, `debug/info`, `verbose`, `silly`

package.json
mikehazell: Done

src/eventDispatchers/mouseEventHandlers/addNewMeasurement.js
mikehazell: I've done something like this now. I'm not worried about creating too many instances of the logger though. That's not likely to cause performance issues unless someone starts doing that inside a loop or event handler.

src/thirdParty/register.js
mikehazell: Yep, done. I think there might be a conversation to be had / documentation to be written around how to use logging and when log, warn or error should be used.

mikehazell:
> * How would we enable/disable logging entirely?
>   
>   * `init` setting?
>   * WebPack to remove all `functions` when building for prod?
>     * There is a plugin that does this for `console.log()`

Logging is disabled by default. The exceptions to this are: 
- `logger.error` will always print to console
- `logger.warn` will print to console in dev but be turned into a regular `logger.log` in prod.

> * If we wanted to replace logging with a different library/implementation, but keep this same interface, how hard would it be? Do we need to implement something `adapter pattern`-y

I have updated the implementation to enable this.

> * Is there a way for us to enable/disable logging for portions of the library? Or show different "log levels"?
>   
>   * `warnings`, `errors`, `debug/info`, `verbose`, `silly`

For "portions of the library" yes, for levels no. I think really this comes down to contributors having some restraint around what gets logged. We can of course add other log levels later if someone feels strongly about it.

`cornerstoneTools` now exports `enableLogger` and `disableLogger` methods.

`enableLogger` takes a scoping string with which you can specify scopes to be included / excluded.
Multiple scopes are comma separated. Excluding a scope is done by prefixing with `-`. You can use the `*` match against many scopes.

Some examples:

```js
// Enable logging
csTools.enableLogger();
// defaults to "cornerstoneTools:*" which will print all logs from the cornerstoneTools package

// only show logs from setToolMode.js
csTools.enableLogger("cornerstoneTools:store:setToolMode");

// exclude the eventDispatchers
csTools.enableLogger("cornerstoneTools:*,-cornerstoneTools:eventDispatchers:*");

// Disable logging
const prevSettings = csTools.disableLogger();
// This returns a string of the previous settings in case
// you wanted to toggle logging off and back on later.
csTools.enableLogging(prevSettings);
```

Under the hood, `debug` stores the config in `localStorage` so settings persist across a refresh.

JamesAPetts:
Just had a look through, 3 green lights from me, this is really nice @mikehazell !

mikehazell:
@dannyrb - I think I'm happy with this. Was there anything else you were after from me?

dannyrb:
I love everything about this PR. Excellent work, @mikehazell! 🎉

PR #913, opened 2019-04-04
adds MEASUREMENT_COMPLETED event

mikehazell:

**FEATURE** ✨

> This PR should replicate the behaviour added to v2.4 in PR #845 

Introduce a new measurement event "cornerstonemeasurementcompleted" that is fired for the following tools:

- [x] AngleTool.js
- [x] ArrowAnnotateTool.js
- [x] BidirectionalTool.js
- [x] CobbAngleTool.js
- [x] EllipticalRoiTool.js
- [x] FreehandMouseTool.js
- [x] LengthTool.js
- [x] ProbeTool.js
- [x] RectangleRoiTool.js
- [x] TextMarkerTool.js

No such event exists. No way to tell when a user completes drawing the measurement/annotation.

* **Other information**:


src/tools/annotation/FreehandMouseTool.js
```diff
@@ -1604,14 +1607,29 @@ export default class FreehandMouseTool extends BaseAnnotationTool {
...
+  fireModifiedEvent(element, measurementData) {
+    const eventType = EVENTS.MEASUREMENT_MODIFIED;
+    const eventData = {
+      toolName: this.name,
       element,
-      EVENTS.MEASUREMENT_MODIFIED,
-      data
-    );
+      measurementData,
+    };
```
mikehazell: NOTE: the shape of the data was different here to every other tool which fires a `MEASUREMENT_MODIFIED` event. This is potentially a breaking change 😞or maybe its a bug fix 🤔?

mikehazell:
@dannyrb, @Zaid-Safadi – This is my first functional pass. Would appreciate some feedback if you have any.

I'm not super keen on this pattern:

https://github.com/cornerstonejs/cornerstoneTools/blob/f6816a8318228c9a58f5c7aeed99740bf966f463/src/eventDispatchers/mouseEventHandlers/addNewMeasurement.js#L28-L42

This would override a `doneMovingCallback` passed in by the lib user which is not great. I don't have a good plan for where else to trigger the event though as this is more often than not, the first time that all required information is available.

Let me know your thoughts.

Also, any hints on how I might add tests to this would be appreciated.

src/tools/annotation/FreehandMouseTool.js
Zaid-Safadi: This looks the right structure to me. comparing the structure you have with v2.4 is similar. I would consider the old format a bug 
https://github.com/cornerstonejs/cornerstoneTools/blob/v2.4.x/src/imageTools/freehand.js#L1195-L1204

mikehazell:
Hey @Zaid-Safadi,

To be honest, I'm not super keen on adding another event in here.

I think the least invasive/disruptive solution is to make sure that any call to "doneMovingCallback" checks for and call the user defined callback if it exists.

Alternatively, there could be a callback "onDoneMoving" option which can be defined by a tool but NOT by an end user so a tool can set this as needed and we avoid conflicts.

mikehazell:
@dannyrb - Could you comment on the `doneMovingCallback`.

Is this meant to be a user facing api?
I'm thinking maybe not for the following reasons:

1. It is not documented

2. It's being used in a bunch of places internally to queue up the "next" interaction or clean up when an action is complete. 

3. Using it would break some tools. Eg. the AngleTool's implementation would be overridden if `this.options.doneMovingCallback` were set.
https://github.com/cornerstonejs/cornerstoneTools/blob/7c38bec28bb4bccec8edaa3ad1eb891d25fe05e4/src/tools/annotation/AngleTool.js#L292-L312

The above pattern using `Object.assign` to add `doneMovingCallback` to the tool options happens in quite a few places, all of which will break in unexpected ways if the user passes `doneMovingCallback` as an option.

@Zaid-Safadi - is the reason you had for avoiding overriding `doneMovingCallback` that it would prevent end users from using this feature or was there some other reason I'm not seeing.

Zaid-Safadi:
@mikehazell yes, this looks to be an external user callback and I don’t think we should touch it for this. Honestly, even wrapping it in our own callback that calls the user set one is not ideal and will cause bugs/unexpected behavior imho.

May you explain what is your concern of using a different callback? Any specific reason for not using the pattern I described above?

mikehazell:
> this looks to be an external user callback

I agree it looks like it but I’m not convinced that that was the intention.

Given the way it’s being used in the lib I think it might be meant to be internal API only. 

> what is your concern of using a different callback

It feels like adding a new pattern which I don’t think is justified for this use case.

What I’d like best is to have a method on the  BaseAnnotationTool which is called at the appropriate time but that would require a much larger refactor.

mikehazell:
Hey @dannyrb , given our chat recently and the generally positive feedback on #915 I'm happy with this PR as it stands now.

I'd love to document it but I can't find a good place to add this. Let me know if you have any suggestions there.

dannyrb:
image

Really like seeing your logging in action ^_^
Your update caught my eye. I'm going to pull the lever on this 👍 
Thanks again, and sorry for the delay. If you're waiting on me for anything else, tag me and I'll take a look. I'm off tomorrow, but available Tuesday.

PR #915, opened 2019-04-24
WIP - makes doneMoveCallback explicitly Tool API and not user API

mikehazell:

This change is intended to clarify the purpose of the `doneMovingCallback` and prevent end users accidentally breaking tools.

Currently, `doneMovingCallback` is an attribute of `tool.options`. It is used by tools which need to be notified that a `move` or drag event has finished, allowing the tool to update its state.

As `tool.options` can be set when calling `setToolActive`, `doneMovingCallback` can be overridden by end users. In my opinion this is unintentional API and for most tools and users, unexpected behaviour.

If a user does set `doneMovingCallback` it will break several of the multi step tools.

* **What is the new behaviour (if this is a feature change)?**

`doneMovingCallback` is moved out of the `options` object and passed to the `moveHandle` methods as an extra function parameter. This makes using this feature as a tool developer more reliable and clarifies that the callback is intended as API for tool developers.

End users can no-longer set/override this callback. 

Yes, any tools which are not built into cornerstoneTools and which rely on this callback will need to be updated.

* **Other information**:

This PR has come out of work to add the `MEASUREMENT_COMPLETE` event #913

### DIFF
--- a/src/eventDispatchers/touchEventHandlers/tap.js
+++ b/src/eventDispatchers/touchEventHandlers/tap.js
@@ -68,23 +68,14 @@ export default function(evt) {
     moveableHandle.active = true; // Why here, but not touchStart?
     external.cornerstone.updateImage(element);
 
-    const toolOptions = Object.assign(
-      {},
-      {
-        doneMovingCallback: () => {
-          deactivateAllToolInstances(toolState);
-        },
-      },
-      firstToolWithMoveableHandles.options
-    );
-
     moveHandle(
       evt.detail,
       firstToolWithMoveableHandles.name,
       toolState.data,
       moveableHandle,
-      toolOptions,
-      'touch'
+      firstToolWithMoveableHandles.options,
+      'touch',
+      () => deactivateAllToolInstances(toolState)
     );
     evt.stopImmediatePropagation();
     evt.preventDefault();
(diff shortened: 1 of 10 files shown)


dannyrb:
@mikehazell, I'm pretty happy with this :+1:
I only glanced at the changes, but they look good. I'm in agreement that this should reduce confusion or accidental issues caused by overriding the method.

JamesAPetts:
Hey, also looked through this and agree with your points. Anything that tightens up/clarifies the API is great.

Zaid-Safadi:
@dannyrb, I don't think the doneMovingCallback is used anywhere (at least I haven't seen it used) and it is introduced in v3 only. If we don't want to go through the process of a breaking change, which I would vote against, we can add this as a new callback to the internal tools and leave what we have.

Anyway, I think either way is fine as long as we have some event that tells the end user when drawing a measurement is completed which is the main goal. So we should make sure there is an event to the end user as we did in v2.x  that is called "cornerstonemeasurementcompleted"

mikehazell:
Yeah, the breaking change is annoying. Might be good to deprecate the old method on the options object but make sure it keeps working.

mikehazell:
I'll try to get #913 through first, then rebase this PR and add deprecation warnings for the old `doneMovingCallback`.

mikehazell:
I've re-implemented the `options.doneMovingCallback` calls and added deprecation warnings
[in this commit](https://github.com/cornerstonejs/cornerstoneTools/pull/915/commits/70702c30f4abc26de3fe49fa8a26a7e400c8363b)

JamesAPetts:
Its likely we'll have a 4.0 in the not too distant future, if you want to rebase this against the new `vNext` branch. More info in #953.

JamesAPetts:
@mikehazell , @dannyrb  I think with deprecation warnings this could go core in 3.X, or we could put it in [4.0](https://github.com/cornerstonejs/cornerstoneTools/projects/5) only, thoughts?

JamesAPetts:
I've rebased this on v4, I'm happy for it to go in 3.0, but perhaps its best to wait until 4.0 is out so we have to deal with less conflicts?

dannyrb:
To be released in next major version bump: https://github.com/cornerstonejs/cornerstoneTools/pull/1039#issue-305567494

JamesAPetts:
Addressed in 4.0.

PR #917, opened 2019-04-26
fix(update-statistics): Update statistics for annotation tools when handle movement is done

evren217:

* **What kind of change does this PR introduce?**
Bug fix

* **What is the current behavior?**
Annotation tool data is sometimes not up-to-date when measurement modified event is triggered.

* **Does this PR introduce a breaking change?**
No, but the developers who implemented custom annotation tools are encouraged to override `updateCachedStats` method of `BaseAnnotationTool` class for more accurate results when measurement modified event is handled.

* **Other information**:
We need to override `updateCachedStats` method for all annotation tools before merging this PR:
- [x] AngleTool
- [x] ArrowAnnotateTool
- [x] BidirectionalTool
- [x] CobbAngleTool
- [x] EllipticalRoiTool
- [x] FreehandMouseTool
- [x] LengthTool
- [x] ProbeTool
- [x] RectangleRoiTool
- [x] TextMarkerTool
- [x] CircleRoi


src/tools/base/BaseAnnotationTool.js
```diff
@@ -5,6 +5,9 @@ import {
   moveHandleNearImagePoint,
   moveAnnotation,
 } from './../../util/findAndMoveHelpers.js';
+import { getLogger } from '../../util/logger';
+
+const logger = getLogger('addTool');
```
dannyrb: https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/util/logger.js#L29-L39

I'm pretty new to our logging updates, but we may want a scope that's more specific to the module we're in. Maybe `baseAnnotationTool` or something equally as namespace-y?

src/tools/annotation/LengthTool.js
```diff
@@ -207,6 +199,8 @@ export default class LengthTool extends BaseAnnotationTool {
         // So that it sits beside the length tool handle
         const xOffset = 10;
 
+        this.updateStatistics(evt, data);
```
dannyrb: If we're going to continue calling this in the render loop, should we have a dirty "invalidated" flag we can use to prevent unnecessary recalcs?

src/util/getActiveTool.js
```diff
@@ -0,0 +1,28 @@
+import { getters, state } from '../store';
+import getActiveToolsForElement from '../store/getActiveToolsForElement';
+import filterToolsUseableWithMultiPartTools from '../store/filterToolsUsableWithMultiPartTools';
+
+// Todo: We could simplify this if we only allow one active
+// Tool per mouse button mask?
+export default function getActiveTool(eventData) {
```
dannyrb: 1. Can we update the params here to be a bit more explicit instead of taking the entire event?
    - IE. `element` and `buttons`
2. Can we rename this to something more explicit? `getActiveMouseTool` or we can keep `getActiveTool`, but provide overloads where one of the params is `InteractionType` (mouse or touch)

evren217:
@dannyrb Thanks for your feedback. Please review again.

dannyrb:
This looks good to me.
The demo seems to play nice, and if you've verified the timing resolves your issue, then I think this is a solid value add :+1:

In a perfect world, I would love to see some tests added for the new functionality (especially anything on `BaseTool`).

dannyrb:
@evren217, I'm going to remove myself as a reviewer. Re-add me when/if you want me to take another peak

PR #925, opened 2019-05-07
fix(BaseBrushTool.js): prevents a crash in passiveCallback

### LINKED ISSUE #924: brush tool's passiveCallback throws when changing image if globalToolSyncEnabled is true
### Prerequisites

* [x] Are you running the latest version?
* [x] Are you reporting to the correct repository?
* [x] Did you perform a cursory search?

### Description

Crashing Bug: using the Brush tool with `globalToolSyncEnabled` can cause the application to crash when a new element is enabled.

### Steps to Reproduce

1. initialize csTools with `{ globalToolSyncEnabled: true }`
2. enable an element and load an image
3. set the brush tool mode to "active"
4. set the brush tool mode to "passive"
5. enable a second element.

**Expected behavior:** [What you expected to happen]

The 2nd element should be enabled

**Actual behavior:** [What actually happened]

The application crashes with the following error.

```
Uncaught Error: updateImage: image has not been loaded yet
    at Module.<anonymous> (cornerstone.js:7603)
    at BrushTool.passiveCallback (cornerstoneTools.js:24482)
    at setToolModeForElement (cornerstoneTools.js:11614)
    at cornerstoneTools.js:11694
    at Array.forEach (<anonymous>)
    at _resolveInputConflicts (cornerstoneTools.js:11684)
    at setToolActiveForElement (cornerstoneTools.js:11430)
    at cornerstoneTools.js:10756
    at Array.forEach (<anonymous>)
    at _repeatGlobalToolHistory (cornerstoneTools.js:10753)
```

mikehazell:
prevents a crash when _repeatGlobalToolHistory replays setting tool mode passive for brush tools

fix #924

Bug fix

Crashing Bug: using the Brush tool with `globalToolSyncEnabled` can cause the application to crash when a new element is enabled.

- `_repeatGlobalToolHistory` replays all tool changes on a new element before the image is loaded
- BaseBrushTool calls `cornerstone.updateImage` in it's `passiveCallback`
- `cornerstone.updateImage` will crash if the image is not loaded

Wrapped the call to `updateImage` in a `try ... catch`.

This should be seen as a temporary fix. A more permanent fix could be to keep track of the current tool state instead of recording and replaying changes. Alternatively, there appears to be no reason why `cornerstone.updateImage` need to crash when the image is not loaded so changing this behaviour could be more effective.

No

### DIFF
--- a/src/tools/base/BaseBrushTool.js
+++ b/src/tools/base/BaseBrushTool.js
@@ -141,7 +141,13 @@ class BaseBrushTool extends BaseTool {
    */
   // eslint-disable-next-line no-unused-vars
   passiveCallback(evt) {
-    external.cornerstone.updateImage(this.element);
+    try {
+      external.cornerstone.updateImage(this.element);
+    } catch (error) {
+      // It is possible that the image has not been loaded
+      // when this is called.
+      return;
+    }
   }
 
   /**


src/tools/base/BaseBrushTool.js
```diff
@@ -141,7 +141,13 @@ class BaseBrushTool extends BaseTool {
    */
   // eslint-disable-next-line no-unused-vars
   passiveCallback(evt) {
-    external.cornerstone.updateImage(this.element);
+    try {
+      external.cornerstone.updateImage(this.element);
```
dannyrb: I really wish `updateImage` and similar methods were safe to call with elements that don't have a loaded image or are not called with an enabled element.

dannyrb:
LGTM 👍

src/tools/base/BaseBrushTool.js
mikehazell: Totally agree.

Particularly `updateImage`. It doesn't access the image itself, it just calls `drawImage` which also doesn't touch the image but sets `enabledElement.needsRedraw`.

`needsRedraw` is then checked in the run-loop which also checks to make sure the image is loaded.

https://github.com/cornerstonejs/cornerstone/blob/f3b4accef3a700b0719e8373c08c414a1448b563/src/enable.js#L97-L99

There's nothing here that would break if `updateImage` was called when no image is loaded.

PR #934, opened 2019-05-14
fix(MagnifyTool.js): too many call _drawZoomedElement leads to memory …

xingbofeng:
…leak because of cornerstone.enable, and fix disable callback of this tool

bug fix

1. too many call _drawZoomedElement leads to memory, fix it by add condition.
2. fix disableCallback, because params is element, but not event.

* **Other information**:


xingbofeng:
and there is a problem that is magnify don't draw any annotation, i think it draw annotations is necessary. if should, i'd like to create the second pull request.

dannyrb:
> and there is a problem that is magnify don't draw any annotation, i think it draw annotations is necessary. if should, i'd like to create the second pull request.

If you're using the `globalToolSyncEnabled` flag, this actually already works. The trick is adding/enabling tools that have annotation data for zoom canvas.

dannyrb:
@xingbofeng, it looks like a `prettier` error is preventing `netlify` and the `build` from completing successfully. If you can resolve, I would be more than happy to merge:

image

xingbofeng:
@dannyrb Thanks, and there is an other problem that my annotation is not added to the magnifyElement,I want to show the annotation, for Example(this is the code I have changed, in my custom cornerstoneTools forked from official):

```javascript
// be called after _drawZoomedElement
_addAllToolsForElement(element) {
    const { tools } = cornerstoneToolsStore.state;
    tools.forEach(tool => {
        if (tool.element !== element) {
            return;
        }
        if (tool instanceof BaseAnnotationTool) {
            this._addNewToolForZoomedElement(tool);
        }
    });
}

_addNewToolForZoomedElement(tool) {
    let { constructor } = tool;
    const { initialConfiguration, name, mode } = tool;
    if (!constructor) {
        const proto = Object.getPrototypeOf(tool);
        constructor = proto.constructor; // eslint-disable-line
    }
    if (!getToolForElement(this.zoomElement, name)) {
        // add the tool to zoomElement
        addToolForElement(this.zoomElement, constructor, initialConfiguration);
        // set the mode
        cornerstoneTools[`setTool${this._upperCaseFirstLetter(mode)}ForElement`](this.zoomElement, name, {});
    }
}

_upperCaseFirstLetter(string) {
    return `${string.charAt(0).toUpperCase()}${string.slice(1)}`;
}
```

If you think it's necessary to add this feature to it, I will change the code, and PR.

xingbofeng:
Oh, thanks, i will try to use `globalToolSyncEnabled ` but not my custom code

dannyrb:
@xingbofeng, here is a quick GIF of a viewer with that flag set:

jOEGyqCqfy

```
cornerstoneTools.init({
  globalToolSyncEnabled: true
})
```

It's a less utilized flag, and it changes the default behavior of `addTool`, `setToolActive` and other methods that apply to all elements. The fact that it does this magic for magnify is entirely by accident.

dannyrb:
@all-contributors, please add @xingbofeng for this bug fix and code

xingbofeng:
Thanks for your help @dannyrb

PR #936, opened 2019-05-21
Add 10px x-offset to ROI textboxes

kevinleedrum:
The ROI textboxes are dangerously close to the handles if you draw a shape that is shorter in height than the textbox.  This is very evident if the textboxes have a background color.

image

This change adds a 10px x-offset to those textboxes, which matches the x-offset already used by the Length tool (probably for the same reason?).


dannyrb:
I think 10px out is a much better default placement. Thanks for taking the time to PR the change, @kevinleedrum! 

image

dannyrb:
@allcontributors please add @kevinleedrum for some stellar code

PR #921, opened 2019-04-30
[minor] fixed the typo in the lib.js convertToVectro3 should be conve…

mbauskar:
* **Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
   Bug Fix

* **What is the current behavior?** (You can also link to an open issue here)

    resolves https://github.com/cornerstonejs/cornerstoneTools/issues/862
    Currently, the `convertToVector3` is mapped to `convertToVectro3 `

* **What is the new behavior (if this is a feature change)?**

    fixed the typo in the lib.js, 
    https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/lib.js#L137

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
    Yes, Imports for the external packages or plugins will be affected by this changes

* **Other information**:

PR #953, opened 2019-06-05
vNext - Segmentation overhaul

JamesAPetts:
# Segmentation Overhaul - This is a breaking change!

Hey all, this is a breaking change. I've forked the current master branch to a `vNext` that we can PR against for anything which is a breaking change, namely these PRs currently: #915, #921.

In this PR I rebuild the segmentation mask system from the ground up to solve two main issues:

- Interoperability with other libraries (e.g. [`vtkjs`](https://github.com/Kitware/vtk-js))

- Scalability for large segment counts.

## Interoperability with other libraries
During the [NAMIC Project Week 30](https://projectweek.na-mic.org/PW30_2019_GranCanaria/), a bunch of us [cornerstone devs](https://projectweek.na-mic.org/PW30_2019_GranCanaria/Projects/DICOMSEG-Cornerstone-VTKJS/) attempted to link vtkjs and cornerstone. We made some good headway with a [simple POC](https://deploy-preview-5--react-vtkjs-viewport.netlify.com/cornerstone-sync-painting) where we allocated an `ArrayBuffer` globally for the segmentation, and (albeit rather hackily) made both `cornerstoneTools` and `vtkjs` read and write to it.

We only did this for one segment in our POC. `vtkjs` renders a labelmap, and `cornerstoneTools` had binary maps, so we were a bit limited.

With this PR `cornerstoneTools` now uses labelmaps, whilst supporting overlapping segment functionality. All memory management is done by `cornerstoneTools`, and vtkjs and other libraries that want to operate on the labelmap data will be able to through a simple yet powerful api.

## Scalability
Previously each segment was stored as a binary label map which was stored per segment. This wasn't scalable in the browser due to memory constraints if you had multiple segments on every slice. We even had a hard-coded limit of 19 segments in `cornerstoneTools` 3.0.

### Non-overlapping labelmaps

With this PR a 3D labelmap can be assigned to a scan, which can have up to 65535 _non-overlapping_ segments. This means we will be able to support things such as [freesurfer](https://surfer.nmr.mgh.harvard.edu/) segmentations, which often have 200-400 labels per map. And detailed fullbody segmentations which may come out of future AI pipelines, etc.

Previously I used to only allocate per-slice binary maps for each segment when it was drawn on. This is nice whilst living in 2D cornerstone land, but its just way to slow to dynamically interoperate with vtkjs in such a fashion. Internal discussions made it quite clear that allocating enough space for the whole labelmap is worth it for cleaner and faster interoperability as we make more advanced PWAs.

### Dealing with overlapping segments

For the _vast majority_ of use cases, segmentation files (DICOMSEG, NIFTI, etc) contain non-overlapping segments. Overlapping segments do exist however, and they are still possible in this library. To have multiple segments overlap, we simply add _multiple labelmaps_. These can then be encoded as a single overlapping DICOMSEG provided the correct bridges are built in [`dcmjs`](https://github.com/dcmjs-org/dcmjs).

The only segments I've previously seen that are overlapping are radiotherapy targets, and even then, these are much much more likely to be encoded as RTSTRUCTs, which can be easily visualized by the `FreehandMouseTool`.

As we are now using 2 Bytes/voxel for a labelmap rather than 1 byte/voxel on the binary map. In the _worst case_ scenario where all the segments overlap, you will use twice the memory allocation. In my opinion this is a worthy trade-off to have the more commonly seen segmentations packed up to 3,276,750% more densely than previously. **If you are currently using overlapping segments for a more difficult use, please comment and we can discuss solutions.**

## Extensibility
New labelmap types which might be predicted as output of AI/ML pipelines, such as probability maps or occupation maps should be easy to add as a new segment type, without causing another breaking change. One such fractional segment can be stored in the same amount of memory.

### "Standard Usage"

For general `cornerstoneTools` usage, nothing much has changed. If you start painting on a scan, all will work as expected. As before the `BaseBrushTool` has methods to change:
- The actively drawn segment color:
```js
instanceOfBaseBrushTool.nextSegment();
instanceOfBaseBrushTool.previousSegment();
```
- The brush size:
```js
instanceOfBaseBrushTool.increaseBrushSize();
instanceOfBaseBrushTool.decreaseBrushSize();
```

Assume this from here to the end of this section:

```js
const brushModule = cornerstoneTools.store.modules.brush;
```

Labelmap `0` will be drawn on by default, this can be changed in the brush store currently by:
```js
brushModule.setters.activeLabelmap(element, labelmapIndex);
```

The alpha rendering settings of active labelmaps is configured by `brushModule.state.alpha`, and the alpha of inactive labelmaps is set by `brushModule.state.alphaOfInactiveLabelmap`.

If multiple labelmaps are present on a scan, the active one is rendered on top. If `brushModule.state.alphaOfInactiveLabelmap === 0`, inactive labelmaps aren't rendered at all, saving on rendering computation.

Cached bitmaps are used  if the labelmap doesn't change between rendering frames, and are only regenerated when the labelmap is invalidated, or a new slice is scrolled to.

### Read/Write labelmap from external lib

To get a reference to all the labelmaps on a slice:

```js
const labelmapBuffers = brushModule.getters.labelmapBuffers(element);
// Optional pass a labelmapIndex as a second parameter and you only get that buffer back.
```
Which will return an object like this:

```js
[
  {
    labelmapIndex: 0,
    bytesPerVoxel: 2,
    buffer
  },
  {
    labelmapIndex: 1,
    bytesPerVoxel: 2,
    buffer
  }
  ...
]
```

Even more laziliy, you may fetch only the active labelmap via:
```js
brushModule.getters.activeLabelmapBuffer(element);
```

You can then read from this `ArrayBuffer` to render your data using another lib, and write straight to this buffer directly to update the labelmap. When you do so make sure to call:
```js
brushModule.setters.invalidateBrushOnEnabledElement(elementOrEnabledElementUID, labelmapIndex);
```
to let cornerstoneTools know the data has been modified, so that it will regenerate cached images.

### Load a segmentation in from an external source.
A labelmap can be set from an external labelmap buffer of size rows * columns * frames by:

```js
// By providing the cornerstone enabled element or an enabled element's UUID:
brushModule.setters.labelmap3DForElement(
  elementOrEnabledElementUID,
  buffer,
  labelmapIndex,
  metadata = []
);

// Or by providing the imageId of the first image of the stack. This is useful if you want to load a
// segmentation into cornerstone when its not yet being displayed. As we don't have access to an element, the numberOfFrames must be explicitly specified.
brushModule.setters.labelmap3DByFirstImageId(
  firstImageId,
  buffer,
  labelmapIndex,
  metadata = [],
  numberOfFrames
);
```
The metadata is optional and is currently available for arbitrary data. I typically store the segment metadata info from DICOMSEG here.

Adapaters for stuff like DICOMSEG should be written in [`dcmjs`](https://github.com/dcmjs-org/dcmjs). The current `dcmjs` implementations for segmentations are out of date with regards to this PR.

### Custom colorLUTs.
Color LUTs are generated automatically by default for each subsequent labelmap, however if you wish to set your own LUT, you can by:

```js
brushModule.setters.colorLUT(labelmapIndex, colorLUT);
```

Where colorLUT is an array-of-arrays of 4 integers: `[R, G, B, A].` The zero value (i.e. no segment) does not need to be supplied and is prepended to the input.

I want to add an API endpoint to set the visibility of individual segment colors by directly manipulating the alpha, which will then be multiplied by the global alpha (similar to ITK snap if you've used it), but I haven't done this yet.

### DIFF
--- a/netlify-example/brush/index.html
+++ b/netlify-example/brush/index.html
@@ -35,7 +35,7 @@
 							<td>
 								<table>
 									<tr>
-										<td>Click and drag to use the paint with the selected segmentation.</td>
+										<td>Click and drag to paint with the selected segment.</td>
 									</tr>
 									<tr>
 										<td>Release the mouse to stop painting.</td>
@@ -306,10 +306,10 @@
 	function brushApiCall (opperation) {
 		switch (opperation) {
 			case 'next-segment':
-				brushTool.nextSegmentation();
+				brushTool.nextSegment();
 				break;
 			case 'previous-segment':
-				brushTool.previousSegmentation();
+				brushTool.previousSegment();
 				break;
       case 'switch-labelmap':
         if (activeLabelmap === 0) {
--- a/netlify-example/brush/index.html
+++ b/netlify-example/brush/index.html
@@ -81,9 +81,17 @@
 								<input type="range" id="alpha" min="0" max="255">
 							</td>
 							<td>
-								alpha - The transparency of segmentations.
+								alpha - The transparency of the active segmentation.
 							</td>
 						</tr>
+            <tr>
+              <td>
+                <input type="range" id="alpha-inactive" min="0" max="255">
+              </td>
+              <td>
+                alphaOfInactiveLabelmap - The transparency of inactive segmentations.
+              </td>
+            </tr>
 					</table>
 				</div>
 			</div>
@@ -325,13 +333,22 @@
 	// API Sliders
 	const alphaSlider = document.getElementById("alpha");
 
-	alphaSlider.defaultValue = Math.floor(0.4*255);
+  alphaSlider.defaultValue = Math.floor(0.6*255);
+  alphaSlider.addEventListener('input', event => {
+    const normalisedAlpha = event.target.value/255.0;
 
-	alphaSlider.addEventListener('input', event => {
-		const normalisedAlpha = event.target.value/255.0;
+    brushTool.alpha = normalisedAlpha;
+  });
 
-		brushTool.alpha = normalisedAlpha;
-	});
+
+  const alphaOfInactiveLabelmapSlider = document.getElementById("alpha-inactive");
+
+  alphaOfInactiveLabelmapSlider.defaultValue = Math.floor(0.0);
+  alphaOfInactiveLabelmapSlider.addEventListener('input', event => {
+    const normalisedAlpha = event.target.value/255.0;
+
+    brushTool.alphaOfInactiveLabelmap = normalisedAlpha;
+  });
 
 
 </script>
--- a/netlify-example/brush/index.html
+++ b/netlify-example/brush/index.html
@@ -47,87 +47,45 @@
 								</table>
 							</td>
 						</tr>
-						<tr>
-							<th>Keyboard Input</th>
-							<td>
-								<table>
-									<tr>
-										<td>Next Segmentation:</td>
-										<td>]</td>
-									</tr>
-									<tr>
-										<td>Previous Segmentation:</td>
-										<td>[</td>
-									</tr>
-									<tr>
-										<td>Increase Brush Size:</td>
-										<td>+</td>
-									</tr>
-									<tr>
-										<td>Decrease Brush Size :</td>
-										<td>-</td>
-									</tr>
-								</table>
-							</td>
-						</tr>
 					</table>
 					<h2>Brush Tool API</h2>
 					<table>
 						<tr>
 							<th>
-								<button type="button" api-call="hide-all">
-									Hide All Segmentations
+								<button type="button" api-call="next-segment">
+									Next Segment
 								</button>
 							</th>
 							<td>
-									Hides all segmentations unless that segmentation is actively being drawn.
+								Switch to the next segment index.
 							</td>
 						</tr>
-						<tr>
-							<th>
-								<button type="button" api-call="show-all">
-									Show All Segmentations
-								</button>
-							</th>
-							<td>
-								Shows all segmentations on the canvas, all the time.
-							</td>
-						</tr>
-						<tr>
-							<th>
-								<button type="button" api-call="hide-third">
-									Hide 3rd Segmentation
-								</button>
-							</th>
-							<td>
-								An example of hiding a single segmentation. Hide's the 3rd (yellow) segmentation. When hidden it can only be seen whilst actively being drawn.
-							</td>
-						</tr>
-						<tr>
-							<th>
-								<button type="button" api-call="show-third">
-									Show 3rd Segmentation
-								</button>
-							</th>
-							<td><p>
-								An example of un-hiding a single segmentation. Unhide's the 3rd (yellow) segmentation.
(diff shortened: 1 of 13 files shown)


JamesAPetts:
TODO:

- [ ] Rename `instanceOfBaseBrushTool.nextSegmentation()` => `instanceOfBaseBrushTool.nextSegment()` to be semantically correct with respect to DICOM and to keep @pieper happy.
- [ ] Emit a `LABELMAP_MODIFIED` event that external libraries can pick up on to know when _they_ should re-render based on modifications made through `cornerstoneTools` usage.

JamesAPetts:
Danny looked through this on Friday and doesn't have any large concerns. I'm going to merge it to `vNext` now, but the whole `vNext` branch is a WIP, so feel free to add any further comments/criticisms/requests post-merge.

dolremi:
@JamesAPetts when will these changes merge into the master branch？ I am looking for a way to handle multiple segmentations, the only thing I can find is using next/previousSegment(), is there any other way to handle multiple segmentation, thanks.

JamesAPetts:
@dolremi This has been out for a while. You can do multiple segmentations by changing the active label map. There 100% needs some better documentation, but checkout the demo at `netlify-example/brush/index.html`, a live example is here: 
https://deploy-preview-1291--cornerstone-tools.netlify.app/netlify-example/brush/

There does appear to be no good api hooks for it, it feels a bit "manual", I'd be happy to review a PR for something like "setActiveLabelMapIndex".

dolremi:
@JamesAPetts Thx, then we need to save all the labelMap3D if we want to save the segmentation information.

dolremi:
@JamesAPetts BTW, how could I set the color of segmentation tool when setting a new  segmentindex, does "brushModule.setters.colorLUT(labelmapIndex, colorLUT);" work?

ashishbairwa:
@JamesAPetts @dannyrb @sisobus I'm trying to load a bitmap mask of dimensions 32x32, so the resulted array is like [[0,0,1... till 32],[],[],[] ... till 32] on study images of size 32x32. I tried to load that mask something like this:

```
const data = new Uint16Array(maskData);
parseSegment(data, element);

export const parseSegment = (arrayBuffer, element) => {
    // const stackToolState = getToolState(
    //     element,
    //     "stack"
    // );
    // const imageIds = stackToolState.data[0].imageIds;
    const {
        labelmapBufferArray,
        segMetadata,
        segmentsOnFrame
    } = dcmjs.adapters.Cornerstone.Segmentation.generateToolState(
        "1.3.12.2.1107.5.2.40.38828.2018013012333548393110760",
        arrayBuffer,
        cornerstone.metaData,
    );
    const { setters } = getModule(
        "segmentation"
    );
    setters.labelmap3DByFirstImageId(
        "1.3.12.2.1107.5.2.40.38828.2018013012333548393110760",
        labelmapBufferArray[0],
        0,
        segMetadata,
        1,
        segmentsOnFrame
    );
}
```
But getting invalid dicom issue. Is there any specific to load a 2d bitmap segment?

PR #960, opened 2019-06-07
vNext - spherical brush

JamesAPetts:
A spherical brush implementation for `cornerstoneTools`.

The brush paints outwards from the input slice in both directions on slices that have `imagePlane` metadata accessible from a `cornerstone` metadata provider. The brush stops looking for more slices in that direction when either no `imagePlane` metadata is found, or when a slice is out of the influence range of the brush.

### DIFF
--- a/index.html
+++ b/index.html
@@ -71,6 +71,9 @@
             <a href="#" data-tool="Brush">BrushTool</a>
           </li>
           <li>
+            <a href="#" data-tool="SphericalBrush">SphericalBrushTool</a>
+          </li>
+          <li>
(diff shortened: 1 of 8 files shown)


JamesAPetts:
I'll get the next round of changes through before I focus my attention back on this. Perhaps this will be a mini-project during NAMIC Project Week 31.

JamesAPetts:
Pulled in `vNext` changes and added some fallback behavior if there is no metadata.
Added the tool to the index demo, the brush will only paint on slices in range, this also prevents the brush being applied to localizers in the stack.the

PR #968, opened 2019-06-11
WIP vNext - Segmentation Changes II.

JamesAPetts:
# Round 2, lets go!

## Efficiency
On mouse down the labelmap is fetched and the reference to it cached within the tool. Whether the stroke should be additive or subtractive is also cached on the initial click (either via `ctrl` being pressed, or `configuration.alwaysEraseOnClick` being cached) .

On mouse up the labels present on the slice are cached in the `labelmap2D.segmentsOnSlice` object (which could probably do with a better name). This is useful, as it cheap to do for a single slice, and then when you eventually want to output this segmentation you may do so a lot quicker as you already have this info. For full body with many segmentations, `dcmjs` was taking up to 10 seconds to save a DICOMSEG, which is hardly good.

In the future it'd be good to draw to on a separate, smaller image and bake it into the larger image when needed, this might be quicker than redrawing the whole frame? We currently have issues with very large DX images.

### `getters`
- `getters.metadata(elementOrEnabledElementUID, labelmapIndex, segmentIndex)`:
  -  now defaults to the `activeLabelmapIndex` if the `labelmapIndex` argument isn't given.
- `getters.activeSegmentIndex(elementOrEnabledElementUID, labelmapIndex)`:
  -  Returns the active segment segment index  for the active labelmap for the series displayed on the element.
  -  Defaults to the `activeLabelmapIndex` if the `labelmapIndex` argument isn't given.
-  `getters.activeLabelmapIndex(elementOrEnabledElementUID)`:
    -  Returns the activeLabelmapIndex of the labelmap displayed on the element.
### `setters`
- `setters.deleteSegment( elementOrEnabledElementUID, segmentIndex, labelmapIndex)`:
    -  Deletes the segment from the labelmap and any associated metadata.
    -  Defaults to the `activeLabelmapIndex` if the `labelmapIndex` argument isn't given.


JamesAPetts:
TODO: For input we need to set up the `segmentsOnSlices` objects, this could be quite expensive for CT volumes, and should likely be done in a worker.

PR #969, opened 2019-06-11
Modify toolstate compare function for removeToolState method

saseokjun:

Bug fix

Modify toolState data compare function

Previously compare only simple object was possible, but now complex object is also possible.

Nope

* **Other information**:


JamesAPetts:
Unless I'm mistaken, this isn't needed.

Intenrally JavaScript actually has two different approaches for testing equality. Primitives like strings and numbers are compared by their value, while objects like arrays, dates, and plain objects are compared by their reference. That comparison by reference basically checks to see if the objects given refer to the same location in memory.

As `toolData.data[i]` and `data` are both objects, the equality will only be true when they are the same object in memory, which is precisely the thing you wish to delete.

Issue #935, opened 2019-05-17
Prevent WheelEvent from propagating when middle mouse button is pressed


Karl-Heinrich:
When pressing the middle mouse button to activate a tool (e.g. PanTool, LengthTool, ..) a WheelEvent might be triggered and propagated as well. 

This itself might not be problematic, but when using a stack of images and the middle mouse button is pressed to e.g. zoom, pan, measure, .. the images in the stack start scrolling.

This behaviour is unwanted and can be deactivated by manually grabbing the native event and disabling the usual scroll behaviour:

```
enabledElement.onwheel = evt => {
            if (evt.deltaY < 1 || evt.deltaY > -1) {
                evt.stopImmediatePropagation();
            }
        };
```

Recreating this is straightforward, just set the `mouseButtonMask` to 4 (middle mouse button) in any of the public examples and e.g. log something when the WheelEvent happens or take a look at [this fiddle](https://jsfiddle.net/g2jefho7/1/).

**Important**: If you have a good mouse, you might not be affected! I am using a Logitech M500 for which the bug certainly is real.

![wheelOnFire](https://user-images.githubusercontent.com/38315992/57937300-f07b1780-78c5-11e9-8dc9-2d9297180361.gif)

dannyrb:
@Karl-Heinrich, I am able to reproduce, and I agree with your assessment. Is this something you would be interested in drafting a Pull Request for?

Karl-Heinrich:
Sure! Could you point me to the correct file(s), where you think a fix to this should go?

dannyrb:
Pretty much all of the events we support tool interaction for go through our global listeners, so we can normalize them across browsers, and tack on some extra data. You can find them here:

- [src/eventListeners](https://github.com/cornerstonejs/cornerstoneTools/tree/master/src/eventListeners)

For `wheelEvent` specific logic, check out:

- [src/eventListeners/wheelEventListener.js](https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/eventListeners/wheelEventListener.js) and
- [src/eventListeners/wheelEventListener.test.js](https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/eventListeners/wheelEventListener.test.js)

It should be as simple as modifying the first few lines of the wheelEventListener's handler function.

PR #979, opened 2019-06-18
Pr/wheelevent prevention

Karl-Heinrich:

Bug fix

Current behavior is documented in Issue #935

No

* **Other information**:
It think we can safely assume that if a user tries to scroll less than one pixel, he is not trying to scroll. The event triggered by the users mouse will emit a vertical scrolling value (deltaY) of either 0 or -0, if he uses his middle mouse button without scrolling. This happens e.g. when pressing down.

However, a few questions remain:
* Where do you document bug fixes?
I've added comments in the code so other developers stumbling upon this will understand what the code does. You might not want to have this commented in the code so if I should remove it, just tell me.
* The test for this is a little bit of a hack - is it ok though?
If the event does not get emitted because we return early, the `assert` part of the test implementation never gets reached and therefore the test will always succeed. If we dont return early, the assert gets reached and it always fails.


JamesAPetts:
This makes sense and LGTM 👍 .

dannyrb:
> Where do you document bug fixes?

In a perfect world, two places:

- The changelog, updated by using the correct commit syntax
- In a unit test; codifying the fix as a business requirement

The inline comment is fine as well. I'm all for semantic code, where possible, to reduce comments though. Code gets updated, comments don't. For example:

```
  const isMouseWheelEventWithNoScroll = evt.deltaY > -1 && evt.deltaY < 1
  if (isMouseWheelEventThatDoesNotScroll ) {
    return;
  }
```

> The test for this is a little bit of a hack - is it ok though?

I generally prefer that an `expect` validates the behavior, but I'm not sure what the best practice is. If I'd written the test, I probably would have aimed for something like:

```
const cornerstoneMouseWheelEventHandler = jest.fn();

localElement.addEventListener('cornerstonetoolsmousewheel', cornerstoneMouseWheelEventHandler);

// SUT
wheelEventListener.enable(localElement);
localElement.dispatchEvent(middleMouseButtonClickedEvent);

expect(cornerstoneMouseWheelEventHandler.mock.calls.length).toBe(0);
```

Actually, that's still not a unit test, is it? The only thing we _need_ to test here is `wheelEventHandler(evt)`, which you can call directly, and then use a mocked `triggerEvent()` to verify that it was called 0 times.

dannyrb:
@Karl-Heinrich, feel free to make the above changes. As things stand, I'll likely merge this as-is in the next day or two even if you don't. It's a solid improvement! Thanks for taking the time to PR it and add tests ^_^

dannyrb:
@all-contributors please add @Karl-Heinrich for these unit tests and code fixes

PR #1004, opened 2019-07-12
adding the possibility to handle multiples start/end/error loading handlers

biharck:

It adds the possibility to handle multiple Start/Stop/Error Load Handlers.

There is no way to set more one handler for each viewport.

The possibility to add more than one handler function

No


src/util/scrollToIndex.js
```diff
@@ -44,9 +44,19 @@ export default function(element, newImageIdIndex) {
     newImageIdIndex += stackData.imageIds.length;
   }
 
-  const startLoadingHandler = loadHandlerManager.getStartLoadHandler();
-  const endLoadingHandler = loadHandlerManager.getEndLoadHandler();
-  const errorLoadingHandler = loadHandlerManager.getErrorLoadingHandler();
+  let viewportIndex = {};
```
swederik: Viewport index is only in OHIF. Can you use setStartLoadHandler(handler, element) where element is the actual DOM element? Not sure that works as a key, but I think it does.

If not we can use the enabled element's UUID: https://github.com/cornerstonejs/cornerstone/blob/master/src/enable.js#L69

swederik:
You also need to add the removal of the handlers when the element is disabled.

src/util/scrollToIndex.js
biharck: I tried with the DOM element but it didn't work. I'll try with UUID.

src/stateManagement/loadHandlerManager.js
```diff
@@ -33,6 +78,7 @@ const loadHandlerManager = {
   getEndLoadHandler,
   setErrorLoadingHandler,
   getErrorLoadingHandler,
+  removeHandler,
```
swederik: Could you make this `removeHandlers` (plural)?

src/store/internals/removeEnabledElement.js
```diff
@@ -98,6 +100,17 @@ const _removeEnabledElement = function(enabledElement) {
   }
 };
 
+/**
+ * Remove load handler for the enabled element
+ * @private
+ * @method
+ * @param {HTMLElement} enabledElement
```
swederik: JSDoc is wrong, this should be `element`

src/stateManagement/loadHandlerManager.js
biharck: done

swederik: It's still exporting the name as `removeHandler`....

src/store/internals/removeEnabledElement.js
```diff
@@ -98,6 +100,17 @@ const _removeEnabledElement = function(enabledElement) {
...
+ * @returns {void}
+ */
+const _removeLoadHandlers = function(element) {
+  loadHandlerManager.removeHandler(element);
```
swederik: This needs to be updated to `removeHandlers` as well.

src/store/internals/removeEnabledElement.js
biharck: done

src/store/internals/removeEnabledElement.js
biharck: done

PR #1024, opened 2019-07-23
Adding Husky and Prettier/lint

biharck:

Adding Husky and Prettier/lint so that it can prevent CI failing due to not aligned pushed files

Currently, developers are able to push changes where files are not allowed with the Linter

Husk is going to run prettier and Lint tasks before commit.

No


biharck:
Before you merge, I'll have to fix all issues found when running lint

biharck:
> Before you merge, I'll have to fix all issues found when running lint

Done @dannyrb

biharck:
@dannyrb I am not sure whether we should count the number of lines for test files: https://codeclimate.com/github/cornerstonejs/cornerstoneTools/pull/1024

.eslintrc.js
```diff
@@ -99,7 +99,7 @@ module.exports = {
     'no-caller': 'warn',
     'no-catch-shadow': 'warn',
     'no-confusing-arrow': 'warn',
-    'no-console': 'error',
+    'no-console': 'off',
```
dannyrb: Good call. We use `console.warn` heavily, but should probably either provide guidance on how to remove w/ a bundler, or use `debug` and turn off when built for production.

src/store/modules/brushModule.js
```diff
@@ -136,31 +136,6 @@ function enabledElementCallback(enabledElement) {
...
-  );
-
-  const enabledElementUID = cornerstoneEnabledElement.uuid;
-  const colormap = external.cornerstone.colors.getColormap(state.colorMapId);
-  const numberOfColors = colormap.getNumberOfColors();
-
-  // Remove enabledElement specific data.
-  delete state.visibleSegmentations[enabledElementUID];
-  delete state.imageBitmapCache[enabledElementUID];
-}
```
dannyrb: Was this removed because there are no references/calls to it, and it's not exported?

src/store/setToolMode.js
```diff
@@ -404,10 +404,9 @@ function _resolveMouseInputConflicts(tool, element, options) {
  *
  * @param {Object} tool
  * @param {HTMLElement} element
- * @param {Object} options
  * @returns {undefined}
  */
-function _resolveTouchInputConflicts(tool, element, options) {
```
dannyrb: I guess these were internal and unused as well. Hmm

src/tools/annotation/ArrowAnnotateTool.js
```diff
@@ -91,10 +91,6 @@ export default class ArrowAnnotateTool extends BaseAnnotationTool {
     );
   }
 
-  updateCachedStats(image, element, data) {
```
dannyrb: I think this is fine, but need to verify it's not called generically from somewhere

src/tools/annotation/CircleRoiTool.test.js
```diff
@@ -246,21 +246,44 @@ describe('CircleRoiTool.js', () => {
...
-        100, 3, 6,
-        100, 100, 100,
-        100, 4, 5,
-        100, 3, 6,
-        100, 100, 100,
-        100, 4, 5,
-        100, 3, 6
-      ];
-    };
+    external.cornerstone.getPixels = () => [
```
dannyrb: May revert and add an eslint ignore for just this line to keep formatting

src/tools/annotation/EllipticalRoiTool.test.js
```diff
@@ -16,11 +16,7 @@ jest.mock('./../../externalModules.js', () => ({
     metaData: {
       get: jest.fn(),
     },
-    getPixels: () => {
-      return [100, 100, 100,
-        100, 4, 5,
-        100, 3, 6];
-    }
+    getPixels: () => [100, 100, 100, 100, 4, 5, 100, 3, 6],
```
dannyrb: Same here

src/tools/annotation/ProbeTool.test.js
```diff
@@ -17,11 +17,10 @@ jest.mock('./../../externalModules.js', () => ({
       get: jest.fn(),
     },
     getStoredPixels: (element, x, y) => {
-      const storedPixels = [10, 20, 30,
-        40, 50, 60,
-        70, 80, 90];
+      const storedPixels = [10, 20, 30, 40, 50, 60, 70, 80, 90];
```
dannyrb: Same here

src/tools/annotation/RectangleRoiTool.test.js
```diff
@@ -16,11 +16,7 @@ jest.mock('./../../externalModules.js', () => ({
     metaData: {
       get: jest.fn(),
     },
-    getPixels: () => {
-      return [100, 100, 100,
-        100, 4, 5,
-        100, 3, 6];
-    }
+    getPixels: () => [100, 100, 100, 100, 4, 5, 100, 3, 6],
```
dannyrb: And here

src/tools/annotation/TextMarkerTool.js
```diff
@@ -117,10 +117,6 @@ export default class TextMarkerTool extends BaseAnnotationTool {
     return distanceToPoint < 10 || insideBoundingBox;
   }
 
-  updateCachedStats(image, element, data) {
```
dannyrb: Again, need to double check we're not calling this generically; but I think we should be safe

src/tools/base/BaseAnnotationTool.js
```diff
@@ -16,10 +13,6 @@ const logger = getLogger('baseAnnotationTool');
  * @extends Tools.Base.BaseTool
  */
 class BaseAnnotationTool extends BaseTool {
-  constructor(...args) {
-    super(...args);
```
dannyrb: I guess if removing this caused issues, we'd notice pretty quickly.

src/tools/base/BaseAnnotationTool.js
```diff
@@ -162,18 +155,6 @@ class BaseAnnotationTool extends BaseTool {
...
-   * Updates cached statistics for the tool's annotation data on the element
-   *
-   * @param {*} image
-   * @param {*} element
-   * @param {*} data
-   * @returns {void}
-   */
-  updateCachedStats(image, element, data) {
-    logger.warn(`updateCachedStats not implemented for ${this.name}.`);
-  }
```
dannyrb: Hmm... It says that it's abstract and must be implemented. Why was this nuked? At the very least, it would be virtual to indicate usage/pattern?

src/tools/cursors/MouseCursor.js
```diff
@@ -23,7 +23,7 @@ export default class MouseCursor {
   }
 
   /**
-   * getIconSVG - Returns an SVG of the icon only.
+   * GetIconSVG - Returns an SVG of the icon only.
```
dannyrb: Need to touch and remove the name prefix from each of these

src/store/modules/brushModule.js
dannyrb: CC: @JamesAPetts, just a heads up that this was removed. It looks like we never exported it as part of the module.

src/tools/base/BaseAnnotationTool.js
dannyrb: So it looks like we added this to force annotation tools to consistently update their data in the same way. `moveHandles` verifies tool is an instance of `BaseAnnotationTool`, and it calls its underlying `updateCachedStats` method. I don't love the implementation, but removing this means we don't surface a helpful warning to tool authors.

PR #1032, opened 2019-07-26
refactor(FreehandMouseTool): Consistently access event `element`

dlwire:
Committing this in smaller parts just to make it easier to follow the refactor as it happens. This
part just makes sure we are accessing the element and using the element in a consistent manner
throughout. Also made some small changes to other attributes that are accessed off the event data.

This is a refactor of the FreehandMouseTool to make sure we're consistently accessing the `element` associated with the event data.

No change.

No change.

No.

* **Other information**:

### DIFF
--- a/src/tools/annotation/FreehandMouseTool.js
+++ b/src/tools/annotation/FreehandMouseTool.js
@@ -394,13 +394,9 @@ export default class FreehandMouseTool extends BaseAnnotationTool {
               // The mouse location
               lines.push(config.mouseLocation.handles.start);
             }
-            drawJoinedLines(
-              context,
-              eventData.element,
-              data.handles.points[j],
-              lines,
-              { color }
-            );
+            drawJoinedLines(context, element, data.handles.points[j], lines, {
+              color,
+            });
           }
         }
 
@@ -573,9 +569,8 @@ export default class FreehandMouseTool extends BaseAnnotationTool {
   }
 
   handleSelectedCallback(evt, toolData, handle, interactionType = 'mouse') {
-    const eventData = evt.detail;
-    const element = eventData.element;
-    const toolState = getToolState(eventData.element, this.name);
+    const { element } = evt.detail;
+    const toolState = getToolState(element, this.name);
 
     if (handle.hasBoundingBox) {
       // Use default move handler.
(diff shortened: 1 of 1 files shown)


src/tools/annotation/FreehandMouseTool.js
```diff
@@ -1032,11 +1029,8 @@ export default class FreehandMouseTool extends BaseAnnotationTool {
   _startDrawing(evt) {
     const eventData = evt.detail;
     const measurementData = this.createNewMeasurement(eventData);
-    const element = eventData.element;
+    const { element } = eventData;
     const config = this.configuration;
-
-    this._referencedElement = element;
```
dlwire: This wasn't used anywhere. It seems like it would be expedite grabbing the element throughout but I'm not sure we want to save this type of information from event to event.

There's also this._activeDrawingToolReference, which caches a handle for the data, which looks like it was intended to do the same thing but isn't used consistently. Probably best to replace with a helper function that takes the element or event as an argument.

src/tools/annotation/FreehandMouseTool.js
```diff
@@ -1265,7 +1263,7 @@ export default class FreehandMouseTool extends BaseAnnotationTool {
     if (this._dragging) {
       invalidHandlePlacement = this._checkHandlesPencilMode(data, eventData);
     } else {
-      invalidHandlePlacement = this._checkHandlesPolygonMode(data, eventData);
+      invalidHandlePlacement = this._checkHandlesPolygonMode(data, element);
```
dlwire: Fixing this.

src/tools/annotation/FreehandMouseTool.js
JamesAPetts: Huh, interesting, `newImageCallback` uses the `_activeDrawingToolReference`, `_referencedElement` can indeed be nuked.

> Probably best to replace with a helper function that takes the element or event as an argument.

The difficulty here is that the event will contain the new imageId, etc. What's needed is a reference to the actively drawn data, so that it can be closed cleanly if you are part way through the drawing loop and decide to change the image, either by scrolling or dragging a new scan to the canvas.

dlwire: Sounds like I'll have to investigate more with `this._activedrawingToolReference` if we're going to do anything with that (it would definitely be a separate PR). For this, I was trying to keep it pretty small just to keep continuity and make the review pretty straightforward.

PR #1033, opened 2019-07-28
feature(brush): add configuration option to skip paint if the segmentation is not visible (skips event calls)

dannyrb:
```js
cornerstoneTools.addTool(cornerstoneTools.BrushTool, {
	name: 'Brush',
  configuration: {
	  skipPaintForInvisibleSegmentations: true
  }
})
```

Tests:
- `cornerstoneTools.store.modules.brush.state.alpha = 0`
- `cornerstoneTools.getToolState(element, 'brush').data[segIndex].alpha = 0`
- `cornerstoneTools.store.modules.brush.setters.brushVisibilityForElement(enabledElementUID, segIndex, visible)`

CC: @JamesAPetts 

Other stuff:

- Shifts duplicate `get/set` toolState check out of strategies and into `BrushTool`'s `_paint` method.
- Provides a way to prevent UX confusion of adding/removing from a segmentation while it's not visible
- Provides a way to prevent modified/completed events if we're not actually modifying brush data

### DIFF
--- a/src/tools/base/BaseBrushTool.js
+++ b/src/tools/base/BaseBrushTool.js
@@ -7,7 +7,7 @@ import isToolActive from './../../store/isToolActive.js';
 import store from './../../store/index.js';
 import triggerEvent from './../../util/triggerEvent.js';
 
-const { state, setters } = store.modules.brush;
+const { state, getters, setters } = store.modules.brush;
 
 /**
  * @abstract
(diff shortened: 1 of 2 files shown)


dannyrb:
I'm going to merge this for now; but please feel free to call out issues/complaints. Ideally, this should not impact consumers by default, as the functionality is hidden by a config flag.

PR #1034, opened 2019-07-30
 fix(Measurement) Measurement Table uses wrong units

jfmedeiros1820:

Bug fix https://github.com/OHIF/Viewers/issues/702

Measurement unit suffix used was not returned in the result object

The measurement result object now returns the unit suffix to be used in other places

No

* **Other information**:


src/tools/annotation/LengthTool.js
```diff
@@ -228,11 +228,12 @@ export default class LengthTool extends BaseAnnotationTool {
     function textBoxText(data, rowPixelSpacing, colPixelSpacing) {
       // Set the length text suffix depending on whether or not pixelSpacing is available
       let suffix = ' mm';
-
       if (!rowPixelSpacing || !colPixelSpacing) {
         suffix = ' pixels';
       }
 
+      data.suffix = suffix;
```
evren217: I think it would be better to keep measurement units in `data.unit` instead of `data.suffix` without the leading white space, so it will be more flexible to use in the application level.

jfmedeiros1820: Done

src/tools/annotation/CircleRoiTool.js
```diff
@@ -324,20 +333,19 @@ function _createTextBoxContent(
...
-  const showHounsfieldUnits = options.showHounsfieldUnits !== false;
   const textLines = [];
 
   // Don't display mean/standardDev for color images
   const otherLines = [];
 
   if (!isColorImage) {
     const hasStandardUptakeValues = meanStdDevSUV && meanStdDevSUV.mean !== 0;
-    const suffix = modality === 'CT' && showHounsfieldUnits ? ' HU' : '';
+    const suffix = _getSuffix(modality, options.showHounsfieldUnits);
```
evren217: Could you rename this to `unit` and `_getUnit` as well?

evren217:
@jfmedeiros1820 The changes look good to me now. I am happy to merge it after you fix the build/lint issues. Thanks!

PR #1045, opened 2019-08-13
vNext-CorrectionTool

JamesAPetts:
Correction Tool for `vNext` using a slightly modified Tobias Heimann Correction Algorithm in order to try to do it quick in the browser on the CPU.

Using the stroke given, determine which action(s) to perfom:
- Stroke starts and ends inside a segmentation: Behaves as an subtractive freehand scissors.
- Stroke starts and ends outside a segmentation: Behaves as an additive freehand scissors.
- Stroke out-in-out: Section is subtracted.
- Stroke in-out-in: Section is added.

It could further be improved by multithreading operation chains with workers.

### DIFF
--- a/src/util/segmentation/operations/correction.js
+++ b/src/util/segmentation/operations/correction.js
@@ -51,7 +51,7 @@ export default function correction(
 }
 
 /**
- * snapPointsToGrid - Snap the freehand points to the labelmap grid and attach a label for each node.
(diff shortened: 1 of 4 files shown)


JamesAPetts:
Going to pragmatically push this for now as its vNext, not master, and we want to get vNext going.

Anyone is of course still free to open/give me feedback.

PR #1048, opened 2019-08-16
vNext - Seg Canvas Fill + Outline Rendering, Seg Demo + Store Module Refactor Wombo Combo

JamesAPetts:
- Segmentations can now be rendered as just an outline, just fill, or both.
- The segmentation renderer no longer creates image bitmaps using `createImageBitmap`. Instead only the regions occupied by a label are rendered to the canvas using canvas draw methods. This is huge speed up for platforms like Safari with poor `createImageBitmap` support. There are still room for optimisations in the future.
- Can toggle visibility of segments through the `Labelmap3D`'s `visibileSegments` property, which is a boolean array indexed by `segmentIndex`, where `true` is visible and `false` is hidden. If an element is undefined for a segment, it defaults to visible.
- More configuration options for segmentations:
```
const configuration = {
  renderOutline: true,
  renderFill: true,
  renderInactiveLabelmaps: true,
  radius: 10,
  minRadius: 1,
  maxRadius: 50,
  segmentsPerLabelmap: 65535, // Max is 65535 due to using 16-bit Unsigned ints.
  fillAlpha: 0.2,
  fillAlphaInactive: 0.1,
  outlineAlpha: 0.7,
  outlineAlphaInactive: 0.35,
  outlineWidth: 1,
};
```

- The `state` object of a `cornerstoneTools` module has been split into two objects to have a seperation of concerns:
- `state`: internal, only modified by `cornerstoneTools` and `cornerstoneTools` plugins.
- `configuration`: Configuration settings intended to be modified at run time by the host application.

The [live seg demo](https://deploy-preview-1048--cornerstone-tools.netlify.com/netlify-example/brush/) demonstrates most of these configurables.

TLDR:
giphy


src/eventListeners/onImageRenderedBrushEventHandler.js
```diff
@@ -142,9 +145,533 @@ function renderInactiveLabelMaps(
...
+
+  const pixelData = labelmap2D.pixelData;
+  const activeSegmentIndex = labelmap3D.activeSegmentIndex;
+  const lineSegments = [];
+
+  labelmap2D.segmentsOnLabelmap.forEach(segmentIndex => {
+    lineSegments[segmentIndex] = [];
+  });
+
+  // TEMP - Do this in a cleaner way.
```
JamesAPetts: Not sure if this _is_ the right way for now. `segmentsOnLabelmap` is set at the end of a brush stroke, so you need to render its outline before its set. For context, the intent of `segmentsOnLabelmap` is to have a quick way to check what memory needs to be allocated for things like DICOM seg export.

swederik: I don't think anyone is going to export mid brush stroke so I guess that's fine.

JamesAPetts:
Just fixing tests.

PR #1051, opened 2019-08-20
New Feature: Allow custom Textbox content for ROI tools

ejkim1996:

Feature

The ROI tools have a default textbox linked to it with calculations.

ROI tools will still have a default textbox, but they can be hidden. Custom text can also be added to the textbox content.

No

* **Other information**:
I am currently using custom versions of the ROI tools that I only need in order to change the textbox content.
To hide default content, set showDefaultContent to false in the configuration of any ROI tool.
To add custom string, set addToTextBoxContent to the string you want to add in the configuration of any ROI tool.
It's my first time doing this, so any feedback would be appreciated!


ejkim1996:
Hi @dannyrb ,

Any updates or suggestions for this PR? Having custom tools just to modify text has introduced bugs for other cornerstone tools and so we are hoping for this update to happen soon. Please let me know if there's anything else I should do.

JamesAPetts:
Such a minor custom feature really feels like something that should just use the plugin framework to extend the ROI tools you need to, and add the custom text hooks to the tool's `configuration` and `renderToolData` methods.

> Having custom tools just to modify text has introduced bugs for other cornerstone tools

If the plugin architecture is causing you some difficulties, it might be useful to give us examples of these issues, as this is the best attack vector from which to solve your problem. The idea of the plugin framework in general is having a nice middleground between keeping the library clean/light, and preventing people from hardforking, so I'd be very interested in the difficulties around your usecase here.

PR #1059, opened 2019-09-04
Cornerstone Tools 4.0: Undo/Redo for labelmaps.

JamesAPetts:
A history of undo and redo actions are stored per `labelmap3D`.

- All core tools in the lib now save a diff to the undo stack after usage (for brush tools, after the brush stroke).

- Two public methods added to the segmentation module:
  - `setters.undo(element, labelmapIndex)` - Undoes one action, and adds to the redo stack. If `labelmapIndex` is not set the `activeLabelmapIndex` is used.
  - `setters.redo(element, labelmapIndex)` - Redoes one action, and adds to the undo stack. If `labelmapIndex` is not set the `activeLabelmapIndex` is used.

- Diffs are calculated for each `Labelmap2D` affected, if multiple frames are affected, you just store an array of the diffs for each. This gives very simple API for all 2D tools, and 3D tools.
- Diffs poped and pushed between undo and redo stacks. The redo stack is invalidated upon a new action.
- Most actions currently in cornerstoneTools are currently 2D, so storing time scales on slice length (i.e. `width * height`), spherical brush is `width * height` * the number of slices painted on.

Scope to compress changes per-slice + multithread easily in the future, to have better support for large-scale 3D operations if improved performance is required.

[live demo here](https://deploy-preview-1059--cornerstone-tools.netlify.com/netlify-example/brush/). `ctrl/cmd + z` is undo and `ctrl/cmd + y` is redo.

### DIFF
--- a/docs/latest/modules/segmentation.md
+++ b/docs/latest/modules/segmentation.md
@@ -91,7 +91,7 @@ const shouldErase = false;
 // Imported from src/util/segmentation
 drawBrushPixels(
   pointerArray,
-  labelmap2D,
+  labelmap2D.pixelData,
   labelmap3D.activeSegmentIndex,
   columns,
   shouldErase
--- a/netlify-example/brush/index.html
+++ b/netlify-example/brush/index.html
@@ -1,631 +1,615 @@
 <!DOCTYPE html>
 <html>
-  <head>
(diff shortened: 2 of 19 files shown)


JamesAPetts:
TODO: Test suite.

Issue #1054, opened 2019-08-23
Bidirectional tool is not perpendicular with non-square pixels


alejandracascru:
### Prerequisites

* [ ] Are you running the latest version?
* [ ] Are you reporting to the correct repository?
* [ ] Did you perform a cursory search?

For more information, see the `CONTRIBUTING` guide.

### Description

When I draw the bidirectional tool over images with non-square pixel sizes, the tool doesn't keep the perpendicularity. 

### Steps to Reproduce

1. Load an image with non-square pixel sizes.
2. Use the bidirectional tool.

**Expected behavior:** [What you expected to happen]
The two lines should be perpendicular.
**Actual behavior:** [What actually happened]
The two lines are not perpendicular.

![Screenshot from 2019-08-23 11-33-03](https://user-images.githubusercontent.com/39317465/63608298-3cf78280-c59a-11e9-9215-cbabafc4b5f0.png)

### Codepen With Reproduction of Issue:

- [Basic Template](https://codepen.io/dannyrb/pen/OQQwza)

dannyrb:
@alejandracascru are you able to include a CodePen or CodeSandbox demo with an image that has non-square pixels that demonstrates this behavior?

alejandracascru:
> @alejandracascru are you able to include a CodePen or CodeSandbox demo with an image that has non-square pixels that demonstrates this behavior?

Sure, I will try to put this image in a CodePen. Thanks for the quick reply.

PR #1063, opened 2019-09-10
fix: 1054 perpendicular bidirectional tool on non-square pixel images

brunoalvesdefaria:

This PR fixes the issue #1054 where the bidirectional tool was not being created with its lines perpendicular for images with non-square pixels.
The angle between its lines were always assuming an angle different from 90º.
The issue was happening both when the users were creating a bidirectional tool measurement or editing an existing one.

With this fix, most part of the complexity and duplicated code was removed from the logic responsible to calculate the handles position when creating or editing bidirectional tool measurements.


netlify-example/imageLoader.js
```diff
@@ -45,23 +51,23 @@
...
       rows: height,
       columns: width,
-      height: height,
-      width: width,
+      height,
+      width,
       color: false,
       columnPixelSpacing: 0.8984375,
-      rowPixelSpacing: 0.8984375,
+      rowPixelSpacing: 0.3,
```
dannyrb: Is this the only _real_ change for this file?

I'm not sure if we have any tests that use this, but it might be worth adding a new image that actually has non-square pixel spacing? Something semantic, so it's more obvious what the difference between each is, so we can better use them and observe behavior.

For example:
- keep: `imageId === 'example://1` 
- add: `imageId === 'example://non-square-spacing'`

Also, some of these `let`s can be `const`s

src/tools/annotation/bidirectionalTool/utils/updatePerpendicularLineHandles.js
```diff
@@ -21,19 +24,19 @@ export default function(eventData, data) {
...
     const perpendicularLineLength = length / 2;
 
-    startX = mid.x + (perpendicularLineLength / 2) * vectorY;
-    startY = mid.y - (perpendicularLineLength / 2) * vectorX;
-    endX = mid.x - (perpendicularLineLength / 2) * vectorY;
-    endY = mid.y + (perpendicularLineLength / 2) * vectorX;
+    startX = mid.x + (perpendicularLineLength / (2 * cps)) * vectorY;
+    startY = mid.y - (perpendicularLineLength / (2 * rps)) * vectorX;
+    endX = mid.x - (perpendicularLineLength / (2 * cps)) * vectorY;
+    endY = mid.y + (perpendicularLineLength / (2 * rps)) * vectorX;
```
dannyrb: I take it this was the culprit causing the askew rendering?

src/tools/annotation/bidirectionalTool/utils/updatePerpendicularLineHandles.js
```diff
@@ -7,6 +7,9 @@ export default function(eventData, data) {
   let startX, startY, endX, endY;
 
   const { start, end } = data.handles;
+  const { columnPixelSpacing, rowPixelSpacing } = eventData.image;
```
dannyrb: I think in this instance, that this is fine because the `cps` and `rps` are used near this, and the meaning is pretty clear. In general, I like keeping the longer, descriptive variable names.

`const { columnPixelSpacing = 1, rowPixelSpacing = 1 } = eventData.image;`

Would be a way to destructure this with default values.

netlify-example/imageLoader.js
dannyrb: Note skewed demo image with new spacing:

image

src/tools/annotation/bidirectionalTool/moveHandle/getBaseData.js
```diff
@@ -0,0 +1,34 @@
+import external from '../../../../externalModules.js';
+import getDistance from './getDistance.js';
+
+export default function getBaseData(data, eventData, fixedPoint) {
```
dannyrb: What is `data`?
Instead of passing in _all_ of `eventData`, can we scope this to:

```js
// optional
{
   columnPixelSpacing: int,
   rowPixelSpacing: int
}
```

What is the type/structure of `fixedPoint`?

src/tools/annotation/bidirectionalTool/moveHandle/getBaseData.js
```diff
@@ -0,0 +1,34 @@
...
+    rps,
+    start,
+    end,
+    perpendicularStart,
+    perpendicularEnd,
+    longLine,
+    intersection,
+    distanceToFixed,
+    fixedPoint,
+  };
```
dannyrb: I haven't used this tool much. For my own understanding:

```js
  return {
    cps,    // column pixel spacing
    rps,    // row pixel spacing
    start,   // start of the "primary" line?
    end,    // end of the "primary" line?
    perpendicularStart,  // start of "secondary" line?
    perpendicularEnd,   // end of "secondary" line?
    longLine,  // If this is just { start, end }, why return it?
    intersection,  // point where "primary" and "secondary" lines intersect
    distanceToFixed,  // distance to fixed point from the intersection?
    fixedPoint,  // the point opposite the grabbed handle? "Fixed"
  };
```

src/tools/annotation/bidirectionalTool/moveHandle/getLineVector.js
```diff
@@ -0,0 +1,12 @@
+export default function getLineVector(cps, rps, pointA, pointB) {
```
dannyrb: Please use more descriptive names. JSDoc also appreciated.

A little surprised we don't have _something_ like this already lying around.

src/tools/annotation/bidirectionalTool/moveHandle/moveLongLine.js
```diff
@@ -0,0 +1,49 @@
...
+
+  // Calculate the new intersection point
+  const k = distanceToFixed / newLineLength;
+  const newIntersection = {
+    x: fixedPoint.x + (proposedPoint.x - fixedPoint.x) * k,
+    y: fixedPoint.y + (proposedPoint.y - fixedPoint.y) * k,
+  };
+
+  // Calculate and set the new position of the perpendicular handles
+  updateLine(baseData, newIntersection, data);
```
dannyrb: Instead of updating by reference, can we make this a a bit more pure?

`updateLine` should return instead of set:

This:

```js
  perpendicularStart.x = mid.x + vector.y * distancePS * rps * multiplier;
  perpendicularStart.y = mid.y + vector.x * distancePS * cps * multiplier * -1;
  perpendicularEnd.x = mid.x + vector.y * distancePE * rps * multiplier * -1;
  perpendicularEnd.y = mid.y + vector.x * distancePE * cps * multiplier;
```

Becomes:

```js
// or any other shape that works for you
return {
  start: {
    x: mid.x + vector.y * distancePS * rps * multiplier,
    y: mid.y + vector.x * distancePS * cps * multiplier * -1
  },
  end: {
    x: mid.x + vector.y * distancePE * rps * multiplier * -1,
    y: mid.y + vector.x * distancePE * cps * multiplier
  }
}
```

Then that is likely what `moveLongLine` would also return. Then use those returned values to update appropriately wherever `moveLongLine` is called.

> A pure function doesn’t depend on and doesn’t modify the states of variables out of its scope.
>
> Concretely, that means a pure function always returns the same result given same parameters. Its execution doesn’t depend on the state of the system.

Much easier to test. Much easier to debug.

src/tools/annotation/bidirectionalTool/moveHandle/moveLongLine.js
```diff
@@ -0,0 +1,49 @@
...
+
+  const newLineLength = getDistance(cps, rps, fixedPoint, proposedPoint);
+
+  // Stop here if the handle tries to move before the intersection point
+  if (newLineLength <= distanceToFixed) {
+    return false;
+  }
+
+  // Calculate the new intersection point
+  const k = distanceToFixed / newLineLength;
```
dannyrb: k?

src/tools/annotation/bidirectionalTool/utils/updatePerpendicularLineHandles.js
brunoalvesdefaria: Only for the measurement creation.
The worst part was in the logic for modifying the lines when the user drags a handle.

src/tools/annotation/bidirectionalTool/moveHandle/moveLongLine.js
brunoalvesdefaria: I'm not sure what that `k` means too.
It was migrated from old code.

src/tools/annotation/bidirectionalTool/moveHandle/movePerpendicularLine.js
```diff
@@ -0,0 +1,87 @@
+import external from './../../../../externalModules.js';
+import getDistance from './getDistance.js';
+import getLineVector from './getLineVector.js';
+import getBaseData from './getBaseData.js';
+
+function isFixedEnd(baseData) {
```
dannyrb: instead of taking all of `baseData` reduce to data required?

src/tools/annotation/bidirectionalTool/moveHandle/movePerpendicularLine.js
```diff
@@ -0,0 +1,87 @@
...
+  const { fixedPoint, perpendicularEnd } = baseData;
+
+  return fixedPoint === perpendicularEnd;
+}
+
+function getMultiplier(baseData) {
+  return isFixedEnd(baseData) ? -1 : 1;
+}
+
+function getMovingPoint(baseData) {
```
dannyrb: For all of these methods, really, we should try to author more explicit signatures

src/tools/annotation/bidirectionalTool/moveHandle/movePerpendicularLine.js
```diff
@@ -0,0 +1,87 @@
...
+  // Find the new intersection in the long line
+  const newIntersection = lineSegment.intersectLine(longLine, helperLine);
+
+  // Stop the flow here if there's no intersection point between lines
+  if (!newIntersection) {
+    return false;
+  }
+
+  // Change the position of the perpendicular line handles
+  updateLine(baseData, newIntersection, helperLine, vector);
```
dannyrb: Similar to our `moveLine` method, we should probably make these "pure"

dannyrb:
@brunoalvesdefaria, please feel free to ping me if you have questions/pushback.

src/tools/annotation/bidirectionalTool/utils/updatePerpendicularLineHandles.js
brunoalvesdefaria: Done.
Thank you!

src/tools/annotation/bidirectionalTool/moveHandle/moveLongLine.js
```diff
@@ -0,0 +1,49 @@
...
+
+  // Calculate and set the new position of the perpendicular handles
+  perpendicularStart.x = mid.x + vector.y * distancePS * rps * multiplier;
+  perpendicularStart.y = mid.y + vector.x * distancePS * cps * multiplier * -1;
+  perpendicularEnd.x = mid.x + vector.y * distancePE * rps * multiplier * -1;
+  perpendicularEnd.y = mid.y + vector.x * distancePE * cps * multiplier;
+}
+
+// Move long line handle
+export default function(proposedPoint, data, eventData, fixedPoint) {
```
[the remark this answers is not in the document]
brunoalvesdefaria: Done.
Thank you!

src/tools/annotation/bidirectionalTool/moveHandle/movePerpendicularLine.js
brunoalvesdefaria: Done.
Thank you!

src/tools/annotation/bidirectionalTool/moveHandle/getBaseData.js
brunoalvesdefaria: I renamed that to `toolData`.
Thank you!

src/tools/annotation/bidirectionalTool/moveHandle/getDistance.js
```diff
@@ -0,0 +1,6 @@
+export default function getDistance(cps, rps, pointA, pointB) {
```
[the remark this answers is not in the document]
brunoalvesdefaria: I renamed that function to `getDistanceWithPixelSpacing` and placed it under `utils` directory.
Thank you!

src/tools/annotation/bidirectionalTool/moveHandle/getLineVector.js
brunoalvesdefaria: I renamed the parameters of that function, moved it under `utils` folder and added JSDoc.
Actually we have a lot of bugs related to column and row pixel spacing in other tools.
Using the image with non-square pixels we are able to see a lot of bugs in other tools (even on simple tools like angle).

src/tools/annotation/bidirectionalTool/moveHandle/movePerpendicularLine.js
brunoalvesdefaria: Changed all the methods and added separate files and unit test files for them.
Also documented them all with JSDoc.
Thank you!

src/tools/annotation/bidirectionalTool/moveHandle/movePerpendicularLine.js
brunoalvesdefaria: Done.
Thank you!

src/tools/annotation/bidirectionalTool/moveHandle/moveLongLine.js
brunoalvesdefaria: Done.
Thank you!

src/tools/annotation/bidirectionalTool/moveHandle/getBaseData.js
brunoalvesdefaria: I've added comments to the returned attributes.
Thank you!

netlify-example/imageLoader.js
brunoalvesdefaria: I've changed that in this branch only for tests purposes.
The PR was still a draft.
I've just reverted it back to the original values.

src/tools/annotation/bidirectionalTool/moveHandle/getBaseData.test.js
```diff
@@ -0,0 +1,102 @@
+/* eslint-disable no-eval */
+import getBaseData from './getBaseData';
+
+jest.mock('./../../../../externalModules.js', () => {
+  const intersectLine = eval(
```
dannyrb: I'm not sure I understand what's going on here. Why the `eval`?

src/tools/annotation/bidirectionalTool/moveHandle/getBaseData.test.js
```diff
@@ -0,0 +1,102 @@
+/* eslint-disable no-eval */
+import getBaseData from './getBaseData';
+
+jest.mock('./../../../../externalModules.js', () => {
```
dannyrb: We can probably mock these in `__mocks__`, given that they're all external dependencies?

src/tools/annotation/bidirectionalTool/moveHandle/getBaseData.test.js
```diff
@@ -0,0 +1,102 @@
...
+    expect(result.start).toEqual(start);
+    expect(result.end).toEqual(end);
+    expect(result.perpendicularStart).toEqual(perpendicularStart);
+    expect(result.perpendicularEnd).toEqual(perpendicularEnd);
+    expect(result.longLine.start).toEqual(start);
+    expect(result.longLine.end).toEqual(end);
+    expect(result.intersection.x).toEqual(4);
+    expect(result.intersection.y).toEqual(4);
+    expect(result.distanceToFixed).toEqual(2);
+    expect(result.fixedPoint).toEqual(fixedPoint);
```
dannyrb: Instead of this, can you do object value equality? Does "expectedValue" equal "actualValue"?

dannyrb: Make sure you do an value compare, and not an identity compare

package.json
```diff
@@ -39,6 +39,7 @@
     "start": "npm run build:prod && npm run build:dev",
     "test": "npm run test:unit",
     "test:unit": "jest --clearCache && jest --runInBand \\.test\\.js$",
+    "test:tool": "jest --clearCache && jest --runInBand .*bidirectionalTool.*\\.test\\.js$",
```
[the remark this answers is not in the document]
brunoalvesdefaria: I've already removed that here:
https://github.com/cornerstonejs/cornerstoneTools/pull/1063/commits/7d120099617971fff9431b672f010c5f408f5588

It seems that you got an outdated version on review

src/tools/annotation/bidirectionalTool/moveHandle/getBaseData.test.js
```diff
@@ -0,0 +1,102 @@
...
+
+jest.mock('./../../../../externalModules.js', () => {
+  const intersectLine = eval(
+    "(a,b)=>{function sign(x){return typeof x==='number'?x?x<0?-1:1:x===x?0:NaN:NaN}const i={};let x1=a.start.x,y1=a.start.y,x2=a.end.x,y2=a.end.y,x3=b.start.x,y3=b.start.y,x4=b.end.x,y4=b.end.y;let a1,a2,b1,b2,c1,c2;let r1,r2,r3,r4;let d,n;a1=y2-y1;b1=x1-x2;c1=x2*y1-x1*y2;r3=a1*x3+b1*y3+c1;r4=a1*x4+b1*y4+c1;if(r3!==0&&r4!==0&&sign(r3)===sign(r4)){return}a2=y4-y3;b2=x3-x4;c2=x4*y3-x3*y4;r1=a2*x1+b2*y1+c2;r2=a2*x2+b2*y2+c2;if(r1!==0&&r2!==0&&sign(r1)===sign(r2)){return}d=(a1*b2)-(a2*b1);n=(b1*c2)-(b2*c1);const x=parseFloat(n/d);n=(a2*c1)-(a1*c2);const y=parseFloat(n/d);i.x=x;i.y=y;return i}"
+  );
+
+  return {
+    cornerstoneMath: {
+      lineSegment: {
+        intersectLine: jest.fn(intersectLine),
```
dannyrb: If you're only calling this a few times, you can just `mockReturnValue` or `mockReturnValueOnce`

https://jestjs.io/docs/en/mock-function-api.html#mockfnmockreturnvaluevalue

src/tools/annotation/bidirectionalTool/moveHandle/getBaseData.test.js
```diff
@@ -0,0 +1,102 @@
...
+
+function createPoint(x, y) {
+  return {
+    x,
+    y,
+  };
+}
+
+describe('getBaseData.js', () => {
+  it('base data is extracted with all attributes', () => {
```
dannyrb: `it('base data is extracted with all attributes'` doesn't read naturally. Try to phrase things conversationally:

`it('returns a BaseDataObject with the correct values'`

src/tools/annotation/bidirectionalTool/moveHandle/getBaseData.test.js
```diff
@@ -0,0 +1,102 @@
...
+    expect(result.perpendicularEnd).toEqual(perpendicularEnd);
+    expect(result.longLine.start).toEqual(start);
+    expect(result.longLine.end).toEqual(end);
+    expect(result.intersection.x).toEqual(4);
+    expect(result.intersection.y).toEqual(4);
+    expect(result.distanceToFixed).toEqual(2);
+    expect(result.fixedPoint).toEqual(fixedPoint);
+  });
+
+  it('base data is extracted with row and column pixel spacing defaulted to 1', () => {
```
dannyrb: Same feedback for this test :+1:

src/tools/annotation/bidirectionalTool/moveHandle/getBaseData.test.js
```diff
@@ -0,0 +1,102 @@
+/* eslint-disable no-eval */
+import getBaseData from './getBaseData';
```
dannyrb: Is `getBaseData` public API?

brunoalvesdefaria: Nope, it's being used only as a helper for moving handles from bidirectional tool lines.

src/tools/annotation/bidirectionalTool/moveHandle/getBaseData.test.js
brunoalvesdefaria: Done.
Thank you!

src/tools/annotation/bidirectionalTool/moveHandle/getBaseData.test.js
brunoalvesdefaria: Done.
Thank you!

src/tools/annotation/bidirectionalTool/moveHandle/getBaseData.test.js
```diff
@@ -0,0 +1,116 @@
...
+    expect(result.perpendicularEnd.y).toEqual(2);
+    expect(result.longLine.start.x).toEqual(0);
+    expect(result.longLine.start.y).toEqual(4);
+    expect(result.longLine.end.x).toEqual(8);
+    expect(result.longLine.end.y).toEqual(4);
+    expect(result.intersection.x).toEqual(4);
+    expect(result.intersection.y).toEqual(4);
+    expect(result.distanceToFixed).toEqual(2);
+    expect(result.fixedPoint.x).toEqual(8);
+    expect(result.fixedPoint.y).toEqual(4);
```
dannyrb: It looks like my comment was deleted, but this should be object comparison, not per prop comparison

src/tools/annotation/bidirectionalTool/moveHandle/getBaseData.test.js
```diff
@@ -0,0 +1,116 @@
...
+  });
+
+  it('baseData object is returned with row and column pixel spacing defaulted to 1', () => {
```
dannyrb: this sentence could use some love as well

src/tools/annotation/bidirectionalTool/moveHandle/moveLongLine/getDistanceToIntersection.test.js
```diff
@@ -0,0 +1,44 @@
...
+
+    const baseData = {
+      columnPixelSpacing,
+      rowPixelSpacing,
+      intersection,
+    };
+
+    const distance = getDistanceToIntersection(baseData, point);
+
+    expect(distance).toEqual(8.94427190999916);
```
dannyrb: Is that float precision? I think you want to use `.toBeCloseTo` https://jest-bot.github.io/jest/docs/expect.html#tobeclosetonumber-numdigits

src/tools/annotation/bidirectionalTool/moveHandle/moveLongLine/getDistanceToIntersection.test.js
```diff
@@ -0,0 +1,44 @@
...
+
+function createPoint(x, y) {
+  return {
+    x,
+    y,
+  };
+}
+
+describe('getDistanceToIntersection.js', () => {
+  it('distance is positive', () => {
```
dannyrb: Doesn't read well. You can use `it('description')` or `test('description')`, but it should read like a simple requirement.

`it('returns a positive distance, regardless of point order')` or...
`test('returned distance is positive')`

src/tools/annotation/bidirectionalTool/moveHandle/moveLongLine/getDistanceToIntersection.js
```diff
@@ -0,0 +1,21 @@
...
+ * Returns the distance between the intersection and the given point
+ * considering the row and column pixel spacings.
+ *
+ * @param {*} baseData Base data for bidirectional line moving
+ * @param {*} point Point which the distance to intersection will be calculated
+ *
+ * @returns {number} Returns the distance betwen the intersection and point
+ */
+export default function getDistanceToIntersection(baseData, point) {
+  const { columnPixelSpacing, rowPixelSpacing, intersection } = baseData;
```
dannyrb: The constructor doesn't need to take the entire `baseData` object if we're only using these three values from it.

dannyrb: Follow-up question. Why does this method exist? It looks like all it does is map data to call a different utility function? I don't know that it adds any value.

src/tools/annotation/bidirectionalTool/moveHandle/getBaseData.test.js
brunoalvesdefaria: I was mocking the original implementation of intersectLine from `cornerstoneMath`.
Just changed that and started mocking the returned value for the test instead.
Thank you!

src/tools/annotation/bidirectionalTool/moveHandle/getBaseData.test.js
brunoalvesdefaria: I was mocking the original implementation of intersectLine from `cornerstoneMath`.
Just changed that and started mocking the returned value for the test instead.
Thank you!

src/tools/annotation/bidirectionalTool/moveHandle/getBaseData.test.js
brunoalvesdefaria: I was mocking the original implementation of intersectLine from `cornerstoneMath`.
Just changed that and started mocking the returned value for the test instead.
Thank you!

src/tools/annotation/bidirectionalTool/moveHandle/moveLongLine/getDistanceToIntersection.js
brunoalvesdefaria: This function was doing something else then I refactored and that changed a bit.
It was behaving basically as a proxy function.
Then I removed it.

src/tools/annotation/bidirectionalTool/moveHandle/getBaseData.test.js
brunoalvesdefaria: Changed to: `ensure that baseData object is returned with default row and column pixel spacing`

Thank you!

src/tools/annotation/bidirectionalTool/moveHandle/moveLongLine/getDistanceToIntersection.test.js
brunoalvesdefaria: This function its test has been removed.
But thanks for the review.

src/tools/annotation/bidirectionalTool/moveHandle/moveLongLine/getDistanceToIntersection.test.js
brunoalvesdefaria: This test was also deleted as impact of refactoring.
But I'll observe it in the other tests and change them to use this assertion.
Thank you!

src/tools/annotation/bidirectionalTool/moveHandle/moveLongLine/moveLongLine.js
```diff
@@ -0,0 +1,60 @@
...
+  fixedPoint
+) {
+  const baseData = getBaseData(measurementData, eventData, fixedPoint);
+  const {
+    columnPixelSpacing,
+    rowPixelSpacing,
+    distanceToFixed,
+    perpendicularStart,
+    perpendicularEnd,
+  } = baseData;
```
dannyrb: You can one-line this (but don't have to):

```js
  const {
    columnPixelSpacing,
    rowPixelSpacing,
    distanceToFixed,
    perpendicularStart,
    perpendicularEnd,
  } = getBaseData(measurementData, eventData, fixedPoint);
```

src/tools/annotation/bidirectionalTool/moveHandle/moveLongLine/moveLongLine.js
```diff
@@ -0,0 +1,60 @@
...
+    proposedPoint
+  );
+
+  // Stop here if the handle tries to move before the intersection point
+  if (newLineLength <= distanceToFixed) {
+    return false;
+  }
+
+  // Calculate the new intersection point
+  const k = distanceToFixed / newLineLength;
```
dannyrb: Can you use a more descriptive variable name?

src/tools/annotation/bidirectionalTool/moveHandle/moveLongLine/moveLongLine.js
```diff
@@ -0,0 +1,60 @@
...
+  };
+
+  // Calculate and the new position of the perpendicular handles
+  const newLine = updatePerpendicularLine(baseData, newIntersection);
+
+  // Update the perpendicular line handles
+  perpendicularStart.x = newLine.start.x;
+  perpendicularStart.y = newLine.start.y;
+  perpendicularEnd.x = newLine.end.x;
+  perpendicularEnd.y = newLine.end.y;
```
dannyrb: I'm a little confused. This seems to be updating the values returned by `getBaseData`, but those changes aren't returned. So this is either updating by reference (not pure), or having no effect.

src/tools/annotation/bidirectionalTool/moveHandle/moveLongLine/moveLongLine.js
brunoalvesdefaria: Actually I can but it's not a good idea.
If I do so, I'll be creating lots of warnings due to the ESLint restriction of 80 chars per line.
But those line breaks are also causing a lot of LoC issues on CodeClimate: https://codeclimate.com/github/cornerstonejs/cornerstoneTools/pull/1063#issues

src/tools/annotation/bidirectionalTool/moveHandle/getBaseData.test.js
brunoalvesdefaria: Done.
Thank you!

src/tools/annotation/bidirectionalTool/moveHandle/moveLongLine/updatePerpendicularLine.js
```diff
@@ -0,0 +1,64 @@
...
+export default function updatePerpendicularLine(baseData, mid) {
+  const {
+    columnPixelSpacing,
+    rowPixelSpacing,
+    start,
+    perpendicularStart,
+    perpendicularEnd,
+    intersection,
+    fixedPoint,
+  } = baseData;
```
dannyrb: This does use more/most of `baseData` than other methods. I'm not opposed to passing the entire object if it makes your life easier, but having a JSDoc definition for the object would add some clarity. If we'd rather not maintain the `@typdef`, then destructuring in the signature might be a preferable alternative.

src/tools/annotation/bidirectionalTool/moveHandle/moveLongLine/moveLongLine.js
brunoalvesdefaria: That's a good point.
I've changed that to update the attributes from `measurementData` directly.
Thank you!

dannyrb: It looks like `updatePerpendicularLine` is a pure function, so if this code is working as expected, that may mean `getBaseData`'s `perpendicularStart` and `perpendicularEnd` are references to the original handle data, and that these assignments update that reference. (That's a scary side-effect, in my opinion)

src/tools/annotation/bidirectionalTool/moveHandle/movePerpendicularLine/movePerpendicularLine.js
```diff
@@ -66,10 +64,10 @@ export default function movePerpendicularLine(
...
 
   // Change the position of the perpendicular line handles
-  perpendicularStart.x = newLine.start.x;
-  perpendicularStart.y = newLine.start.y;
-  perpendicularEnd.x = newLine.end.x;
-  perpendicularEnd.y = newLine.end.y;
+  measurementData.handles.perpendicularStart.x = newLine.start.x;
+  measurementData.handles.perpendicularStart.y = newLine.start.y;
+  measurementData.handles.perpendicularEnd.x = newLine.end.x;
+  measurementData.handles.perpendicularEnd.y = newLine.end.y;
```
dannyrb: This is still updating by reference ;)

src/tools/annotation/bidirectionalTool/moveHandle/moveLongLine/moveLongLine.js
brunoalvesdefaria: I've renamed it to `distanceRatio` to make it more meaningful.
Thank you!

src/tools/annotation/bidirectionalTool/moveHandle/movePerpendicularLine/movePerpendicularLine.js
brunoalvesdefaria: As per method description in JSDoc `Move the long line updating the perpendicular line handles position.`, I think it should update by reference.

How do you suggest to improve that implementation?
Because if I move to the parent function, it will still update by reference.

src/tools/annotation/bidirectionalTool/utils/updatePerpendicularLineHandles.js
```diff
@@ -1,12 +1,27 @@
...
+ * Update the perpendicular line handles when the measurement is being created.
+ * This method will make the perpendicular line intersect in the middle of the
+ * long line and assume half the size of the long line.
+ *
+ * @param {*} eventData Data object associated with the event
+ * @param {*} measurementData Data from current bidirectional tool measurement
+ *
+ * @returns {boolean} False in case the handle is not locked or true when moved
+ */
+export default function updatePerpendicularLineHandles(
```
dannyrb: `getUpdatedPerpendicularLineHandles`?

src/tools/annotation/bidirectionalTool/moveHandle/moveLongLine/moveLongLine.js
dannyrb: The `CodeClimate` restriction is arbitrary (afaik). We can bump it or ignore it.
A lot of static analysis is "hey, this is a little odd and could be a red flag, can you take a quick look?" Conceptually, I think having explicit knowledge of the variables your function requires is a cognitive victory.

Certainly open for debate -- happy to hear your thoughts.

src/tools/annotation/bidirectionalTool/moveHandle/movePerpendicularLine/movePerpendicularLine.js
dannyrb: I think I would expect any changes by reference to happen in the tool. Basically, wherever we "fetch" the reference, we should update it by reference (if we have to), in the same method. Isolating those kinds of changes to the tool itself feels logical.

I'm not opposed to alternative implementations -- feel free to discuss pros/cons of other approaches you might want to take.

brunoalvesdefaria: I just don't want to touch/refactor the upper levels of the tool.
Otherwise what should be a quick fix will become a huge refactor (which is already turning into).

src/tools/annotation/bidirectionalTool/utils/updatePerpendicularLineHandles.js
brunoalvesdefaria: The same applies here.
I don't want to touch upper levels of the tool code structure.

src/tools/annotation/bidirectionalTool/utils/updatePerpendicularLineHandles.js
```diff
@@ -20,24 +35,28 @@ export default function(eventData, data) {
...
   }
 
-  data.handles.perpendicularStart.x = startX;
-  data.handles.perpendicularStart.y = startY;
-  data.handles.perpendicularEnd.x = endX;
-  data.handles.perpendicularEnd.y = endY;
+  measurementData.handles.perpendicularStart.x = startX;
+  measurementData.handles.perpendicularStart.y = startY;
+  measurementData.handles.perpendicularEnd.x = endX;
+  measurementData.handles.perpendicularEnd.y = endY;
```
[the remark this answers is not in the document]
brunoalvesdefaria: The same again.
I don't want to touch upper levels of the tool code structure.
It might have some undesired impacts and would add a lot of effort for me to test everything again.

src/tools/annotation/bidirectionalTool/moveHandle/movePerpendicularLine/movePerpendicularLine.js
dannyrb: That's fine. We can table this for a different day :+1:

PR #1067, opened 2019-09-12
fix(FreehandRoiSculptorTool.js): Use merged props value over default

mikehazell:
The FreehandRoiSculptor tool needs to be able to reference the state of the FreehandRoiTool to modify measurements. This reference is configured via the `referencedToolName` prop. The current implementation defines this as configurable API but then uses the default value instead of the merged value.

- [-] Tests for the changes have been added (for bug fixes / features)
- [-] Docs have been added / updated (for bug fixes / features)

Bug Fix

The `FreehandRoiSculptor` tool uses the default `referencedToolName` instead of the result from merging `props` and `defaultProps`.

Now uses the merged value.

No

* **Other information**:

### DIFF
--- a/src/tools/FreehandRoiSculptorTool.js
+++ b/src/tools/FreehandRoiSculptorTool.js
@@ -38,7 +38,7 @@ export default class FreehandRoiSculptorTool extends BaseTool {
 
     this.updateOnMouseMove = true;
     this.isMultiPartTool = true;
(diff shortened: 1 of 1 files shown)


dannyrb:
_wonders what @mikehazell is working on_

Good catch! This looks good to me.
I'm going to go ahead and merge.

PR #1074, opened 2019-09-13
Fix issue 1071 - StackPrefetch stops when selecting different element…

frolic06:
Fix issue 1071 - StackPrefetch stops when selecting different element

### DIFF
--- a/src/stackTools/stackPrefetch.js
+++ b/src/stackTools/stackPrefetch.js
@@ -14,7 +14,7 @@ const requestType = 'prefetch';
 
 let configuration = {
   maxImagesToPrefetch: Infinity,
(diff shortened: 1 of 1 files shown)


dannyrb:
👋 @frolic06, you have two `prettier` (formatting) errors that are preventing a merge:

image

`prettier` is supposed to fix these for you automatically when you add modified files with `git`, but it appears to have failed us here. You can either fix these manually (two small changes), try running `npm run lint`, or if you have prettier installed globally, you can run it against the single file you've changed.

dannyrb:
@all-contributors, please add @frolic06 for their awesome code contribution!

PR #1080, opened 2019-09-18
feat: 🎸 Expose isToolActiveForElement

JamesAPetts:
Renamed isToolActive to isToolActiveForElement and expose it as top
level API.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -147,6 +147,7 @@ import {
   setToolPassive,
   setToolPassiveForElement,
 } from './store/setToolMode.js';
+import isToolActiveForElement from './store/isToolActiveForElement';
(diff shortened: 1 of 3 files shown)


jlopes90:
What do you think?

```
function isToolModeForElement(element, toolName, mode) {
	const tool = getToolForElement(element, toolName);
	
	return tool.mode === mode;
}
```

dannyrb:
@jlopes90, is this something you think you would use? Are there other convenience methods you feel like we're missing?

dannyrb:
Also, almost always better to discuss in a new issue. It's hard to keep track of topics like this on anything closed/merged.

PR #1095, opened 2019-10-04
New Feature: Event Measurement Selected

jlopes90:

Feature

No

## Event Measurement Selected (cornerstonetoolsmeasurementselected)
```
const element = document.querySelector('.cornerstone-element');
const toolType = cornerstoneTools.EVENTS.MEASUREMENT_SELECTED;
element.addEventListener(toolType, function (evt) {
  const eventData = evt.detail;
  console.log(eventData);
});
```

No

* **Other information**:
No


jlopes90:
Some don't work because click is handles:
- [ ] Probe
- [ ] Freehand

shaunakv1:
@dannyrb @jlopes90 Curious as to why this was never merged. I see some coverage checks failing. Is there anything I can help with to move this along?

I have a use case where we are trying to implement a label tool for a measurement. And while approach recommended by @dannyrb  in https://github.com/cornerstonejs/cornerstoneTools/issues/1144 work in the sense that we can detect which measurement is being right clicked on, we cannot get it to "activate" or turn green in simple words.  I might be missing something here, but this happens because we have to put other tools in ` cornerstoneTools.setToolEnabled` mode so that they don't  trap the `cornerstonetoolsmousedown` event.

PR #1099, opened 2019-10-10
Export getROITextBoxCoords

kevinleedrum:
This simply exports `getROITextBoxCoords` so custom ROI tools can use it.


dannyrb:
👋 Hi @kevinleedrum, can you add this to [lib exports](https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/lib.js) instead?

kevinleedrum:
Oops.  Of course.  Fixed and squashed. ✔️

PR #1109, opened 2019-10-17
fix: probe only active

jlopes90:

Probe without need for activated handles

* **Other information**:

### DIFF
--- a/src/tools/annotation/ProbeTool.js
+++ b/src/tools/annotation/ProbeTool.js
@@ -93,7 +93,7 @@ export default class ProbeTool extends BaseAnnotationTool {
       data.handles.end
     );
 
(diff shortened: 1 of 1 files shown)


jlopes90:
[ ] Test
[ ] Review

dannyrb:
@jlopes90, I'm not sure I understand this change? Can you expand on why you've created a PR? If this is something that should be customizable, is there a better approach we can take than bumping the number?

jlopes90:
Sorry, I just analyzed, after all there is no need to change.
I was testing event "measurement selected" when I clicked "handles" and then calculated "firstAnnotationNearPoint" to get index but received "null" because it's not handles, it's "annotationToolsWithMoveableHandles".

PR #1139, opened 2019-11-25
fix #1123 latest version debug mode crash

### LINKED ISSUE #1123: Cannot show debug in console
Hi guys,
Im using v4.6.2. After set debug mode on by command 
`localStorage.setItem("debug", "cornerstoneTools:*");`
the tools was crashed. Can you check this error.

image

DStar2013:
Developer tools in Google chrome.Logging level support Verbose/Info/Warnings/Errors.

- [ X ] The commit message follows our guidelines
- [ X ] Tests for the changes have been added (for bug fixes / features)
- [ X ] Docs have been added / updated (for bug fixes / features)

bug fix

#1123 

N/A

N/A

* **Other information**:
i still wonder why remove lib debug. @dannyrb  this can also be solved by use old way. import debugLib from 'debug'

### DIFF
--- a/src/util/debugLib/index.js
+++ b/src/util/debugLib/index.js
@@ -156,7 +156,7 @@ function formatArgs(args) {
     this.namespace +
     (this.useColors ? ' %c' : ' ') +
     args[0] +
-    (this.useColors ? '%c ' : ' ')}+${debugLib.humanize(this.diff)}`;
+    (this.useColors ? '%c ' : ' ')}+${setupDebug.humanize(this.diff)}`;
 
   if (!this.useColors) {
     return;


dannyrb:
> i still wonder why remove lib debug. @dannyrb this can also be solved by use old way. import debugLib from 'debug'

@DStar2013, it may not have been the best fix, but `debug` was packaged and deployed with an ES6 target. This meant it broke all browsers that don't support ES6 behavior out of the box.

We had two options:

1. Pull in the relatively small source and allow transpile it to ES5 as part of our own bundling process.

OR

2. Add an explicit include rule (or ignore exception) for it in our Webpack loader so that Babel would transpile it.

We opted for the first option, but the second would have worked as well. That's the history 🤷‍♂

DStar2013:
> > i still wonder why remove lib debug. @dannyrb this can also be solved by use old way. import debugLib from 'debug'
> 
> @DStar2013, it may not have been the best fix, but `debug` was packaged and deployed with an ES6 target. This meant it broke all browsers that don't support ES6 behavior out of the box.
> 
> We had two options:
> 
> 1. Pull in the relatively small source and allow transpile it to ES5 as part of our own bundling process.
> 
> OR
> 
> 1. Add an explicit include rule (or ignore exception) for it in our Webpack loader so that Babel would transpile it.
> 
> We opted for the first option, but the second would have worked as well. That's the history 🤷‍♂

～ O..  Got it! 😄

PR #1140, opened 2019-11-27
feat: Add drawHandles prop to all annotation components

igoroctaviano:

Adds drawHandles props to all annotation components which renders handles by default.

Currently, only the ArrowAnnotationTool have this prop.

Now all annotation tools will have the drawHandles prop available.

No breaking changes.

* **Other information**:

### DIFF
--- a/src/tools/FreehandRoiSculptorTool.js
+++ b/src/tools/FreehandRoiSculptorTool.js
@@ -50,18 +50,12 @@ export default class FreehandRoiSculptorTool extends BaseTool {
 
   renderToolData(evt) {
     const eventData = evt.detail;
-
-    if (this.configuration.currentTool === null) {
-      return false;
-    }
-
     const element = eventData.element;
     const config = this.configuration;
-
     const toolState = getToolState(element, this.referencedToolName);
-    const data = toolState.data[config.currentTool];
+    const toolData = toolState.data[config.currentTool];
 
-    if (!data) {
+    if (this.configuration.currentTool === null || !toolData) {
       return false;
     }
 
--- a/src/tools/FreehandRoiSculptorTool.js
+++ b/src/tools/FreehandRoiSculptorTool.js
@@ -65,7 +65,7 @@ export default class FreehandRoiSculptorTool extends BaseTool {
       return false;
     }
 
-    if (this._active) {
+    if (this._active && this.configuration.drawHandles) {
       const context = eventData.canvasContext.canvas.getContext('2d');
       const options = {
         color: this.configuration.dragColor,
@@ -73,14 +73,12 @@ export default class FreehandRoiSculptorTool extends BaseTool {
         handleRadius: this._toolSizeCanvas,
       };
 
-      if (this.configuration.drawHandles) {
-        drawHandles(
-          context,
-          eventData,
-          this.configuration.mouseLocation.handles,
-          options
-        );
-      }
+      drawHandles(
+        context,
+        eventData,
+        this.configuration.mouseLocation.handles,
+        options
+      );
     } else if (this.configuration.showCursorOnHover && !this._recentTouchEnd) {
       this._renderHoverCursor(evt);
     }
--- a/src/tools/FreehandRoiSculptorTool.js
+++ b/src/tools/FreehandRoiSculptorTool.js
@@ -73,12 +73,14 @@ export default class FreehandRoiSculptorTool extends BaseTool {
         handleRadius: this._toolSizeCanvas,
       };
 
-      drawHandles(
-        context,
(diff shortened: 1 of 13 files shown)


src/tools/FreehandRoiSculptorTool.js
```diff
@@ -252,12 +246,14 @@ export default class FreehandRoiSculptorTool extends BaseTool {
...
     };
 
-    drawHandles(
-      context,
-      eventData,
-      this.configuration.mouseLocation.handles,
-      options
-    );
+    if (this.configuration.drawHandles) {
+      drawHandles(
```
dannyrb: I wasn't aware that the sculptor hand handles? It's not a persistent annotation, so we likely don't need this?

src/tools/annotation/AngleTool.js
```diff
@@ -198,7 +201,9 @@ export default class AngleTool extends BaseAnnotationTool {
           drawHandlesIfActive: drawHandlesOnHover,
         };
 
-        drawHandles(context, eventData, data.handles, handleOptions);
+        if (this.configuration.drawHandles) {
+          drawHandles(context, eventData, data.handles, handleOptions);
+        }
```
dannyrb: 👍

src/tools/annotation/BidirectionalTool.js
```diff
@@ -35,6 +35,7 @@ export default class BidirectionalTool extends BaseAnnotationTool {
         getMeasurementLocationCallback: emptyLocationCallback,
         textBox: '',
         shadow: '',
+        drawHandles: true,
```
dannyrb: Does this conflict with `drawHandlesOnHover`?

src/tools/annotation/CircleRoiTool.js
```diff
@@ -43,6 +43,9 @@ export default class CircleRoiTool extends BaseAnnotationTool {
       name: 'CircleRoi',
       supportedInteractionTypes: ['Mouse', 'Touch'],
       svgCursor: circleRoiCursor,
+      configuration: {
+        drawHandles: true,
+      },
```
dannyrb: This is a tricky one. We may want to adopt the `drawHandlesOnHover` approach here. Example:

- `drawHandles` usually works be showing the handle if the _handle_ is hovered
- `drawHandlesOnHover` usually works by showing the handles if `pointNearTool` is true

For a circle, there's no clear "line end" like in a `length` tool. You'd have to trace the entire circle to hunt for where the handle might be.

dannyrb: I think the previous OHIF Meteor version simply ignored ROI tools like this. It continued to show their handles instead of addressing the UX issue here. Only tools, like length, had their handles hidden.

For your current task, you can continue without supporting tools like this, or you can work in support. Whichever you have the appetite for is fine by me.

src/tools/annotation/EllipticalRoiTool.js
```diff
@@ -221,7 +222,10 @@ export default class EllipticalRoiTool extends BaseAnnotationTool {
           'pixel',
           data.handles.initialRotation
         );
-        drawHandles(context, eventData, data.handles, handleOptions);
+
+        if (this.configuration.drawHandles) {
+          drawHandles(context, eventData, data.handles, handleOptions);
+        }
```
dannyrb: Same note as circleRoi

src/tools/annotation/BidirectionalTool.js
igoroctaviano: `// Draw the handles
      const handleOptions = {
        color,
        handleRadius,
        drawHandlesIfActive: drawHandlesOnHover,
      };

      // Draw the handles
      if (this.configuration.drawHandles) {
        drawHandles(context, eventData, data.handles, handleOptions);
      }`

igoroctaviano: So the drawHandlesOnHover will be passed as option to drawHandles function, the drawHandle prop will take precedence over it.

src/tools/annotation/EllipticalRoiTool.js
igoroctaviano: Removed. Thanks.

src/tools/annotation/CircleRoiTool.js
igoroctaviano: Removed. Thanks.

src/tools/FreehandRoiSculptorTool.js
igoroctaviano: 👍  I have removed the conditional along with those other Roi tools mentioned here.

PR #1188, opened 2020-03-12
fix production build (fixes #941)

### LINKED ISSUE #941: Published npm release always showing "dev mode" logger message
It appears that the `NODE_ENV` is set to "development" during the auomated build, so the published npm package is always in dev mode, and the logger message in the console is always being shown.  In other words, [this line](https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/util/logger.js#L6) always becomes `var devMode = "development" === 'development';`

Betalos:

 The issue was caused with multiple build (dev/prod at the same time)

issue with the log

in the production build the log is removed

Update deps (did not cause issues with the build as far as I know)

* **Other information**:


dannyrb:
@Betalos, is the solution here to leave logging in `production`, or to conditionally remove `debug` in production builds? I don't follow how this PR achieves the desired outcome

Betalos:
@dannyrb the issue was that in package.json build script:
```cmd
npm run test && npm run clean:dist && npm run version && npm run build:prod && npm run build:dev
```
it generate the dev build + prod one (prod with .min extention) this was the issue.
the import from cornerstoreTools would pull from dev build not .min (unless we import directly from it ex: 
```typescript
import * as tools from 'cornerstoneTools/dist/cornerstoneTools.min'
```
so in this PR I removed the dev build and added another entry 'cornerstoneTools.min' and a condition on terser plugin to only minimize the file with .min at the end. With this we achieve the desired outcome, the issue with:
```javascript
var devMode = "production" === 'development'; // If we're in dev-mode, tell the user how to get logs
```
instead of 👍 ```javascript
var devMode = "development" === 'development'; // If we're in dev-mode, tell the user how to get logs
```

Betalos:
@dannyrb any progress on this ???

PR #1211, opened 2020-04-07
Fix: Error moving other tools' handles with bidirectional active

brunoalvesdefaria:


JamesAPetts:
Good catch! This prevents that strange behaviour when the tool is active and you interact with other handles, thanks!

PR #991, opened 2019-06-28
fix: toolType to toolName

jlopes90:
* **Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Fix

* **What is the current behavior?** (You can also link to an open issue here)
Measurement Added and Removed, toolType change to toolName.
Because Modified and Completed is toolName.

* **What is the new behavior (if this is a feature change)?**
No

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
No

* **Other information**:
No

PR #1216, opened 2020-04-10
fix: Divergence on toolType/toolName naming convention

brunoalvesdefaria:

  * Bug fix with divergent `toolType`/`toolName` being added to event data.
  * Updated documentation with `toolName` as being the new attribute to be used.
  * Added comments to the code with deprecation notice of `toolType` attribute.

  * Some tools/manipulators were triggering events with `toolType` attribute and others with `toolName` attribute.

  * Until the `toolType` attribute is removed, all the event data are including both `toolType` and `toolName` attributes.

  * No, but it introduce a deprecation notice for the usage of `toolType` attribute.
  * It will be a breaking change when `toolType` is removed for good.

* **Other information**:
  * `toolType` naming is now going to be used to categorize types of tools (e.g. `general`, `annotation`, `stack`, `segmentation`, etc.)

### DIFF
--- a/src/eventListeners/touchEventListeners.js
+++ b/src/eventListeners/touchEventListeners.js
@@ -23,7 +23,7 @@ let lastScale = 1.0,
 const pressDelay = 700,
   pressMaxDistance = 5;
 
-const toolName = 'touchInput';
+const inputName = 'touchInput';
 
 function onTouch(e) {
   const element = e.currentTarget || e.srcEvent.currentTarget;
@@ -578,11 +578,13 @@ function enable(element) {
     element.addEventListener(eventType, onTouch, { passive: false });
   });
 
-  const options = getToolOptions(toolName, element);
+  // TODO: Check why we are using tool options if it's not a tool
+  const options = getToolOptions(inputName, element);
 
   options.hammer = mc;
 
-  setToolOptions(toolName, element, options);
+  // TODO: Check why we are using tool options if it's not a tool
+  setToolOptions(inputName, element, options);
 }
 
 function disable(element) {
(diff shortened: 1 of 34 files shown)


src/eventListeners/touchEventListeners.js
```diff
@@ -23,7 +23,7 @@ let lastScale = 1.0,
 const pressDelay = 700,
   pressMaxDistance = 5;
 
-const toolType = 'touchInput';
+const toolName = 'touchInput';
```
dannyrb, line 26: I'm not convinced this is supposed to be `toolName`. "touchInput" appears to be specific to the interaction that's happening here, this is taking place in the `touchEventListener`

src/stateManagement/stackSpecificStateManager.js
```diff
@@ -11,11 +11,11 @@ import {
  * @constructor newStackSpecificToolStateManager
  * @memberof StateManagement
  *
- * @param  {string[]} toolTypes       The tool types to apply to the stack.
+ * @param  {string[]} toolNames       The tool names to apply to the stack.
```
dannyrb, line 14: This description is odd, I had to double check usage, but `toolNames` does appear to be what we want here.

src/eventListeners/touchEventListeners.js
brunoalvesdefaria: Renamed to `inputName` as it's not a tool.
Also added some `TODO`s as it's calling `setToolOptions` to store `hammer.js` options.

src/stateManagement/stackSpecificStateManager.js
brunoalvesdefaria: Updated the jsDoc giving a better explanation of this parameter.

mikehazell:
This PR resolved #991

PR #1225, opened 2020-04-22
Add default value 1 for the rowPixelSpacing and columnPixelSpacing

aminica:

Bug fix

When there are no pixel spacing data of the image, the values of the angle measurements are not able to be calculated, since the getPixelSpacing function returns null for the rowPixelSpacing and columnPixelSpacing values. 

getPixelSpacing function returns default 1 if there is no pixel spacing data.

No.

* **Other information**:
Default value 1 is added since in the past every tool had it, but due to some refactoring (in which the getPixelSpacing function was created), this scenario was left out.

### DIFF
--- a/src/util/getPixelSpacing.js
+++ b/src/util/getPixelSpacing.js
@@ -9,14 +9,14 @@ export default function getPixelSpacing(image) {
   if (imagePlane) {
     return {
       rowPixelSpacing:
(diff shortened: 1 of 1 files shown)


dannyrb:
This is actually dangerous to do. There's some discussion in this repository, and in the `cornerstone-wado-image-loader` repository that discuss this.

PR #1229, opened 2020-04-28
Angle and CobbAngle tool: set default 1 for rowPixelSpacing and columnPixelSpacing

aminica:
Only added in the Angle and CobbAngle tool: rowPixelSpacing and columnPixelSpacing default to 1

Bug fix

When there are no pixel spacing data of the image, the values of the angle measurements are not able to be calculated, since the getPixelSpacing function returns null for the rowPixelSpacing and columnPixelSpacing values.

Value 1 is used in the angle value calculation if the rowPixelSpacing and/or columnPixelSpacing values are null.

No.

* **Other information**:
Default value 1 is added since in the past every tool had it, but due to some refactoring (in which the getPixelSpacing function was created), this scenario was left out.

### DIFF
--- a/src/tools/annotation/AngleTool.js
+++ b/src/tools/annotation/AngleTool.js
@@ -115,18 +115,20 @@ export default class AngleTool extends BaseAnnotationTool {
     const { rowPixelSpacing, colPixelSpacing } = getPixelSpacing(image);
 
     const sideA = {
-      x: (data.handles.middle.x - data.handles.start.x) * (colPixelSpacing || 1),
-      y: (data.handles.middle.y - data.handles.start.y) * (rowPixelSpacing || 1)
+      x:
+        (data.handles.middle.x - data.handles.start.x) * (colPixelSpacing || 1),
+      y:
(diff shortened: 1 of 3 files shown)


dannyrb:
@JamesAPetts, this doesn't change labeling from `px` to `mm`, but it does assume square pixels. That feels like a safer assumption. Thoughts on merging this?

sroopra:
Any idea when this will be merged? Or is it blocked by the codeclimate issue?

aminica:
So, all checks have now passed. When could we expect the merge?

milosdjurdjevic:
@dannyrb some info on this?

aminica:
@JamesAPetts can we have some update on this?

dannyrb:
Appreciate your patience on this one. I'll merge, but may revert if someone comes forward and forces the non-square pixel argument.

We may eventually want this behavior as an optional flag that can be sit when cornerstone.init() is called.

aminica:
@dannyrb Thank you very much! :)

PR #1254, opened 2020-06-24
feat: 🎸 Dashed rendering options for tools with lines/handles

JamesAPetts:
Dashed line rendering options for all tools.

Open the console and paste:

```js
const tools = cornerstoneTools.store.state.tools;
tools.forEach(tool => { const config = tool.configuration;  config.renderDashed = true; tool.configuration = config});
```

And all tools will be rendered in dashed mode. This can be done per tool, so you could have dashed in one viewport and solid in another.

The dash style is set in the `globalConfiguation` module, with a default lineDash of `[4,4]`.

### DIFF
--- a/src/drawing/drawArrow.js
+++ b/src/drawing/drawArrow.js
@@ -12,9 +12,10 @@ import drawJoinedLines from './drawJoinedLines.js';
  * @param  {Object} end       The end position.
  * @param  {string} color     The color of the arrow.
  * @param  {number} lineWidth The width of the arrow line.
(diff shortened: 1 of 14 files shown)


dannyrb:
Is the lineDash hard coded? [4, 4]
I think that's fine for now.

Long-term, it would be nice if tools/annotations had a way to override toolStyle in a consistent/granular way.

Thanks for authoring this PR

JamesAPetts:
@dannyrb 
> Is the lineDash hard coded? [4, 4]

That's just the default, you can set it in the `globalConfiguration` module.

Issue #1172, opened 2020-01-29
Should it be possible to add magnify tool and have it work for all elements? 


kuehlc:
### Prerequisites

* [ ] Which version are you using? (Is it latest?)
Latest
* [ ] Are you reporting to the correct repository?
Yes
* [ ] Did you search existing issues? (Were any related?)

### Description

I have tested adding/enabling tools that I want to work on up to 4 images at a time. Pan, zoom, and wwwc tool seem to work fine on each element when added, but the magnify tool only works for a single element.
### Steps to Reproduce the issue

1. When I do X
Right click on second element
2. Then Y
Drag mouse
3. I see behavior Z
Magnify glass shows up on first element

**Expected behavior:** (What you expected to happen)
Magnify tool to work for each individual element when it is added globally.
**Actual behavior:** (What actually happened)
Magnify only works on one element. When trying to use it on second element, it shows up on first element instead.

### CodeSandbox With Reproduction of Issue:
https://codesandbox.io/s/nervous-chebyshev-nfz39
- [Basic Template](https://codesandbox.io/s/cornerstone-basic-issue-template-9slk4)

kuehlc:
@dannyrb or anyone, Have you gotten a chance to look into this yet? I don't mean to rush anyone if it's already being worked on. I just wasn't sure if it has been seen since it has been up for a few days. I don't know what the expected turnaround time for these should be since it is open source.

jlopes90:
I already found a solution.
```
.cornerstone-element {
  position: relative;
}
```

kuehlc:
@jlopes90 that makes it work for when you have more than one image for it to work on?

kuehlc:
@jlopes90 I'll try it, but even if that works, it seems like a bug that you need to add that type of styling. It should work by default like the others do.

dannyrb:
@kuehlc, does that do the trick for you? We could inline the style on the dynamically created element; I believe we don't because we don't want it to interfere with any user defined styles.

kuehlc:
@dannyrb yes, the style change made it work. It just seems like a bug because the other tools work by default for this. I don't see how the same thing for those tools can't be done for magnify, but you know the code better than me.

dannyrb:
@kuehlc, that's a good point. I wouldn't be against a PR that adds the appropriate style a la JS when the magnifying glass should be visible.

kuehlc:
[...]

PR #1271, opened 2020-07-24
Magnify fix

JamesAPetts:
- Fix cleanup of tools on disabling of magnify element.
- Prevent magnify orphans from being formed when an element is removed without tools being disabled.

This appears to fix the following issues:

RE: #1172 
RE: https://github.com/OHIF/Viewers/issues/1518

### DIFF
--- a/src/tools/MagnifyTool.js
+++ b/src/tools/MagnifyTool.js
@@ -3,6 +3,7 @@ import { getNewContext } from '../drawing/index.js';
 import BaseTool from './base/BaseTool.js';
 import { hideToolCursor, setToolCursor } from '../store/setToolCursor.js';
 import { magnifyCursor } from './cursors/index.js';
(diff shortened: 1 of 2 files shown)


src/tools/MagnifyTool.js
```diff
@@ -3,6 +3,7 @@ import { getNewContext } from '../drawing/index.js';
 import BaseTool from './base/BaseTool.js';
 import { hideToolCursor, setToolCursor } from '../store/setToolCursor.js';
 import { magnifyCursor } from './cursors/index.js';
+import { isDebuggerStatement } from '@babel/types';
```
swederik, line 6: What is this?

JamesAPetts: My IDE making bad guesses about my life decisions.

Issue #1279, opened 2020-08-03
Inconsistency between fontSize and font in textStyle API


brunoalvesdefaria:
Currently, the `cornerstoneTools.textStyle` API has inconsistency between the current `fontSize` and `font` internal variables.

### The problem

In the current `CornerstoneTools`' implementation, the `textStyle` API use [defaultFontSize](https://github.com/cornerstonejs/cornerstoneTools/blob/fc4495ca288bd38f34699d94f84b274a7c8d4bfa/src/stateManagement/textStyle.js#L1) and [defaultFont](https://github.com/cornerstonejs/cornerstoneTools/blob/fc4495ca288bd38f34699d94f84b274a7c8d4bfa/src/stateManagement/textStyle.js#L2) internal variables to control the current font styling.
This API also exposes [setFontSize](https://github.com/cornerstonejs/cornerstoneTools/blob/fc4495ca288bd38f34699d94f84b274a7c8d4bfa/src/stateManagement/textStyle.js#L13) and [setFont](https://github.com/cornerstonejs/cornerstoneTools/blob/fc4495ca288bd38f34699d94f84b274a7c8d4bfa/src/stateManagement/textStyle.js#L5) methods to change the value of these variables.
However, when calling the `setFont` method, it will not change the `defaultFontSize`. Also, calling the `setFontSize` method will not change the `defaultFont` value.

### Steps to Reproduce the issue

1. Load the `cornerstoneTools`' library;
2. Then call `cornerstoneTools.setFontSize(50)` and check the values by calling both [getFont](https://github.com/cornerstonejs/cornerstoneTools/blob/fc4495ca288bd38f34699d94f84b274a7c8d4bfa/src/stateManagement/textStyle.js#L9) and [getFontSize](https://github.com/cornerstonejs/cornerstoneTools/blob/fc4495ca288bd38f34699d94f84b274a7c8d4bfa/src/stateManagement/textStyle.js#L17) methods.
3. I see that changing the font size doesn't have any effect on the font string.
4. Additionally, call `cornerstoneTools.setFont('10px Verdana')` and check the values again.
5. I see that changing the font string will also not affect the font size value.

**Expected behavior:** Changing any of the values might reflect the changes in the related variables

**Actual behavior:** Both variables are controlled independently, which can cause inconsistencies when rendering text due to the calculations made with the font size

### CodeSandbox With Reproduction of Issue:

[...]

PR #1280, opened 2020-08-04
Fixing textStyle API inconsistency between font size and font style

brunoalvesdefaria:
* **What kind of change does this PR introduce?**
Fixes https://github.com/cornerstonejs/cornerstoneTools/issues/1279

* **What is the current behavior?**
https://github.com/cornerstonejs/cornerstoneTools/issues/1279

* **What is the new behavior?**
 * A new `textStyle` object was created to control the styling.
 * New `setFontFamily` and `getFontFamily` methods were implemented.
 * Method `getFontSize` was changed to consider both `fontSize` and `fontFamily` attributes.
 * Method `setFontSize` was changed to consider both attributes as well but must be deprecated.

* **Does this PR introduce a breaking change?**
No, it will fix an existing issue.

### DIFF
--- a/src/stateManagement/textStyle.js
+++ b/src/stateManagement/textStyle.js
@@ -1,38 +1,56 @@
-let defaultFontSize = 15,
-  defaultFont = `${defaultFontSize}px Arial`,
-  defaultBackgroundColor = 'transparent';
+const textStyle = {
+  fontSize: 15,
+  fontFamily: 'Arial',
+  backgroundColor: 'transparent',
+};
 
+// Deprecate
 function setFont(font) {
-  defaultFont = font;
+  const split = font.split('px ');
+
+  if (split.length === 2) {
+    setFontSize(split[0]);
+    setFontFamily(split[1]);
+  }
 }
 
 function getFont() {
-  return defaultFont;
+  return `${textStyle.fontSize}px ${textStyle.fontFamily}`;
+}
+
+function setFontFamily(fontFamily) {
+  textStyle.fontFamily = fontFamily;
+}
+
+function getFontFamily() {
+  return textStyle.fontFamily;
 }
 
 function setFontSize(fontSize) {
-  defaultFontSize = fontSize;
+  textStyle.fontSize = parseFloat(fontSize);
 }
 
 function getFontSize() {
-  return defaultFontSize;
+  return textStyle.fontSize;
 }
 
 function setBackgroundColor(backgroundColor) {
-  defaultBackgroundColor = backgroundColor;
+  textStyle.backgroundColor = backgroundColor;
 }
 
 function getBackgroundColor() {
-  return defaultBackgroundColor;
+  return textStyle.backgroundColor;
 }
 
-const textStyle = {
+const textStyleApi = {
   setFont,
   getFont,
   setFontSize,
   getFontSize,
+  setFontFamily,
+  getFontFamily,
   setBackgroundColor,
   getBackgroundColor,
 };
 
-export default textStyle;
+export default textStyleApi;


src/stateManagement/textStyle.js
```diff
@@ -1,38 +1,56 @@
...
+  fontFamily: 'Arial',
+  backgroundColor: 'transparent',
+};
 
+// Deprecate
 function setFont(font) {
-  defaultFont = font;
+  const split = font.split('px ');
+
+  if (split.length === 2) {
```
ladeirarodolfo, line 11: Is it going to be deprecate?
Given param is restritected to: fontSize|fontFamily:
1. Would be nice if we parametrize the position each item
2. Would be nice if there is a minimum validation
3. Can we default values to initial textStyle values?
3.1 Is it possible to accept a list of font-family names and/or generic-family names?

src/stateManagement/textStyle.js
```diff
@@ -1,38 +1,56 @@
...
+  textStyle.fontFamily = fontFamily;
+}
+
+function getFontFamily() {
+  return textStyle.fontFamily;
 }
 
 function setFontSize(fontSize) {
-  defaultFontSize = fontSize;
+  textStyle.fontSize = parseFloat(fontSize);
```
ladeirarodolfo, line 30: Can we do a minimum validation to check if is Number?

src/stateManagement/textStyle.js
```diff
@@ -1,38 +1,56 @@
...
 }
 
 function getFontSize() {
-  return defaultFontSize;
+  return textStyle.fontSize;
 }
 
 function setBackgroundColor(backgroundColor) {
-  defaultBackgroundColor = backgroundColor;
+  textStyle.backgroundColor = backgroundColor;
```
ladeirarodolfo, line 38: as far as default backgroundColor has value equals 'transparent' its ok to not do any kind of validation, but I would recommend to at least add a comment here

ladeirarodolfo (changes requested):

src/stateManagement/textStyle.js
brunoalvesdefaria: Added.
Thank you!

src/stateManagement/textStyle.js
brunoalvesdefaria: Yes, it is going to be deprecated.
Updated the code and added a more detailed jsDoc with explanations.

src/stateManagement/textStyle.js
brunoalvesdefaria: This functions accepts any value that represents a color. Examples:

- transparent
- red, green, blue
- rgb(255,0,0)
- rgba(0,255,0,0.5)
- #FFC
- #FF0000
- #FF993399
- And any format accepted by HTML

brunoalvesdefaria: JSDoc will be added to the function

src/stateManagement/textStyle.js
brunoalvesdefaria: 3. As we discussed, it will not worth changing the original function as it's being deprecated.

igoroctaviano:
@swederik can you please review this one? it will be needed for a client. Currently, we are pointing to a fork with these changes.

igoroctaviano:
@sedghi can you review + test this pull request please? for more details, check out the ticket in the issue description.

src/stateManagement/textStyle.js
```diff
@@ -1,38 +1,110 @@
...
+
+/**
+ * Sets the font family that will be used for general font rendering.
+ * @see {@link https://www.w3.org/TR/css-fonts-3/#font-family-prop|3.1. Font family: the font-family property}
+ *
+ * @param {string} fontFamily Name of the font to be used
+ * @returns {void}
+ */
+function setFontFamily(fontFamily) {
+  textStyle.fontFamily = fontFamily;
```
sedghi, line 43: what happens if i `setFontFamily` to null or undefined or non-standard font names?

src/stateManagement/textStyle.js
sedghi: JSDoc is not updated yet

sedghi:
For some reason I'm not able to change the font size in the demo 

normal font size
image

changing font size via
```
  cornerstoneTools.init([
    {
      moduleName: 'globalConfiguration',
      configuration: {
        showSVGCursors: true
      }
    },
    {
      moduleName: 'segmentation',
      configuration: {
        outlineWidth: 2
      }
    }
  ]);

cornerstoneTools.textStyle.setFontSize(50)
```

after font size change 

image

it is just spaced out

src/stateManagement/textStyle.js
igoroctaviano: I think this is the user's responsibility? maybe adding checks could lead to silently broken code. @brunoalvesdefaria any thoughts on this?

igoroctaviano:
> For some reason I'm not able to change the font size in the demo
> 
> normal font size
> image
> 
> changing font size via
> 
> ```
>   cornerstoneTools.init([
>     {
>       moduleName: 'globalConfiguration',
>       configuration: {
>         showSVGCursors: true
>       }
>     },
>     {
>       moduleName: 'segmentation',
>       configuration: {
>         outlineWidth: 2
>       }
>     }
>   ]);
> 
> cornerstoneTools.textStyle.setFontSize(50)
> ```
> 
> after font size change
> 
> image
> 
> it is just spaced out

That's weird, @brunoalvesdefaria any thoughts on this comment?

PR #1288, opened 2020-08-05
Update convertToVector3 - handle similar objects

mix3d:
In case the Vector3 `instanceof` doesn't match, but the object still has `{x,y,z}` values

No doc change needed

Kind of a bugfix/code hardening

If an object fails the `instanceof` check(which is currently happening for me, likely a webpack issue), it attempts to grab array values that don't exist, passing undefined to the constructor giving you a Vector3 of [0,0,0].

passing in a Vector3-like object will be properly converted to a Vector3

It is technically breaking if you were using it wrong, but no behavior change for the expected inputs.


JamesAPetts:
Nice, tested and LGTM, tests are always nice, thanks for the contrib!

PR #1316, opened 2020-09-25
feat(RotateTool): Add rounded/integer angles option

mix3d:

Adds a rounding option to the RotateTool that rounds to integer numbers when configured.

No integer rounding is applied, and the scale/angle values create tiny fractional angles, which do not display well. We could round the display of these numbers, but other logic in a vtkView uses whole integer angles.
image

Optional config for the `RotateTool` allows for rotating at whole integer increments, rounding up to the nearest angle.

No

* **Other information**:
#hacktober!


mix3d:
looks like my branch history is funky, lemme fix that.

PR #1324, opened 2020-11-05
Request pool manager options

JamesAPetts:
Cornerstone imageLoaders registered to cornerstone can take options which may specific to the loader themselves.

This PR adds an argument to the request pool manager to pass options down options to its managed `cornerstone.loadImage` and `cornerstone.loadAndCacheImage` calls.

These options are merged with the default options request pool manager provides.


swederik (changes requested):
Options shouldn't overwrite priority and type, otherwise it's fine.

PR #1340, opened 2020-12-11
Remove any orphaned event listeners on disable (#1337)

### LINKED ISSUE #1337: onMouseMove listener not removed when element is disabled while mouse button is pressed
This is potentially related to #731.  The [`mousedown` event handler adds `mousemove` and `mouseup` listeners](https://github.com/cornerstonejs/cornerstoneTools/blob/master/src/eventListeners/mouseEventListeners.js#L273) when the mouse button is pressed.  Those are both typically removed when the mouse button is released (in the `mouseup` handler).

However, if the mouse button is still being held down when the element is disabled, those nested `mousemove` and `mouseup` event listeners are never removed.  This causes an endless stream of "element not enabled" exceptions for every mouse movement when it tries to call `cornerstone.pageToPixel(element, e.pageX, e.pageY)` since `element` is no longer enabled.

We are seeing this occasionally when a user happens to press their mouse button on a canvas just prior to using a keyboard shortcut that causes the canvas to be destroyed.

I'll see if I can come up with a PR soon.

kevinleedrum:
This fixes #1337.  No breaking changes.  Just mitigating exceptions.

Currently, if you disable an element while a mouse button is held down, the two `onMouseMove` and `onMouseUp` listeners that are added _by_ the `mousedown` handler are not removed, causing an endless stream of exceptions.

To fix this, I've added a `Map` called `addedListeners` in the parent scope.  When `mousedown` is invoked, its `onMouseMove` and `onMouseUp` listeners get added to that Map.  If the problem scenario occurs, we now have a handle to all the generated listeners so we can remove them in the `disable` function.

I used the listener functions themselves as the keys in the Map because it is possible to have multiple `mousedown` and `mousemove` listeners when you have more than one mouse button pressed.

### DIFF
--- a/src/eventListeners/mouseEventListeners.js
+++ b/src/eventListeners/mouseEventListeners.js
@@ -10,6 +10,8 @@ let isClickEvent = true;
 let preventClickTimeout;
 const clickDelay = 200;
 
+const addedListeners = new Map();
+
 function getEventButtons(event) {
   if (typeof event.buttons === 'number') {
     return event.buttons;
@@ -264,6 +266,8 @@ function mouseDown(e) {
 
     document.removeEventListener('mousemove', onMouseMove);
     document.removeEventListener('mouseup', onMouseUp);
+    addedListeners.delete(onMouseMove);
+    addedListeners.delete(onMouseUp);
 
     element.addEventListener('mousemove', mouseMove);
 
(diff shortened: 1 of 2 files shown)


kivikakk:
I believe I'm hitting a bug in this. It looks like the introduced code is incorrect:

https://github.com/kevinleedrum/cornerstoneTools/blob/059eae364a67bb956b205d492cedf7dfee8ea37f/src/eventListeners/mouseEventListeners.js#L366-L368

[Per MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/forEach), `Map.prototype.foreach` expects the callback to have arguments beginning `value, key`. `addedListeners` is a Map where the **keys are listeners**, but here we define `listener` as the first argument, which receives the value, i.e. the string `"mousemove"` or similar.

As a result, we try to call `removeEventListener` with the callback first, and the event name second. On Firefox this produces:

image

I will open a PR to address this.

PR #1356, opened 2021-02-23
feat(RequestPoolManager): Added custom request type and custom priority

sedghi:
Added a `custom` request type to the `addRequest` in the `requestPoolManager`. The custom request with controllable priority is useful when caching the studies in advance for the next case

**What is the current behavior?**
Only `interaction, thumbnail, prefetch` can be used as the request type, with a fixed priority value.

**What is the new behavior?**
A `custom` request type can be defined with a custom priority. In addition, the priority of the  `interaction, thumbnail, prefetch` can be changed if needed.

**Does this PR introduce a breaking change?**
No, it will only add a new feature to enable custom priorities for the requests

**Other information:**
None

### DIFF
--- a/src/requestPool/requestPoolManager.js
+++ b/src/requestPool/requestPoolManager.js
@@ -5,19 +5,30 @@ const requestPool = {
   interaction: [],
   thumbnail: [],
   prefetch: [],
+  custom:[]
 };
 
 const numRequests = {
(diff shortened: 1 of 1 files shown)


sedghi:
This will be requested again from a cornerstonTools branch

PR #1373, opened 2021-04-22
Fix for #1337: reverse Map forEach arguments

kivikakk:
Per https://github.com/cornerstonejs/cornerstoneTools/pull/1340#issuecomment-824532127, we reverse the order of arguments to the `addedListeners` forEach when removing event listeners, in order to call `document.removeEventListener` correctly.

Tested as fixing the issue in my app, where disabling a Cornerstone element while dragging was causing exceptions here.

I'm afraid I haven't the time to work on adding a full test here, but this orphaned event handler removing code is currently (completely) broken, so hoping the fix alone is valuable.


kevinleedrum (approved):
I could have sworn I tested this thoroughly, but I am getting old.

kivikakk:
:heart: Me too.

PR #1407, opened 2021-07-26
feat(LengthTool): Added digits configuration for Length tool

brunoalvesdefaria:

A new feature which enables the configuration of the amount of digits displayed for the Length tool text box.

The current behavior displays always 2 digits for the Length tool text box.

Now it will be possible to add the `digits` attribute to the Length tool's configuration object in order to change the amount of digits to be displayed by the measurements' text box.

No

* **Other information**:
Steps to validade the new behavior:

1. Open the Netlify preview for this PR;
2. Open the browser's developer tools;
3. Change the Length tool configuration from CornerstoneTools' store: `cornerstoneTools.store.state.tools[20].configuration.digits = 1`;
4. Select the length tool and draw length measurements.
5. Note that the amount of digits for the measurements' text boxes have decreased to 1.


sedghi:
@brunoalvesdefaria Since changes are minor, why not adding them to all annotation tools? Otherwise looks great

brunoalvesdefaria:
> @brunoalvesdefaria Since changes are minor, why not adding them to all annotation tools? Otherwise looks great

Thank you for reviewing it @sedghi!
I didn't include this configuration for BaseAnnotationTool because some of them doesn't display numbers and some display numbers that would not make sense to have decimal digits being displayed.
I think it's better to keep the configuration individual for each tool.

mix3d:
Throwing in my 2 cents: if it's implemented at the base annotation level, each tool can choose what it needs to display and can optionally allow for configuration if it makes sense.

If you only add it to length, then you have to duplicate code when adding it to the circle annotation, for example.

brunoalvesdefaria:
> Throwing in my 2 cents: if it's implemented at the base annotation level, each tool can choose what it needs to display and can optionally allow for configuration if it makes sense.
> 
> If you only add it to length, then you have to duplicate code when adding it to the circle annotation, for example.

I'll summon @swederik to ask his opinion.

We have 3 options:

1. Keep as is (only for Length tool) and replicate the configuration for other tools that might need it.

2. Change the `BaseAnnotationTool` to include digits configuration for all annotation tools (even the ones who doesn't need that, like TextMarkerTool).
If we do that, we would need to append a `constructor` block to the `BaseAnnotationTool` file and include a custom configuration override to add only this `digits` property (might be polluting the file?).

3. Do it at the `BaseTool` level, but it would include digits configuration for all tools (even stack scrolling), which I think it could be confusing. So I think this option is not ideal.

swederik:
I wouldn't do it in BaseAnnotationTool because some tools show more than one value. You might want one level of precision for area and another for mean HU/SUV. I know it's a bit ugly to put it in each tool but it's probably the best option in my opinion.

PR #1422, opened 2021-08-27
fix(deepmerge): fix the prototype methods missing when merging

Enside:
When a custom tool extends one tool to override its svgCursor property,
the methods of MouseCursor class is lost in deepmerge process, so there
will be an error with 'xxx is not a function...'.


Enside:
This problem also occurs in the following code
`    cornerstoneTools.addToolForElement(element, tool, {
      svgCursor: new MouseCursor(stackScrollCursor)
    })`
@dannyrb

Enside:
After fixing deepmerge, svgCursor property is set correctly.

PR #1451, opened 2021-11-17
fix(cornerstoneTools):Fetch the overlay if an async method is available

wayfarer3130:
This fix allows the overlay tool to call an asynchronous retrieveBulkData method if the overlay pixeldata has a retrieveBulkData method, and then render the overlay correctly.


src/tools/OverlayTool.js
```diff
@@ -29,6 +29,56 @@ export default class OverlayTool extends BaseTool {
...
+
+  /** Gets a byte array of overlay data.
+   * Defaults to returning pixelData if it is a Uint8Array, otherwise tries generating a Uint8Array from:
+   * 1. pixelData.InlineBinary, if it is Base64 encoded, then decodes to Value as a Uint8Array
+   * 2. pixelData.Value, if it is a Uint8Array, assumes it is a packed array
+   * 3. pixelData.retrieveBulkData.then(enabledElement.refresh), assuming it needs to fetch it.
+   * This method does depend on some of the JSON DICOM representations, but those don't have to come from
+   * JSON, it falls back to the original Uint8Array representation, which can come from anywhere.
+   */
+  getOverlayData(pixelData, element) {
```
wayfarer3130, line 56: @chafey This is the asynchronous version of the retrieve code.  It assumes 'dicom' format if the retrieveBulkData is present, which isn't unreasonable, and otherwise falls back to the existing behaviour.  To do that, the retrieve call is made, but also the data is handled as InlineBinary and other various formats to support JSON based overlays.

Issue #210, opened 2017-06-06
Ultrasound region calibration information


kofifus:
How can I get the Length tool to show the length in mm / get the correct scale with an image using the  ultrasound region calibration information ?

Thx!

dannyrb:
@kofifus, I might be able to help if you provide an example image and information? I'm not a radiologist, so I'm not sure what "utrasound region calibration information" is.

kofifus:
Thanks!

Please see here:
http://dicom.nema.org/medical/dicom/2016c/output/chtml/part03/sect_C.8.5.5.html

"US images often contain multiple regions that have independent data regions, e.g., quad screen loops that may have different calibration information. The data presented in the various regions of a US image can represent a multiplicity of physical parameters, e.g., spatial distance, blood velocity, time, volume, etc., and these are often contained in the value of the pixel itself. It is therefore imperative that physical information be available for the various regions of a single region independent of each other."

dannyrb:
@kofifus, I believe images currently only support a single calibration. If ultrasounds contain metadata that would allow regions with different calibrations, I'm sure the functionality you want could work.

Are there specific tags that contain the data? If its in the pixel data itself, how would you propose adding this functionality?

kofifus:
The relevant DICOM standard is here http://dicom.nema.org/medical/dicom/2016e/output/chtml/part03/sect_C.8.5.5.html

Cornerstone should have this as its reasonably common, however I don't have the knowledge to implement it...

dannyrb:
Do you know of other views that implement this that we could look to as an example? It's been over a year and you're the only requester that I am aware of. Realistically, for this to be accomplished in the near future would require, at a minimum:

- one of more people to volunteer scoping out requirements
- one or more people spearheading those requirements

I'd be happy to offer guidance, but I don't have the domain knowledge required, or the time, to do either of the above. Is this something you would like to take ownership of?

jdnarvaez:
I’ve implemented this previously and it’s on my list of things to do, but hasn’t been a high priority.

rudlofth:
We are currently looking into using cornerstone for an DICOM ultrasound viewer. In ultrasound images or clips with several regions having different meaning and calibrations are common. Typical examples are

1. Multi images: An DICOM image shows several images next to each other. So each calibration region corresponds to one of those images. The pixel distance in each region might be different. Any measurements should happen only within one region with the correct pixel distance. Measuring across region boundaries or off any region does not make any sense.

[...]

PR #1466, opened 2022-02-14
Ultrasound Region Measurements

jdnarvaez:
Addresses https://github.com/cornerstonejs/cornerstoneTools/issues/210

- [X] The commit message follows our guidelines
- [X] Tests for the changes have been added (for bug fixes / features)
- [N/A] Docs have been added / updated (for bug fixes / features)

This changes how the row pixel spacing, column pixel spacing, and measurement suffix are determined for a Length measurement. If an image contains a non zero length value for the Sequence of Ultrasound Regions (0018,6011) tag, then these regions are searched to find a single region that contains both measurement points (sorted by largest overall area first). If the measurement spans multiple regions, then the spacing information is only used if the regions contain the same information. If no region is found, the measurement is assumed to be that of 'pixels'. If an image does not contain a non-zero length value for this tag, then the original `getPixelSpacing` method is called as before.


src/tools/annotation/LengthTool.js
```diff
@@ -295,3 +302,78 @@ function _sanitizeMeasuredValue(value) {
...
+    // the regions were specified, then points that exist outside of the regions
+    // have an unknown physical spacing
+    const physicalDeltaX =
+      contained.length > 0 ? contained[0].physicalDeltaX : 1;
+    const physicalDeltaY =
+      contained.length > 0 ? contained[0].physicalDeltaY : 1;
+
+    return {
+      rowPixelSpacing: physicalDeltaY,
+      colPixelSpacing: physicalDeltaY,
```
whulbert4, line 365: I was implementing something similar in my app that is using 2.X version and was reviewing your PR and noticed that this should return physicalDeltaX I believe. Excited to see this PR either way once I upgrade my app to a later version.

danielraggs:
@jdnarvaez Hello! Why was this closed? Has this already been implemented in some way in [Cornerstone3D](https://github.com/cornerstonejs/cornerstone3D/tree/a33572ac280ef5d42b56083161c020d04e4d30c2)?

EDIT: Ahh nvm, seems to have been added [here](https://github.com/cornerstonejs/cornerstone3D/pull/927).

jdnarvaez:
@danielraggs no one reviewed it.

PR #1549, opened 2023-07-18
fix(SegRendering): preserve canvas transform matrix after seg render

ViniciusResende:

This PR contains a bug fix.

The current behavior is that my overlay (rendered using the `Overlay Tool`) is getting resized and translated once the segmentation is rendered.

image
Note that the text is not even entirely on the screen, also there are two ellipsis that aren't even being shown, they appear on the screenshot below.

Now my overlay is not being resized nor translated and renders on the position where I expected it to be, also the segmentation kept working fine.

image
Now we can see that the text is well positioned and that the ellipsis are in the right place.

This PR **shoudn't** introduce any breaking change.

* **Other information**:
Previously after rendering the segmentation cornerstone was always transforming the canvas to follow the identity matrix, but in cases where we had an overlay using a different transformation matrix, it would undesirably affect the original overlay.
Also did some minor refactorings on the `renderFill` function of the `src/eventListeners/internals/renderSegmentationFill.js` file.


igoroctaviano:
LGTM 🏅

